PR #8, opened 2015-10-19
Let's not deprecate this

borkweb:
The original rework had eliminated the Tribe__Events__Template_Factory class, however, the end result was the continued existence of that class that extends Tribe__Template_Factory.

See: https://central.tri.be/issues/40318#note-2


zbtirrell:
:+1:

PR #16, opened 2015-10-28
Avoid using $this in a static context (spotfix)

barryhughes:
Use the `menuName` property from the settings object (previously using `$this` and a non-existent property within a static method).


bordoni:
:wink:

PR #21, opened 2015-11-02
If running TEC 3.12.x or lower, don't initialize tribe-common pages

borkweb:
See: https://central.tri.be/issues/41175


bordoni:
Good docblock! :+1:

PR #24, opened 2015-11-06
Compare to 4.0beta

bordoni:
Our Settings page were not been added due to `version_compare` considering `4.0beta` lower than `4.0`.


borkweb:
:+1:

PR #28, opened 2015-11-10
Add AJAX helper function where we can override as necessary

borkweb:
This change - once implemented in all the places, should make testing code that reacts to DOING_AJAX a bit easier without breaking things.

See: https://central.tri.be/issues/41385


barryhughes:
Makes sense :+1:

PR #31, opened 2015-11-13
Add rule to make sure our modals are functional | #41462

barryhughes:
Modals can be obscured under the grey 'screen'. This change is essentially just copying a rule currently used in TEC's admin CSS to common.

[C#41462](https://central.tri.be/issues/41462)


bordoni:
:+1:

PR #38, opened 2015-11-24
Don't escape so early. Escape on output, not on URL generation

borkweb:
This was returning an escaped URL which was breaking downstream escapes

See: https://central.tri.be/issues/41833


zbtirrell:
:+1:

PR #45, opened 2015-12-01
Make sure the Add-Ons page is added to the correct top-level section

borkweb:
See: https://central.tri.be/issues/41978 and https://central.tri.be/issues/41979


bordoni:
That's the weird part I've changed that twice and did testing it was working. Anyways looking good for a merge.

PR #47, opened 2015-12-01
Adjust, add, and remove fields from the tribe-common default settings

borkweb:
The fields that are being removed are TEC specific, so will be moved there.

See: https://central.tri.be/issues/41980


bordoni:
:100:

PR #48, opened 2015-12-01
Remove dateformat fields

borkweb:
These fields are specific to TEC and will be moving back to there.

See: https://central.tri.be/issues/41980#note-3


bordoni:
:8ball:

PR #50, opened 2015-12-02
Load the textdomain on init

borkweb:
So...we had all kinds of localization methods..but we weren't loading the text domain. This fix adds the textdomain loading at init (which mimics when/where the-events-calendar loads its textdomain stuff). This was discovered when I attempted to generate a .pot file and poedit was like: "Yeah...this plugin doesn't use translations"


zbtirrell:
:+1:

PR #51, opened 2015-12-02
Textdomains for plugins need to be loaded relative to the plugins dir

borkweb:
Common is a child directory of either the-events-calendar or event-tickets


zbtirrell:
won't this fail if the plugin is loaded standalone? (not that we are doing that at this point, but...)

borkweb:
Yeah it will, but like you said - we don't expect that to be done at this point.

zbtirrell:
ok, carry on :musical_keyboard:

PR #54, opened 2015-12-08
Use the correct object member

barryhughes:
When trying to figure out a good way to set up a test for QA purposes, I noticed a couple of minor oversights that needed correction in this class.

Also, a minor CSS tweak to accommodate longer version #s more gracefully.

[C#42047](https://central.tri.be/issues/42047)


borkweb:
:+1:

PR #58, opened 2015-12-29
Replace HTTP host field with value of the site and home URL options | 26385

barryhughes:
Previously we just used the HTTP host field for the URL included in the sys info - which was misleading at times. This replaces it with both the site and home URLs.

[C#26835](https://central.tri.be/issues/26835)


borkweb:
:+1:

PR #60, opened 2016-01-09
Feature/28582 tpl ver tracking

barryhughes:
Adds a scanner that examines and compares `@version` tags within templates and overrides and a reporting tool (output -> settings/help tab), plus supporting tests.

[C#28582](https://central.tri.be/issues/28582)


borkweb:
This is pretty slick! :+1:

PR #65, opened 2016-02-01
Ignore beta/rc type suffixes in version numbers | 28582

barryhughes:
Alleviate confusion in the template updates scanner when working with alpha/beta/RC versions.

https://central.tri.be/issues/28582


bordoni:
Me likey!

PR #67, opened 2016-02-24
do not strip slashes

lucatume:
Ticket: https://central.tri.be/issues/41127

This PR remove the slash stripping that's preventing date format from being applied as intended; at that point slashes are the correct ones that are part of the format specification


borkweb:
:+1:

PR #72, opened 2016-03-17
If WP names the downloaded PUE directory oddly, fix it

borkweb:
The WP Updater class is converting the PUE URL into a directory name, hence the funky super long directory names that were being generated:

```
theeventscalendar.compu_request_pluginevents-calendar-proversion4.1pu_get_download1pue_active_version4.0-Ul4C4S
```

See: https://central.tri.be/issues/44870


bordoni:
:+1:

afragen:
You may need to use `upgrader_source_selection` as I remember issues with `upgrsder_post_install`.

Example, https://github.com/afragen/wp-dependency-installer/blob/develop/dependency-installer/wp-dependency-installer.php#L327

borkweb:
@afragen - thanks for the tip! The urgency of this specific issue has been minimized with some tweaks to our packaging script, though I'd still like to implement this fix in some fashion so that using the one-click updater doesn't change directory names at all - whether from a badly packaged zip _or_ the times that users change the name of their plugin directories.

I'll close this out and try to marry the stuff I have above with your insight on `upgrader_source_selection`. :smile:

PR #75, opened 2016-03-25
Generate a timezone string from a UTC offset

borkweb:
Related: https://github.com/moderntribe/events-pro/pull/309

See: https://central.tri.be/issues/42765

### DIFF
--- a/src/Tribe/Timezones.php
+++ b/src/Tribe/Timezones.php
@@ -94,7 +94,9 @@ class Tribe__Timezones {
 	}
 
 	/**
-	 * Generates a timezone string from a UTC offset
+	 * Helper function to retrieve the timezone string for a given UTC offset
+	 *
+	 * This is a close copy of WooCommerce's wc_timezone_string() method
 	 *
 	 * @param string $offset UTC offset
 	 *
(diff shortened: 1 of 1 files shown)


src/Tribe/Timezones.php
```diff
@@ -94,6 +94,36 @@ public static function wp_timezone_abbr( $date ) {
...
+		if ( ! strpos( $offset, ':' ) ) {
+			$offset .= ':00';
+		}
+
+		$offset = str_replace( 'UTC', '', $offset );
+
+		list( $hours, $minutes ) = explode( ':', $offset );
+		$seconds = $hours * 60 * 60 + $minutes * 60;
+
+		$timezone = timezone_name_from_abbr( '', $seconds, 1 );
```
barryhughes: Mostly just curious - but why prefer testing for timezones with DST in effect before looking for non-DST zones?

barryhughes: ... thinking it may not matter for the week view issue at hand, but are there other places this method might be used from in future where it could introduce hard-to-find weirdness and so could it be worth adding a caveat in the docblock about this behaviour?

If you feel that's overkill, that's fine too :+1:

borkweb: I initially was mimicking this [stackoverflow answer](http://stackoverflow.com/a/11896631/124581) to ensure we didn't bump into [a bug with `timezone_name_from_abbr()`](https://bugs.php.net/bug.php?id=44780)...but I suppose that's less clear than what WooCommerce is doing to accomplish a similar goal. I went ahead and duplicated the relevant code from WooCommerce and called that out in the docblock.

barryhughes:
:+1:

PR #76, opened 2016-04-04
Add a new Helper method to allow easier Loading of the L10N files

bordoni:
_Ref:_ [C#36246](https://central.tri.be/issues/36246)


borkweb:
:+1:

PR #78, opened 2016-04-13
Added the Post Root Pool utility class

lucatume:
Ticket: https://central.tri.be/issues/25848

This PR adds the `Post_Root_Pool` utility class.
This is used to generate site wide unique post roots to use for tickets.
Priming of the pool happens at first use and a method `set_pool` exists to allow for programmatic priming.

See the tests for expected output.

### DIFF
--- a/src/Tribe/Utils/Post_Root_Pool.php
+++ b/src/Tribe/Utils/Post_Root_Pool.php
@@ -1,0 +1,53 @@
+<?php
+
+
+class Tribe__Utils__Post_Root_Pool {
+
+	/**
+	 * @var string
+	 */
+	protected $root_separator = '-';
+
+	/**
+	 * Generates a unique root for a post using its post_name.
+	 *
+	 * @param WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function generate_unique_root( WP_Post $post ) {
+		$post_name = $post->post_name;
+
+		$root = $this->build_root_from( $post_name );
+
+		return $root . $this->root_separator;
+	}
+
+	/**
+	 * @param $post_name
+	 *
+	 * @return string
+	 */
+	protected function build_root_from( $post_name ) {
+		$frags = explode( '-', $post_name );
+
+		$candidate = implode( '', array_map( 'strtoupper', $frags ) );
+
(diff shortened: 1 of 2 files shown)


src/Tribe/Utils/Post_Root_Pool.php
```diff
@@ -0,0 +1,164 @@
...
+	protected $postfix = 1;
+
+	/**
+	 * Generates a unique root for a post using its post_name.
+	 *
+	 * @param WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function generate_unique_root( WP_Post $post ) {
```
barryhughes: Is it possible that this method will be called twice for the _same_ post? If not, ignore this comment!

... but if so, wouldn't we end up with multiple unique roots for the same post? It's clear from the tests what we expect when different posts of the same name are passed in, but not what we expect if the same post is passed in more than once (if indeed it matters).

barryhughes: Ah OK, I _think_ I got my answer over here:

[Tribe__Tickets_Plus__Meta__Unique_ID::get_event_root()](https://github.com/moderntribe/event-tickets-plus/pull/147/files#diff-9c31cd9ece5407537804d45d92c59d73R82)

So if the responsibility for preventing multiple unique roots per post is on the calling function maybe add a note in the docblock or even mark the method `@internal`.

lucatume: The post root pool is meant to be generic and agnostic of where it's used.
The context calling it, in this case the `Tribe__Tickets_Plus__Meta__Unique_ID` class, should take care, as it does, to generate new roots only when needed.
Updating the doc block just in case.

lucatume: Even better, see this:
https://github.com/moderntribe/tribe-common/pull/78/commits/d17522ad9b7289f35fe0deeaabcff1e95b5309bd

lucatume:
I've added both in here and in Tickets Plus use and support for a redundant free post root generation

barryhughes:
_Noyce_ :1234:

PR #79, opened 2016-04-13
Add JSON-LD Abstract

bordoni:
_Ref:_ [C#43595](https://central.tri.be/issues/43595)

### DIFF
--- a/src/Tribe/JSON_LD/Abstract.php
+++ b/src/Tribe/JSON_LD/Abstract.php
@@ -52,7 +52,10 @@ abstract class Tribe__JSON_LD__Abstract {
 	 * Compile the schema.org event data into an array
 	 */
 	public function get_data( $post = null, $args = array() ) {
-		$post = get_post( Tribe__Main::post_id_helper( $post ) );
+		if ( ! $post instanceof WP_Post ) {
+			$post = Tribe__Main::post_id_helper( $post );
+		}
+		$post = get_post( $post );
 
 		if ( ! $post instanceof WP_Post ) {
 			return $data;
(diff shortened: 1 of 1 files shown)


src/Tribe/JSON_LD/Abstract.php
```diff
@@ -0,0 +1,149 @@
...
+			 * Allows the event data to be modifed by themes and other plugins.
+			 *
+			 * @example tribe_json_ld_thing_data_object
+			 * @example tribe_json_ld_event_data_object
+			 *
+			 * @param object $data objects representing the Google Markup for each event.
+			 * @param array $args the arguments used to get data
+			 * @param WP_Post $post the arguments used to get data
+			 */
+			$data[ $post_id ] = apply_filters( $this->get_filter( 'object' ), $_data, $args, get_post( $post_id ) );
```
barryhughes: I don't know if this will actually be a problem or not but for reasons of WP Parser compatibility I'm wondering if it might be better to compose the hook name via string interpolation instead of a method call. 

That is, it feels like it might work more gracefully with a hook like `"some_prefix_{$dynamic_section}"` than `$this->get_string()` type constructs. We should confirm unless you already know the answer.

src/Tribe/JSON_LD/Abstract.php
```diff
@@ -0,0 +1,149 @@
...
+		/**
+		 * Allows the event data to be modifed by themes and other plugins.
+		 *
+		 * @example tribe_json_ld_thing_data_array
+		 * @example tribe_json_ld_event_data_array
+		 *
+		 * @param array $data objects representing the Google Markup for each event.
+		 * @param array $args the arguments used to get data
+		 */
+		$data = apply_filters( $this->get_filter( 'array' ), $data, $args );
```
barryhughes: As above (string interpolation vs method call for creating the hook name).

src/Tribe/JSON_LD/Abstract.php
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	public function markup( $post = null, $args = array() ) {
+		$html = $this->get_markup( $post, $args );
+
+		/**
+		 * Allows users to filter the end markup of JSON-LD
+		 * @deprecated
+		 * @todo Remove on 4.4
+		 * @var string The HTML for the JSON LD markup
```
barryhughes: Here and in the following hook: we should use `@param` instead of `@var` ... also, even if it feels weird on those occasions where we pass the result of a method call instead of an actual var, I feel like the WP standard is still to provide a var name, ie we should gun for:

`@param type $name optional description`

Instead of:

`@var type optional description`

bordoni:
@barryhughes Updated the code :wink:

borkweb:
:+1:

PR #81, opened 2016-04-19
Basis for logging across the plugin suite

barryhughes:
Adds a logging framework, with help tab integration, that our plugins can take advantage of. 

Notes:

* Includes a @todo item, I'm aware of this but inclined to defer further work for this specifically
* Changes in `get_content_html()` (stripping out usage of Kses) was discussed in team chat, I don't believe there's a disadvantage to this and it makes it easier to add new items to the help screen - definitely up for more discussion though :-)

[C#36047](https://central.tri.be/issues/36047)

### DIFF
--- a/src/Tribe/Admin/Help_Page.php
+++ b/src/Tribe/Admin/Help_Page.php
@@ -816,12 +816,12 @@ class Tribe__Admin__Help_Page {
 
 					<dt><?php esc_html_e( 'Rating:', 'tribe-common' ); ?></dt>
 					<dd><a href="<?php echo esc_url( $plugin->stars_url ); ?>" target="_blank">
-							<?php wp_star_rating( array(
-								'rating' => $api_data->rating,
-								'type'   => 'percent',
-								'number' => $api_data->num_ratings,
-							) );?>
-						</a></dd>
+						<?php wp_star_rating( array(
+							'rating' => $api_data->rating,
+							'type'   => 'percent',
+							'number' => $api_data->num_ratings,
+						) );?>
+					</a></dd>
 				</dl>
 			<?php } ?>
 
--- a/src/Tribe/Admin/Help_Page.php
+++ b/src/Tribe/Admin/Help_Page.php
@@ -411,7 +411,10 @@ class Tribe__Admin__Help_Page {
 	}
 
 	/**
-	 * Parses the help text from an Array to the final HTML
+	 * Parses the help text from an Array to the final HTML.
+	 *
+	 * It is the responsibility of code calling this function to ensure proper escaping
+	 * within any HTML.
 	 *
 	 * @since  4.0
 	 *
@@ -419,28 +422,6 @@ class Tribe__Admin__Help_Page {
 	 * @return string
 	 */
 	public function get_content_html( $mixed = '' ) {
-		$html_allowed = array(
-			'a'      => array( 'href' => array(), 'title' => array(), 'target' => array(), 'style' => array(), 'class' => array(), 'id' => array() ),
-			'img'    => array( 'title' => array(), 'src' => array(), 'alt' => array(), 'style' => array(), 'class' => array(), 'id' => array() ),
-
-			'br'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'em'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'strong' => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'b'      => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'i'      => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'u'      => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-
-			'div'    => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'h5'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'p'      => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'ol'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'ul'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'li'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'dl'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'dt'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-			'dd'     => array( 'style' => array(), 'class' => array(), 'id' => array() ),
-		);
-
 		// If it's an StdObj or String it will be converted
 		$mixed = (array) $mixed;
 
@@ -452,8 +433,6 @@ class Tribe__Admin__Help_Page {
 			}
 
 			if ( is_string( $line ) ) {
-				// Make sure we are safe when it's a string
-				$line = wp_kses( $line, $html_allowed );
 				continue;
 			} elseif ( is_array( $line ) ) {
 				// Allow the developer to pass some configuration
@@ -577,7 +556,7 @@ class Tribe__Admin__Help_Page {
 
 			// Setup the Base for the content to come
 			'content' => array(),
- 		);
+		);
 
 		$this->sections[ $section->id ] = $section;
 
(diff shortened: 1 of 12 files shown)


src/Tribe/Log.php
```diff
@@ -0,0 +1,354 @@
...
+	 * available.
+	 *
+	 * @return Tribe__Log__Logger|null
+	 */
+	public function get_current_logger() {
+		if ( ! $this->current_logger ) {
+			$engine = tribe_get_option( 'logging_engine', null );
+			$available = $this->get_logging_engines();
+
+			if ( empty( $engine ) || ! isset( $available[$engine] ) ) {
```
bordoni: Spacing on around `[]`

src/Tribe/Log.php
```diff
@@ -0,0 +1,354 @@
...
+	 * @param $class_name
+	 *
+	 * @return Tribe__Log__Logger|null
+	 */
+	public function get_engine( $class_name ) {
+		if ( ! isset( $this->loggers[ $class_name ] ) ) {
+			$object = new $class_name;
+
+			if ( is_a( $object, 'Tribe__Log__Logger' ) ) {
+				$this->loggers[$class_name] = new $class_name();
```
bordoni: Spacing around `[]`

src/Tribe/Log/Admin.php
```diff
@@ -0,0 +1,249 @@
...
+	 */
+	public function listen() {
+		$fields = wp_parse_args( $_POST, array(
+			'check'      => '',
+			'log-level'  => '',
+			'log-engine' => '',
+		) );
+
+		foreach ( $fields as &$single_field ) {
+			$single_field = filter_var( $single_field, FILTER_SANITIZE_STRING, FILTER_FLAG_ENCODE_HIGH );
```
bordoni: shouldn't we use? [`sanitize_text_field`](https://codex.wordpress.org/Function_Reference/sanitize_text_field)

src/Tribe/Log/Admin.php
```diff
@@ -0,0 +1,249 @@
...
+		catch ( Exception $e ) {
+			// The class name did not relate to a valid logging engine
+		}
+	}
+
+	/**
+	 * Adds a script to handle the event log settings.
+	 */
+	protected function setup_script() {
+		wp_enqueue_script(
```
bordoni: Can we register then enqueue?

src/Tribe/Log/Admin.php
```diff
@@ -0,0 +1,249 @@
...
+		if ( empty( $_GET['tribe-common-log'] ) || 'download' !== $_GET['tribe-common-log'] ) {
+			return;
+		}
+
+		if ( ! wp_verify_nonce( @$_GET['check'], 'download_log' ) ) {
+			return;
+		}
+
+		if ( ! empty( $_GET['log'] ) && in_array( $_GET['log'], $this->get_available_logs() ) ) {
+			$log_name = filter_var( $_GET['log'], FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_HIGH );
```
bordoni: shouldn't we use? [`sanitize_text_field`](https://codex.wordpress.org/Function_Reference/sanitize_text_field)

src/Tribe/Log/Admin.php
```diff
@@ -0,0 +1,249 @@
...
+			$this->current_logger()->use_log( $log_name );
+		}
+
+		header( 'Content-Disposition: attachment; filename="tribe-log-' . $log_name . '"' );
```
bordoni: Should we allow users to filter the final filename?

src/resources/js/admin-log-controls.js
```diff
@@ -0,0 +1,171 @@
...
+	/**
+	 * Add, append or update the log download link so it points to the
+	 * currently selected log file.
+	 */
+	function update_download_link() {
+		var url = $download_link.attr( 'href' );
+		var log = encodeURI( get_current_view() );
+		var matches = url.match(/&log=([a-z0-9\-]+)/i);
+
+		// Update or add the log parameter
```
bordoni: Could we try to re-use this code? Maybe having it as a more global method. https://github.com/moderntribe/tribe-common/blob/develop/src/resources/js/notice-dismiss.js#L5-L20

src/Tribe/Log/Admin.php
barryhughes: We could - what would the advantage be?

src/Tribe/Log/Admin.php
barryhughes: Are there advantages that apply here?

src/resources/js/admin-log-controls.js
barryhughes: Oh nice, I was trying to find some sort of utility function for this - will need to think about how best to position it for access by other scripts.

src/Tribe/Log/Admin.php
barryhughes: We should! Adding...

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Admin.php:225](https://github.com/moderntribe/tribe-common/blob/1b44576625a6b419c4762eb405fd80aa270278f1/src/Tribe/Log/Admin.php#L225)

src/Tribe/Log/Admin.php
barryhughes: I don't think so, but I do see a case for `sanitize_file_name()` - going to change it to that.

src/Tribe/Log/Admin.php
bordoni: That was something that me and @borkweb were talking about past few days as something that would be a easy win for us in terms of having a quick way to enqueue scripts only when they are needed but registering everything on all pages.

src/Tribe/Log/Admin.php
bordoni: Because it would remove empty whitespaces, octets, breaklines and tabs which `FILTER_SANITIZE_STRING, FILTER_FLAG_ENCODE_HIGH` wouldn't

barryhughes: I don't think we need to worry about that in this scenario but there's also no harm in it, will update :+1:

src/resources/js/admin-log-controls.js
barryhughes: Like this idea though I do, my inclination at this stage in the game is to defer it and do it as a follow-on piece of work (for timeliness, but also because it impacts more than just the logging framework).

src/Tribe/Admin/Help_Page.php
```diff
@@ -825,25 +804,25 @@ public function print_plugin_box( $plugin ) {
...
+
+					<dt><?php esc_html_e( 'Rating:', 'tribe-common' ); ?></dt>
+					<dd><a href="<?php echo esc_url( $plugin->stars_url ); ?>" target="_blank">
+							<?php wp_star_rating( array(
+								'rating' => $api_data->rating,
+								'type'   => 'percent',
+								'number' => $api_data->num_ratings,
+							) );?>
+						</a></dd>
+				</dl>
```
lucatume: super spaced here?

barryhughes: Woops - will fix.

src/Tribe/Log.php
```diff
@@ -0,0 +1,354 @@
...
+
+	/**
+	 * Returns the currently active logger or null if none is set/none are
+	 * available.
+	 *
+	 * @return Tribe__Log__Logger|null
+	 */
+	public function get_current_logger() {
+		if ( ! $this->current_logger ) {
+			$engine = tribe_get_option( 'logging_engine', null );
```
lucatume: if this is not a slug but a class name would it not be better to make it clear renaming the var (see above) `$engine_class`?

src/Tribe/Log.php
```diff
@@ -0,0 +1,354 @@
...
+			$available = $this->get_logging_engines();
+
+			if ( empty( $engine ) || ! isset( $available[ $engine ] ) ) {
+				return null;
```
lucatume: would it not be better to return `false` rather than `null`?
I think it's quite more ablative about failure. Here and elsewhere.
Thinking about testing too.

src/Tribe/Log/File_Logger.php
```diff
@@ -0,0 +1,265 @@
...
+		$this->close_handle();
+		$this->obtain_handle();
+	}
+
+	/**
+	 * Attempts to obtain a file handle for the current log file.
+	 */
+	protected function obtain_handle() {
+		$this->close_handle();
+		$this->handle = fopen( $this->log_file, $this->context );
```
lucatume: Are we playing nice with the Filesystem API?

src/Tribe/Log/File_Logger.php
```diff
@@ -0,0 +1,265 @@
...
+	 * @param string $type
+	 * @param string $src
+	 */
+	public function log( $entry, $type = Tribe__Log::DEBUG, $src = '' ) {
+		// Ensure we're in 'append' mode before we try to write
+		if ( 'a' !== $this->context ) {
+			$this->set_context( 'a' );
+		}
+
+		fputcsv( $this->handle, array( date_i18n( 'Y-m-d H:i:s' ), $entry, $type, $src ) );
```
lucatume: A lot of these methods would fail on shared hosting erratically.
We should rely on the Filesystem API if we really want to write any file at all. Both in reading, writing and existence operations.

tests/functional/File_Logging_Test.php
```diff
@@ -0,0 +1,174 @@
+<?php
+namespace Tribe\Common;
+
+use Codeception\TestCase\WPTestCase,
+    Tribe__Log,
+    Tribe__Log__Logger,
+    Tribe__Main;
+
+class Logging_Test extends WPTestCase {
```
lucatume: This, along with the file based logger, should use WP Filesystem API and mock it.
Either "manually" or using something like [a mock of it](https://github.com/JDGrimes/wp-filesystem-mock).

src/Tribe/Log.php
barryhughes: I like that.

src/Tribe/Log/File_Logger.php
barryhughes: I'm inclined to disagree, but I might also just be being silly.

I feel like the Filesystem API's chief benefit is when there's a need to change or add files within the WP installation - new images, themes/core/plugin updates, etc. These changes that it is geared towards are also not typically changes that happen multiple times per request.

So I'm a little reluctant to use an API that:

* Might be configured to use something massively slower than direct writing to the file system (such as FTP)
* In direct file mode isn't taking the most efficient path in a logging context (we can't efficiently append, so far as I can see)
* Where we'd have the spectre of file system credentials expiring when we're logging outside of an admin request (we could proactively manage this in the UI where logging is enabled, but it a) wouldn't be bulletproof and b) would add some complexity)

What do you reckon - are these non-issues?

tests/functional/File_Logging_Test.php
barryhughes: If we switch, definitely - good call! Let's hash out that decision (about switching to the FS API) some more first.

src/Tribe/Log/File_Logger.php
lucatume: What you listed is all true: those are issues indeed while dealing with the FS API.
Since we are assuming a direct access to the fs or fail I'd try to write to the `uploads` folder (many shared hosting leave that writeable internally) and/or fallback to a database based logging.
Even in a fully controlled context as service projects I've seen the latter approach taken using a CPT (log) with one day lifespan and where log entries are comments appended to it. You would also reap all the benefits of posts (search, index, access...)

barryhughes:
Some great ideas in the feedback here but after thinking this one through, we most likely will not create alternative logging implementations (that use either the WP Filesystem API or CPTs for logging) - at least not in the first iteration - though there's nothing to prevent us rolling those out in later releases or even on the fly to meet support needs as they arise if we need them.

The FS API and CPTs are both supremely WordPressy structures, but neither were created with logging in mind and are less than optimal for that purpose. The current strategy (writing to `/tmp` or equivalent in append mode) is essentially the same as WooCommerce and other notable plugins use, so I have a high degree of confidence that it will work in a large number of environments.

bordoni:
Based on the latest comment from @barryhughes I think it's good for a merge on this First Iteration..

PR #90, opened 2016-05-06
Disable logging | #36047

barryhughes:
Per discussion in linked issue, we'll keep the logging code in place but effectively disable it until we add code elsewhere that can make use of it.

[C#36047](https://central.tri.be/issues/36047)


bordoni:
🎱

PR #91, opened 2016-05-09
Delete error msgs once displayed, not each shutdown | #46544

barryhughes:
Delete stored errors after displaying them rather than each time `shutdown` fires (otherwise they are lost if we add them immediately before a redirect).

https://central.tri.be/issues/46544


bordoni:
👍

PR #92, opened 2016-05-12
Prevent notices when TEC is not active

bordoni:
_Rel:_ moderntribe/event-tickets#182


barryhughes:
(Ignoring the ticket-ref-check because this is a spotfix.)

:+1:

PR #97, opened 2016-06-09
Better Notice Handling

bordoni:
As I was working on the Ignored Events for Event Aggregator, I noticed that our structure to deliver notices was lacking a few features and it was duplicated in a few places.

_Rel:_ moderntribe/events-pro#334
_Ref:_ [C#61231](https://central.tri.be/issues/61231)

PR #99, opened 2016-06-12
#62408 | Tweak: Have the Slug Conflict Notice Use the Real Post Type

ggwicz:
This tweak makes it so that if a post type _other than_ a page is made whose slug conflicts with The Events Calendar's events slug option, the "conflict" notice reports the conflict as coming from that specific post type—instead of just saying "page" no matter the culprit.

Below is an example of this working when the conflicting item is a media item—which is, by default, the only post type that should show up other than "page", due to the way `get_page_by_path()` works.

_Ref:_ [C#62408](https://central.tri.be/issues/62408)

--

Screenshot

### DIFF
--- a/src/Tribe/Admin/Notice/Archive_Slug_Conflict.php
+++ b/src/Tribe/Admin/Notice/Archive_Slug_Conflict.php
@@ -80,13 +80,15 @@ class Tribe__Admin__Notice__Archive_Slug_Conflict {
 	 */
 	public function notice() {
 		// What's happening?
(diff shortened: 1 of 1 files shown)


tr1b0t:

ggwicz:
Is this idea essentially invalid because of the changes by Gustavo in #97?

bordoni:
Not really, I will be able to port this to the structure after.

bordoni:
This will be fixed when #97 gets merged.

PR #101, opened 2016-06-13
fix the links to got to edit.php everywhere

zbtirrell:
See: https://central.tri.be/issues/61823

### DIFF
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -607,13 +607,13 @@ if ( ! class_exists( 'Tribe__Settings' ) ) {
 		public function get_url( array $args = array() ) {
 			$defaults = array(
 				'page' => $this->adminSlug,
+				'parent' => self::$parent_page,
 			);
 
 			// Allow the link to be "changed" on the fly
 			$args = wp_parse_args( $args, $defaults );
-			$parent = self::$parent_page;
 
-			$url = admin_url( $parent );
+			$url = admin_url( $args['parent'] );
 
 			return apply_filters( 'tribe_settings_url', add_query_arg( $args, $url ), $args, $url );
 		}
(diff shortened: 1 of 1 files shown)


tr1b0t:

src/Tribe/Settings.php
```diff
@@ -613,10 +613,6 @@ public function get_url( array $args = array() ) {
 			$args = wp_parse_args( $args, $defaults );
 			$parent = self::$parent_page;
 
-			if ( ! is_admin() ) {
```
barryhughes: Just asking to be sure we're covered: but removing this won't cause a regression re [PR 95](https://github.com/moderntribe/tribe-common/pull/95) / [issue #46198](https://central.tri.be/issues/46198)?

bordoni: It's a problem for sure. Good Catch @barryhughes

zbtirrell:
Good catches Barry/Gustavo. I have made a change to Event Tickets to make this less fragile.

src/Tribe/Settings.php
```diff
@@ -607,13 +607,13 @@ public function deleteOptions() {
...
 				'page' => $this->adminSlug,
+				'parent' => self::$parent_page,
 			);
 
 			// Allow the link to be "changed" on the fly
 			$args = wp_parse_args( $args, $defaults );
-			$parent = self::$parent_page;
 
-			$url = admin_url( $parent );
+			$url = admin_url( $args['parent'] );
```
barryhughes: Wondering just a couple of things here:

1. Should we unset `$args['parent']` after this call to keep the query args of the resulting URL nice and clean?
2. Are we sure there's no problem if a URL generated with a non-default `$args['parent']` is used (given the submenu page is always registered using `self::$parent_page` - line 212)?

For that 2nd point in particular I'm happy if you've checked this out, really just wanted to make sure it's been covered (as it would _seem_ on initial reading that the result would be we'd end up trying to access a submenu page that hasn't actually been registered).

zbtirrell: Great point on 1. I fell relatively confident on the 2nd point and have tested Community Tickets as well as noted that it should be QA'd in the ticket.

PR #102, opened 2016-06-14
Bug/61672 show updated settings

jesseeproductions:
_Ref:_ [C#61672](https://central.tri.be/issues/61672)
_Ref:_ [C#61739](https://central.tri.be/issues/61739)

Always delete WordPress Option so that Add Option can save


bordoni:
@barryhughes  I think did the changes on these saving form errors, so I think he should take a look.

barryhughes:
:racehorse:

PR #103, opened 2016-06-15
make menu position match with The Events Calendar

zbtirrell:
When you have just ET activated, it's menu position should be the same as when you have The Events Calendar active.


barryhughes:
:+1:

PR #104, opened 2016-06-15
Fix/46581 date utils

barryhughes:
Adds the static `wp_strtotime()` method to `Tribe__Date_Utils` (it's a straight lift almost from core's ical class) - this will allow us to use it from various other locations rather than duplicating it across multiple plugins and classes.

https://central.tri.be/issues/46581


jesseeproductions:
:+1:

PR #105, opened 2016-06-15
add ability for settings fields to have placeholder text

zbtirrell:
This is in support of https://github.com/moderntribe/events-community/pull/128

See: https://central.tri.be/issues/28029

### DIFF
--- a/src/Tribe/Field.php
+++ b/src/Tribe/Field.php
@@ -345,7 +345,7 @@ if ( ! class_exists( 'Tribe__Field' ) ) {
 		 *
 		 * @return string the field value
 		 */
-		public function doFieldPlaceholder() {
+		public function do_field_placeholder() {
 			$return = '';
 			if ( $this->placeholder ) {
 				$return = ' placeholder="' . $this->placeholder . '"';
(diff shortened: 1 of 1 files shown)


src/Tribe/Field.php
```diff
@@ -81,7 +82,7 @@ public function __construct( $id, $field, $value = null ) {
 				'clear_after'      => true,
 			);
 
-			// a list of valid field types, to prevent screwy behaviour
+			// a list of valid field types, to prevent screwy behavior
```
barryhughes: Sad to see this change but ok.

src/Tribe/Field.php
```diff
@@ -340,6 +341,20 @@ public function doFieldName( $multi = false ) {
 		}
 
 		/**
+		 * returns the field's placeholder
+		 *
+		 * @return string the field value
+		 */
+		public function doFieldPlaceholder() {
```
barryhughes: Looks like this is a brand new method ... with that in mind can we use snake_case?

zbtirrell: While I certainly like snake case better, I was making it match all it's neighbors. @borkweb what is your preference here?

borkweb: I prefer snake_case. While there may be merit in matching the neighbors for consistency, I _do_ like the idea of moving forward to what we actually want. Since this is for the current release, let's make this leap:

1. Use snake_case for this new method
2. Deprecate all the other camelCase methods and convert them to snake_case.

borkweb: I'll make a ticket for the method deprecation so it can be done in a different PR.

borkweb: Here's the ticket: https://central.tri.be/issues/62572

borkweb:
🍔

PR #106, opened 2016-06-16
Restore normal behaviour of Tribe__Settings::get_url() | #61749

barryhughes:
This fixes the problem described in [#61749](https://central.tri.be/issues/61749) however it undoes part of @bordoni's fix for [#46189](https://central.tri.be/issues/46198) [(PR 95)](https://github.com/moderntribe/tribe-common/pull/95) so we need to be careful here.

https://central.tri.be/issues/61749

### DIFF
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -24,7 +24,7 @@ if ( ! class_exists( 'Tribe__Settings' ) ) {
 		 * Page of the parent menu
 		 * @var string
 		 */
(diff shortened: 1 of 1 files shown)


barryhughes:
Actually ... I was wrong (re my original notes on this breaking the attendee modal), which is good news. Zach's work [here](https://github.com/moderntribe/tribe-common/pull/101) has us on the right path. This _should_ be safe to merge.

PR #110, opened 2016-06-20
Is_embed move method remove superfluous else

jesseeproductions:
_Ref:_ [C#62577](https://central.tri.be/issues/62577) - method in common for #62577

Miss a note about the superfluous else and removing it here.


zbtirrell:
👍

PR #111, opened 2016-06-21
Assets Handling

bordoni:
_Ref:_ [C#62321](https://central.tri.be/issues/62321)

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -61,7 +61,7 @@ class Tribe__Assets {
 				continue;
 			}
 
-			// Now add an action to enqueue the registred assets
+			// Now add an action to enqueue the registered assets
 			add_action( $asset->action, array( $this, 'enqueue' ), $asset->priority );
 		}
 	}
@@ -122,33 +122,59 @@ class Tribe__Assets {
 	 * @return string|false The path/url to minified version or false, if file not found.
 	 */
 	public static function maybe_get_min_file( $url ) {
+		$urls = array();
+		// If need add the Min Files
 		if ( ! defined( 'SCRIPT_DEBUG' ) || SCRIPT_DEBUG === false ) {
 			if ( substr( $url, - 3, 3 ) === '.js' ) {
-				$url = substr_replace( $url, '.min', - 3, 0 );
+				$urls[] = substr_replace( $url, '.min', - 3, 0 );
 			}
 
 			if ( substr( $url, - 4, 4 ) === '.css' ) {
-				$url = substr_replace( $url, '.min', - 4, 0 );
+				$urls[] = substr_replace( $url, '.min', - 4, 0 );
 			}
 		}
 
-		$file = str_replace( WP_CONTENT_URL, WP_CONTENT_DIR, $url );
-		if ( file_exists( $file ) ) {
-			return $url;
-		} else {
-			return false;
+		// Add the actual url after having the Min file added
+		$urls[] = $url;
+
+		// Check for all Urls added to the array
+		foreach ( $urls as $key => $url ) {
+			$file = str_replace( WP_CONTENT_URL, WP_CONTENT_DIR, $url );
+
+			if ( file_exists( $file ) ) {
+				return $url;
+			}
 		}
+
+		// If we don't have any real file return false
+		return false;
 	}
 
 	/**
 	 * Register an Asset and attach a callback to the required action to display it correctly
 	 *
-	 * @param  object   $origin    The main Object for the plugin you are enqueueing the script/style for
-	 * @param  string   $slug      Slug to save the asset
-	 * @param  string   $file      Which file will be loaded, either CSS or JS
-	 * @param  array    $deps      Dependencies
-	 * @param  string   $action    A WordPress Action, needs to happen after: `wp_enqueue_scripts`, `admin_enqueue_scripts`, or `login_enqueue_scripts`
-	 * @param  array    $arguments Arguments to setup the asset
+	 * @param  object       $origin    The main Object for the plugin you are enqueueing the script/style for
+	 * @param  string       $slug      Slug to save the asset
+	 * @param  string       $file      Which file will be loaded, either CSS or JS
+	 * @param  array        $deps      Dependencies
+	 * @param  string|null  $action    (Optional) A WordPress Action, if set needs to happen after: `wp_enqueue_scripts`, `admin_enqueue_scripts`, or `login_enqueue_scripts`
+	 * @param  string|array $query {
+	 *     Optional. Array or string of parameters for this asset
+	 *
+	 *     @type string|null  $action         Which WordPress action this asset will be loaded on
+	 *     @type int          $priority       Priority in which this asset will be loaded on the WordPress action
+	 *     @type string       $file           The relative path to the File that will be enqueued, uses the $origin to get the full path
+	 *     @type string       $type           Asset Type, `js` or `css`
+	 *     @type array        $deps           An array of other asset as dependencies
+	 *     @type string       $version        Version number, used for cache expiring
+	 *     @type string       $media          Used only for CSS, when to load the file
+	 *     @type bool         $in_footer      A boolean determining if the javascript should be loaded on the footer
+	 *     @type array|object $localize       Variables needed on the JavaScript side {
+	 *          @type string 		$name     Name of the JS variable
+	 *          @type string|array  $data     Contents of the JS variable
+	 *     }
+	 *     @type callable[]   $filter         An callable method or an array of them, that will determine if the asset is loaded or not
+	 * }
 	 *
 	 * @return string
 	 */
@@ -192,7 +218,7 @@ class Tribe__Assets {
 			'media'     => 'all',
 			'in_footer' => true,
 			'localize'  => array(),
-			'filter'   => array(),
+			'filter'    => array(),
 		);
 
 		// Merge Arguments
@@ -222,6 +248,9 @@ class Tribe__Assets {
 
 		/**
 		 * Deprecated filter to allow changing version based on the type of Asset
+		 *
+		 * @todo remove on 4.6
+		 * @deprecated 4.3
 		 *
 		 * @param string $version
 		 */
@@ -288,7 +317,7 @@ class Tribe__Assets {
 	}
 
 	/**
-	 * Removes an Asset from been registred and enqueue
+	 * Removes an Asset from been registered and enqueue
 	 *
 	 * @param  string $slug Slug of the Asset
 	 *
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -33,9 +33,7 @@ class Tribe__Assets {
 	 */
 	private function __construct() {
 		// Hook the actual rendering of notices
-		add_action( 'wp_enqueue_scripts', array( $this, 'hook' ), 1 );
-		add_action( 'admin_enqueue_scripts', array( $this, 'hook' ), 1 );
-		add_action( 'login_enqueue_scripts', array( $this, 'hook' ), 1 );
+		add_action( 'init', array( $this, 'hook' ), 1 );
 	}
 
 	public function hook() {
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -33,11 +33,19 @@ class Tribe__Assets {
 	 */
 	private function __construct() {
 		// Hook the actual rendering of notices
-		add_action( 'init', array( $this, 'hook' ) );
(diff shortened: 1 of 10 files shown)


src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+			return false;
+		}
+	}
+
+	/**
+	 * Stores all the Assets and it's configurations
+	 *
+	 * @var array
+	 */
+	private $assets = array();
```
borkweb: Why declare this so low?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+				continue;
+			}
+
+			// If here is no action we are just registering this asset
+			if ( ! is_string( $asset->action ) ) {
+				continue;
+			}
+
+			// If we have a set of conditionals we loop on then and get if they are true
+			if ( ! empty( $asset->conditionals ) ) {
```
borkweb: We discussed via Hangouts, but just to be thorough: Let's ditch the conditional stuff in favor of an `apply_filters()` call for a boolean, here.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+
+	/**
+	 * Stores all the Assets and it's configurations
+	 *
+	 * @var array
+	 */
+	private $assets = array();
+
+	/**
+	 * Register a Asset and attach a callback to the required action to display it correctly
```
borkweb: Maybe: `Register an asset` rather than `Register a Asset`

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+	 *
+	 * @param  string   $slug      Slug to save the asset
```
borkweb: Missing `@param string $origin yadda yadda`

borkweb: And `$file`, `$deps`, `$action`. Extra `$callback`. Now that I think about it, I think this is just the wrong docblock :)

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+
+			if ( ! defined( $origin_name . '::VERSION' ) ) {
+				// If we have a Object and we don't have instance or version
+				return false;
+			}
+		} else {
+			return false;
+		}
+
+		// Uses Common by default
```
borkweb: Is this comment true?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+			'action'       => null,
+			'priority'     => 10,
+			'file'         => false,
+			'type'         => null,
+			'deps'         => array(),
+			'version'      => $version,
+			'media'        => 'all',
+			'in_footer'    => true,
+			'localize'     => array(),
+			'conditionals' => array(),
```
borkweb: `'conditionals => array(),` won't be needed with the filter stuff.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+		$asset = apply_filters( 'tribe_asset', $asset );
+
+		// Set the Asset on the array of notices
+		$this->assets[ $slug ] = $asset;
+
+		// Return the Slug because it might be modified
+		return $asset;
+	}
+
+	public function remove( $slug ) {
```
borkweb: Docblock?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+		if ( ! $this->exists( $slug ) ) {
+			return false;
+		}
+
+		unset( $this->assets[ $slug ] );
+		return true;
+	}
+
+	public function get( $slug = null ) {
```
borkweb: Docblock?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,286 @@
...
+		}
+
+		if ( ! empty( $this->assets[ $slug ] ) ) {
+			return $this->assets[ $slug ];
+		}
+
+		return null;
+	}
+
+	public function exists( $slug ) {
```
borkweb: Docblock?

src/Tribe/Main.php
```diff
@@ -98,81 +98,39 @@ public function init_libraries() {
...
-		);
-
-		wp_register_style(
-			'tribe-jquery-ui-datepicker',
-			$vendor_base . '/jquery/ui.datepicker.css',
-			array( 'tribe-jquery-ui-theme' ),
-			apply_filters( 'tribe_events_css_version', self::VERSION )
-		);
-
+	public function load_assets() {
```
borkweb: I think `register_resources()` is still an appropriate (and arguably better) method name.

src/Tribe/Main.php
```diff
@@ -98,81 +98,39 @@ public function init_libraries() {
...
+		// These ones are only registred
+		tribe_assets( $this, array(
+			array( 'ba-dotimeout', 'jquery.ba-dotimeout.js', array( 'jquery' ) ),
+			array( 'tribe-jquery-ui-theme', 'vendor/jquery/ui.theme.css' ),
+		) );
```
borkweb: The formatting of this `tribe_assets()` call is kind of hard to read. Maybe:

```php
tribe_assets( $this, array(
  array(
    'ba-dotimeout',
    'jquery.ba-dotimeout.js',
    array( 'jquery' ),
  ),
  array(
    'tribe-jquery-ui-theme',
    'vendor/jquery/ui.theme.css'
  ),
) );
```

Though maybe it'd be easier still with two distinct `tribe_asset()` calls:

```php
tribe_asset( $this, 'ba-dotimeout', 'jquery.ba-dotimeout.js', array( 'jquery' ) );
tribe_asset( $this, 'tribe-jquery-ui-theme', 'vendor/jquery/ui.theme.css' );
```

Rinse and repeat this comment for the next `tribe_assets()` call.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Each line in an array declaration must end in a comma | [Main.php:119](https://github.com/moderntribe/tribe-common/blob/166ed104956dd9bdc41debc72bafc0928f20c22c/src/Tribe/Main.php#L119)

src/functions/template-tags/general.php
```diff
@@ -454,3 +464,29 @@ function tribe_get_date_option( $optionName, $default = '' ) {
 		return Tribe__Date_Utils::unescape_date_format($value);
 	}
 }
+
+function tribe_asset( $origin, $slug, $file, $deps = array(), $action = null, $arguments = array() ) {
```
borkweb: Docblock here?

src/functions/template-tags/general.php
```diff
@@ -454,3 +464,29 @@ function tribe_get_date_option( $optionName, $default = '' ) {
...
+	return Tribe__Assets::instance()->register( $origin, $slug, $file, $deps, $action, $arguments );
+}
+
+function tribe_assets( $origin, $assets, $action = null, $arguments = array() ) {
```
borkweb: Docblock here?

borkweb:
This looks pretty good but I'd love for @barryhughes to weigh in on it as well since it is a pretty big change to how we'd do stuff going forward.

src/functions/template-tags/general.php
bordoni: Good catch.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,338 @@
...
+
+	/**
+	 * Returns the path to a minified version of a js or css file, if it exists.
+	 * If the file does not exist, returns false.
+	 *
+	 * @param string $url   The path or URL to the un-minified file.
+	 *
+	 * @return string|false The path/url to minified version or false, if file not found.
+	 */
+	public static function maybe_get_min_file( $url ) {
```
barryhughes: Looking at the way this method is used (from within the `register()` method for example), I'm wondering if it would make sense to make this more tolerant of missing minified assets.

For instance, right now if `abc.js` exists but `abc.min.js` does not, it will return false: looking back to that `register()` method that would mean we'd be setting `$asset->url` to false, when actually there might be a valid URL we could be using.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,338 @@
...
+		if ( ! in_array( $asset->type, array( 'js', 'css' ) ) ) {
+			return false;
+		}
+
+		/**
+		 * Deprecated filter to allow changing version based on the type of Asset
+		 *
+		 * @param string $version
+		 */
+		$asset->version = apply_filters( "tribe_events_{$asset->type}_version", $asset->version );
```
barryhughes: Add a `@deprecated` tag for 4.3 (?)

src/functions/template-tags/general.php
```diff
@@ -454,3 +464,29 @@ function tribe_get_date_option( $optionName, $default = '' ) {
...
+	$registred = array();
```
barryhughes: `$registered` (typo)

src/Tribe/Assets.php
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	/**
+	 * Register an Asset and attach a callback to the required action to display it correctly
+	 *
+	 * @param  object   $origin    The main Object for the plugin you are enqueueing the script/style for
+	 * @param  string   $slug      Slug to save the asset
+	 * @param  string   $file      Which file will be loaded, either CSS or JS
+	 * @param  array    $deps      Dependencies
+	 * @param  string   $action    A WordPress Action, needs to happen after: `wp_enqueue_scripts`, `admin_enqueue_scripts`, or `login_enqueue_scripts`
```
barryhughes: It took me a few readings to get a sense of what we're doing here: wonder if it might help others to expand on the role of `$action` a little. For instance, cover if it must be set to a valid string (right now the description is suggesting I should supply a hook name, but the signature suggests it's optional).

barryhughes:
I like this change, I think it will move us in the right direction :+1:

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task | [Assets.php:252](https://github.com/moderntribe/tribe-common/blob/470e2140922eeda579d78e8b1b10e677d2a514da/src/Tribe/Assets.php#L252)

borkweb:
🐘 💨

src/Tribe/Assets.php
```diff
@@ -0,0 +1,367 @@
...
+		// If we have any type of input on Filter we hook it
+		if ( ! empty( $asset->filter ) ) {
+			// This allows us to
+			if ( is_array( $asset->filter ) && is_callable( $asset->filter ) ) {
+				$asset->filter = array( $asset->filter );
+			}
+
+			// To make this work we will use the 0 as the number of arguments
+			foreach ( $asset->filter as $filter ) {
+				add_filter( 'tribe_asset_enqueue_' . $asset->slug, $filter, 10, 0 );
```
barryhughes: Hey @bordoni! Had some more thoughts on this now I've seen it working in practice.

All those filters will be enqueued on the same (default) priority, so the last filter added - whichever that happens to be - can potentially override all the rest. For example, say I want an asset to be enqueued on post type screens _or_ on the community submission page, I might want to add multiple filters like so:

```
[
    [ Tribe__Admin__Helpers::instance(), 'is_post_type_screen' ],
    'tribe_is_community_edit_event_page'
]
```

Right now though, it seems like this would have the effect of only enqueuing the asset on the frontend submission page (because the result of the last filter function overrides the first) ... might some sort of _or_ type logic work better for cases where multiple filters are added?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,367 @@
...
+			 * @param object $asset   Which asset we are dealing with
+			 */
+			$enqueue = apply_filters( 'tribe_asset_enqueue_' . $asset->slug, $enqueue, $asset );
+
+			if ( ! $enqueue ) {
+				continue;
+			}
+
+			if ( 'js' === $asset->type ) {
+				wp_enqueue_script( $asset->slug );
```
barryhughes: Since we're only passing the asset slug to `wp_enqueue_script` here this will fail for any assets registered later than `init` (ie, when `action_maybe_register` runs), but that limits our ability to be lazy and register stuff on demand - anything set up with a `tribe_asset()` call later than this point won't make it to the browser.

What do you think about re-running `wp_register_script()` here? We could possibly track which have already been registered and only do this as needed.

Ditto for registration of styles, below.

bordoni: Good catch here, I think this is really important. 👍

src/Tribe/Assets.php
bordoni: Very true, I would love to hope on call to discuss this problem further, hard to solve this problem.

PR #113, opened 2016-06-23
Switch to call_user_func() | spotfix

barryhughes:
Switch from `call_user_func_array()` to plain old `call_user_func()` (we're not passing any params).


borkweb:
:+1:

PR #115, opened 2016-06-28
Make the file paths for Assets handling OS aware

bordoni:


barryhughes:
Though it's a spotfix, I think it could be worth a changelog entry. Seems like a positive change we should highlight.

bordoni:
Talked with @barryhughes on Slack and he gave me the Ok to merge it.

PR #116, opened 2016-06-28
load common translations

zbtirrell:
`Tribe_Log` was trying to get translated strings earlier than they were registered. This combined with the new `load_text_domain` checking that they had been loaded without confirming they had loaded properly led to any strings in common not being translated.

See: https://central.tri.be/issues/62562


barryhughes:
You got to the bottom of it! :8ball:

PR #130, opened 2016-07-08
Improve and add a `undismiss` method for the `Admin_Notices` class

bordoni:
_Ref:_ [C#62408](https://central.tri.be/issues/62408)


barryhughes:
:100:

PR #133, opened 2016-07-13
hacky solution to limit ratings request to only relevant post types

zbtirrell:
See: https://central.tri.be/issues/63557#note-6


borkweb:
:+1:

PR #134, opened 2016-07-14
Feature/jsonls avoid duplicate prints

lucatume:
Ticket: https://central.tri.be/issues/62439

Modifies the JSON-LD abstract class to avoid returning data for already fetched posts.
This is to avoid re-printing JSON-LS data for the same post multiple times on the page (no-no in Google standards)


src/Tribe/JSON_LD/Abstract.php
```diff
@@ -42,27 +49,47 @@ public static function instance( $name = null ) {
 
 	/**
 	 * Compile the schema.org event data into an array
+	 *
+	 * @param mixed $post            Either a post ID or a WP_post object.
+	 * @param array $args            {
+	 *                               Optional. An array of arguments to control the returned data.
```
barryhughes: This looks just slightly weird. Can we format per [this example](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays) (ie, indent the array args)?

src/Tribe/JSON_LD/Abstract.php
```diff
@@ -126,18 +154,79 @@ public function markup( $post = null, $args = array() ) {
...
+
+		self::$fetched_post_ids[] = $post->ID;
+	}
+
+	public function reset_fetched_post_ids() {
+		self::class_reset_fetched_post_ids();
+	}
+
+	public function unset_fetched_post_id( $post ) {
+		$post = $this->get_post_object( $post );
```
barryhughes: We're really only interested in getting the ID here: is it possible to drop the `get_post_object()` method and call `Tribe__Main::post_id_helper()` directly?

tests/wpunit/Tribe/JSON_LD/AbstractTest.php
```diff
@@ -0,0 +1,176 @@
+<?php
+namespace Tribe\JSON_LD;
+
+require_once codecept_data_dir( 'classes/Tribe__JSON_LD__Test_Class.php' );
+
+use Tribe__JSON_LD__Test_Class as Jsonld;
+
+class AbstractTest extends \Codeception\TestCase\WPTestCase {
```
barryhughes: Maybe switch to `Snake_Case` here?

lucatume: We have this format throughout all of the test code, Codeception will generate class names in he `Whatever_Class_Name__IsTest` format. To have it like that we should update all of the `codeception.yml` and tests to stick.

barryhughes: OK, just feels weird to not follow our own coding standards for tests. `Tribe__Template_Checker_Test` and `Logging_Test` aren't camelcased but seem to work ok?

barryhughes:
:fireworks:

PR #136, opened 2016-07-15
Fixing the Tribe__Main::plugin_url value for common

borkweb:
This has been incorrect since 4.0, apparently. It is finally needed for another ticket.

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -43,7 +43,10 @@ class Tribe__Main {
 
 		$this->plugin_path = trailingslashit( dirname( dirname( dirname( __FILE__ ) ) ) );
 		$this->plugin_dir  = trailingslashit( basename( $this->plugin_path ) );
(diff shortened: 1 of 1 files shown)


src/Tribe/Main.php
```diff
@@ -105,8 +108,8 @@ public function load_assets() {
 		tribe_assets(
 			$this,
 			array(
-				array( 'tribe-jquery-ui-theme', 'vendor/jquery/ui.theme.css' ),
-				array( 'tribe-clipboard', 'vendor/clipboard/clipboard.js' ),
+				array( 'tribe-jquery-ui-theme', $this->plugin_url . 'vendor/jquery/ui.theme.css' ),
```
bordoni: Wait what? We shouldn't need to pass the whole URL, the Assets class should be able to handle.

borkweb: Whoops! That was actually from me debugging something and not meant to be in this PR. Good catch.

bordoni:
🎱

PR #138, opened 2016-07-19
Feature/43709 clipboard js safari

jesseeproductions:
_Ref:_ [C#43709](https://central.tri.be/issues/43709) Add some messaging to help with browsers that are unable to automatically copy.


src/resources/js/tribe-common.js
```diff
@@ -17,13 +17,32 @@ tribe_auto_sysinfo.ajax = {
 	 */
 	my.init_copy = function () {
 
-		new Clipboard( '.system-info-copy-btn' );
+		var clipboard = new Clipboard( '.system-info-copy-btn' );
```
barryhughes: Just a thought, no need to change if you don't want to, but we could also write this block in this kind of style:

```
var clipboard   = whatever,
    button_icon = something else,
    button_text = etc;
```

Saves a bit of space, but still makes for clean reading. Admittedly a matter of personal preferences though :smiley:

src/resources/js/tribe-common.js
```diff
@@ -17,13 +17,32 @@ tribe_auto_sysinfo.ajax = {
...
+		var button_icon = '<span class="dashicons dashicons-clipboard license-btn"></span>';
+
+		var button_text = tribe_system_info.clipboard_btn_text;
 
 		//Prevent Button From Doing Anything Else
 		$( ".system-info-copy-btn" ).click( function ( e ) {
 			e.preventDefault();
 		} );
 
+		clipboard.on('success', function(event) {
```
barryhughes: Here are throughout this change - we should have spaces before and after opening parens. So for instance this line should be `clipboard.on( 'success', function( event ) {` and so on for the following lines.

barryhughes:
Sorry Brian, I totally missed you'd applied the `hold` tag.

barryhughes:
:+1:

PR #141, opened 2016-07-20
Ensure the jQuery UI theme/datepicker styles are available in the event editor

barryhughes:
Fixes the unstyled datepicker problem.

[#63135](https://central.tri.be/issues/63135)


borkweb:
:cabbage:

PR #142, opened 2016-07-27
Adding datatables jquery Library

borkweb:
This library is needed for Event Aggregator but will most likely have uses elsewhere.


src/Tribe/Main.php
```diff
@@ -139,6 +149,44 @@ public function load_assets() {
...
+			'tribe-common',
+			'tribe-common.js',
+			array( 'tribe-clipboard' ),
+			'admin_enqueue_scripts',
+			array(
+				'localize' => array(
+					'name' => 'tribe_l10n_datatables',
+					'data' => array(
+						'aria' => array(
+							'sort_ascending' => __( ': activate to sort column ascending', 'the-events-calendar' ),
```
barryhughes: Here and the next few lines: should we switch text domain (as this code is in tribe-common rather than the-events-calendar)?

src/resources/postcss/datatables.pcss
```diff
@@ -0,0 +1,435 @@
+/*
+ * Table styles
+ */
+table.dataTable {
+  border-collapse: separate;
```
barryhughes: Spacing is off a little here - should be a full tab instead of two spaces (here and the next few lines).

src/resources/postcss/datatables.pcss
```diff
@@ -0,0 +1,435 @@
...
+		}
+	}
+
+	.check-column {
+		width: 2.2em;
+	}
+
+	&.row-border tbody th, &.row-border tbody td, &.display tbody th, &.display tbody td {
+		border-top: 1px solid #ddd;
+	}
```
barryhughes: Maybe add newlines after the closing `}` (throughout this file).

barryhughes:
I could be missing it (GH isn't showing every changed file since it's such a large diff) but can you confirm `package-whitelist.json` is covered re these changes?

src/resources/postcss/datatables.pcss
borkweb: Done :) I didn't clean up the copied css file as much as I should have. Good catch.

borkweb:
I had the relevant files added in the-events-calendar's `package-whitelist.json` but not tribe-common's. Added!

barryhughes:
#jeSuisMattsCode

PR #143, opened 2016-07-27
Usability fixes for bumpdowns and datatables in support of Event Aggregator

borkweb:
These issues came out of work for Event Aggregator

* Fixing wonky behavior in bumpdowns where the bumpdown would close on click if the bumpdown was in middle of expanding from a hoverIntent
* Added the auto-closing of bumpdowns if their corresponding `data-depends` field is changed
* Added support for refreshing table contents of the tribe-datatables


src/resources/js/bumpdown.js
```diff
@@ -205,6 +206,14 @@
 
 				// Mark it as the Bumpdown
 				.addClass( selectors.bumpdown.replace( '.', '' ) );
+
+			// support our dependency library
+			if ( data.$trigger.data( 'depends' ) ) {
+				var field_ids = data.$trigger.data( 'depends' );// + ',' + data.$trigger.data( 'depends' ).replace( '#', '#s2id_' );
```
barryhughes: Just checking -- do we want to keep the comment at the end of this line?

barryhughes:
Good to merge, optionally might want to remove a comment or not at your option :tophat:

src/resources/js/bumpdown.js
borkweb: Whee. Nope!

PR #144, opened 2016-07-29
making App Shop local

zbtirrell:
Instead of callign out to TEC.com, load the App Shop from the local site.

Also, optimized the images in the resources/images folder

See: https://central.tri.be/issues/64286


borkweb:
:+1:

PR #155, opened 2016-08-10
System Info fix PHP Strict Standards static method warning

jesseeproductions:
_Ref:_ [C#38137](https://central.tri.be/issues/38137)


bordoni:
👍

PR #159, opened 2016-08-11
Remove duplicate tribe-common js enqueue

borkweb:
We're doing tribe-common below in order to localize the datatable strings.


bordoni:
👍

PR #167, opened 2016-08-22
minor text updates, lots of reorganization

zbtirrell:
Reorganized so the translatable strings do not contain any HTML and are in smaller chunks.

### DIFF
--- a/src/admin-views/tribe-options-licenses.php
+++ b/src/admin-views/tribe-options-licenses.php
@@ -7,10 +7,29 @@ $link = add_query_arg(
 	), Tribe__Main::$tec_url . 'license-keys/'
 );
 
-$link = esc_url( $link );
+$utm_link = esc_url( $link );
 
 // Explanatory text about license settings for the tab information box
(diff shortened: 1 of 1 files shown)


src/admin-views/tribe-options-licenses.php
```diff
@@ -7,10 +7,29 @@
...
+$html = '<p>' . sprintf(
+		esc_html__( 'The license key you received when completing your purchase from %1$s will grant you access to support and updates until it expires. You do not need to enter the key below for the plugins to work, but you will need to enter it to get automatic updates. %3$sFind your license keys at %2$s%4$s.', 'tribe-common' ),
+		'<a href="' . Tribe__Main::$tec_url . '" target="_blank">' . Tribe__Main::$tec_url . '</a>',
```
borkweb: I wonder if we should take this opportunity to add the `'<span class="screen-reader-text">' . __( ' (opens in new window)', 'tribe-common' ) . '</span>'` on all of these links with `target="_blank"` since they are being fiddled with now?

borkweb:
If you don't want to tackle adding the span on those links, then this is :+1:

zbtirrell:
I added it in, take a quick look again to make sure I did it right.

borkweb:
Glorious. Merge it :shipit:

PR #172, opened 2016-08-26
fix/check-write-ability

lucatume:
Ticket context: https://central.tri.be/issues/40720

On some machines the chosen log file destination could be not writeable; let's check for it before trying to open it and raise warnings.


src/Tribe/Log/File_Logger.php
```diff
@@ -59,7 +59,7 @@ protected function set_context( $context ) {
 	 */
 	protected function obtain_handle() {
 		$this->close_handle();
-		$this->handle = fopen( $this->log_file, $this->context );
+		$this->handle = is_writeable( $this->log_file ) && fopen( $this->log_file, $this->context );
```
barryhughes: I think this would generally be safe, yet what if `$this->context` was `'r'` (read only) - we'd potentially be blocking reads of historic log files where we don't care if they are writeable or not.

Could checks elsewhere as to whether `$this->handle` is a false-like value be a better way to go?

lucatume:
Yep, you are right.

barryhughes:
:camel:

PR #173, opened 2016-08-26
Move Permalinks in Support Info

jesseeproductions:
_Ref:_ [C#41136](https://central.tri.be/issues/41136) - move permalinks option closer to top


barryhughes:
:dog2:

PR #177, opened 2016-08-29
Fix checks against debug level, avoid undefined index notice | spotfix

barryhughes:
Minor fixes for logging (check the debug level is valid).


jesseeproductions:
:+1:

PR #178, opened 2016-08-30
Protected against possibility DateTime obj can't be created | #65755

barryhughes:
Guard against the possibility `date_create()` will return false.

[#65578](https://central.tri.be/issues/65678)


lucatume:
ok!

PR #179, opened 2016-08-30
Undo a change that meant the file handle was set to a boolean value | spotfix

barryhughes:
Bah! I saw the previous change adding `is_readable()` come in in an earlier PR and didn't spot it would break things. This keeps the same test, but ensures the handle is set appropriately.


borkweb:
Oh hahaha :D Nice fix. :+1:

PR #180, opened 2016-08-31
Set API error booleans in API response

borkweb:
This is a fix that is helpful for Event Aggregator and was driven by an item in the EA Findings Doc


barryhughes:
:school_satchel:

PR #185, opened 2016-09-02
Try to create new log file before testing if it is readable | #36047

barryhughes:
Further fix to allow creation of log file where it doesn't already exist.

[#36047](https://central.tri.be/issues/36047)


bordoni:
👍

PR #188, opened 2016-09-08
License key notification adjustments

barryhughes:
Further adjustments to our license key notifications:

* Plugin names must now be registered or they won't be included in notifications (thus a persistent notification re PRO will not display after that plugin is deactivated etc)
* Fixes an until now undetected bug: the inline notices returned by the PUE server were effectively dropped and replaced with a fallback
* Moves creation of the plugin-update-message table row markup from JS to PHP, so it can be modified via filters etc

Some other cleanup.

[#66191](https://central.tri.be/issues/66191)


src/Tribe/PUE/Checker.php
```diff
@@ -589,14 +593,15 @@ private function get_api_message( $info ) {
...
-				$message = wp_kses( $info->api_invalid_message, 'post' );
+			if ( ! empty( $info->api_inline_invalid_message ) ) {
+				$message = wp_kses( $info->api_inline_invalid_message, 'post' );
 			}
 
 			$message = str_replace( '%plugin_name%', $this->get_plugin_name(), $message );
 			$message = str_replace( '%plugin_slug%', $this->get_slug(), $message );
 			$message = str_replace( '%update_url%', $this->get_pue_update_url(), $message );
 			$message = str_replace( '%version%', $info->version, $message );
+			$message = str_replace( '%changelog%', '<a class="thickbox" title="' . $this->get_plugin_name() . '" href="plugin-install.php?tab=plugin-information&plugin=' . $this->get_slug() . '&TB_iframe=true&width=640&height=808">what\'s new</a>', $message );
```
zbtirrell: `what's new` should be translatable

barryhughes: This is old code I restored but you're totally right.

Probably a task for another day but I actually wonder about modifying this system to support something like `%changelog="view the changelog"%` as otherwise:

* When setting these messages the author has to work around the pre-coded wording
* There is also nothing (in the PUE server settings) to remind the author of exactly what `%changelog%` and co are transformed into

Then again, I guess we rarely alter these things.

barryhughes: ... Actually, I take that back, I'm not sure you are right.

The rest of the text is set on theeventscalendar.com and is in English, it is not translatable. If we make "what's new" translatable by itself we'd have the spectre of messages like:

> An update is available! Download now or Quoi de neuf.

Which is fun, but an odd mixture.

PR #190, opened 2016-09-08
avoid array_diff_key warning if non array

lucatume:
Ticket: none

Avoid PHP warnings when PUE checks for notices; found out while spinning up latest `develop` code.


barryhughes:
Thanks!

PR #191, opened 2016-09-08
Fix/65617 double notice

lucatume:
Ticket https://central.tri.be/issues/65617

This PR:

* moves the bumpdown style out of the admin file
* modifies the bumpdown code to avoid hiding the bumpdown trigger


barryhughes:
:cat:

PR #193, opened 2016-09-12
Feature/65890 authorize extensions

elimn:
Ticket: https://central.tri.be/issues/65890

This is highly related to our new extension (pluginized snippet) template: https://github.com/moderntribe/snippets/blob/snippet-framework-common-merge/README.md

These libs are the basis for our new extension system. The goal is to create a function that each extension can check to see if it is authorized to run. For now the only reason a extension won't be authorized to run is if the plugins it requires are not activated. In the future there might be other times, such as when a new version of the extension is available, or the need for the extension has been removed.

### DIFF
--- a/src/Tribe/Plugin_Dependencies.php
+++ b/src/Tribe/Plugin_Dependencies.php
@@ -1,0 +1,183 @@
+<?php defined( 'WPINC' ) or die;
+
+if (! class_exists( 'Tribe__Plugin_Dependencies' ) ) {
+	/**
+	 * Tracks which tribe plugins are currently activated
+	 */
+	class Tribe__Plugin_Dependencies {
+
+		/**
+		 * An multidimensional array of active tribe plugins in the following format
+		 *
+		 * array(
+		 *  'class'   => 'main class name',
+		 *  'version' => 'version num',
+		 *  'path'    => 'Path to the main plugin/bootstrap file'
+		 * )
+		 */
+		protected $active_plugins = array();
+
+		protected static $instance;
+
+		/**
+		 * Static Singleton Factory Method
+		 *
+		 * @return Tribe__Plugin_Dependencies
+		 */
+		public static function instance() {
+			if ( ! isset( self::$instance ) ) {
+				$className      = __CLASS__;
+				self::$instance = new $className;
+			}
+
+			return self::$instance;
+		}
+
+		public function __construct() {
+			$this->add_legacy_plugins();
+		}
+
+
+		/**
+		 * Registers older plugins that did not use this class
+		 */
+		private function add_legacy_plugins () {
+			// Version 4.2 and under of the plugins do not register themselves here, so we'll register them
+			// @TODO Consider removing this in 5.0
+
+			$tribe_plugins = new Tribe__Plugins_List();
+
+			foreach( $tribe_plugins->get_list() as $plugin ) {
+				if ( ! class_exists( $plugin[ 'class' ] ) ) continue;
+
+				$version = constant( $plugin[ 'class' ] . '::VERSION' );
+
+				$this->add_active_plugin( $plugin[ 'class' ], $version );
+			}
+		}
+
+
+		/**
+		 * Retrieves active plugin array
+		 *
+		 * @return array
+		 */
+		public function get_active_plugins() {
+			return $this->active_plugins;
+		}
+
+
+		/**
+		 * Adds a plugin to the active list
+		 *
+		 * @param string $main_class Main/base class for this plugin
+		 * @param string $version    Version number of plugin
+		 * @param string $path       Path to the main plugin/bootstrap file
+		 */
+		public function add_active_plugin( $main_class, $version = null, $path = null ) {
+
+			$plugin = array(
+				'class'   => $main_class,
+				'version' => $version,
+				'path'    => $path,
+			);
+
+			$this->active_plugins[ 'class' ] = $plugin;
+		}
+
+
+		/**
+		 * Searches the plugin list for key/value pair and return the full details for that plugin
+		 *
+		 * @param string $search_key The array key this value will appear in
+		 * @param string $search_val The value itself
+		 *
+		 * @return array|null
+		 */
+		public function get_plugin_by_key( $search_key, $search_val ) {
+			foreach ( $this->active_plugins as $plugin ) {
+				if ( isset( $plugin[ $search_key ] ) && $plugin[ $search_key ] === $search_val ) {
+					return $plugin;
+				}
+			}
+
+			return null;
+		}
+
+
+		/**
+		 * Retrieves the plugins details by class name
+		 *
+		 * @param string $main_class Main/base class for this plugin
+		 *
+		 * @return array|null
+		 */
+		public function get_plugin_by_class( $main_class ) {
+			return $this->get_plugin_by_key( 'class', $main_class );
+		}
+
+
+		/**
+		 * Retrieves the version of the plugin
+		 *
+		 * @param string $main_class Main/base class for this plugin
+		 *
+		 * @return string|null Version
+		 */
+		public function get_plugin_version( $main_class ) {
+			$plugin = $this->get_plugin_by_class( $main_class );
+
+			return ( isset( $plugin[ 'version' ] ) ? $plugin[ 'version' ] : null );
+		}
+
+
+		/**
+		 * Checks if the plugin is active
+		 *
+		 * @param string $main_class Main/base class for this plugin
+		 *
+		 * @return bool
+		 */
+		public function is_plugin_active( $main_class ) {
+			return ( $this->get_plugin_by_class( $main_class ) !== null );
+		}
+
+
+		/**
+		 * Checks if a plugin is active and has the specified version
+		 *
+		 * @param string $main_class Main/base class for this plugin
+		 * @param string $version Version to do a compare against
+		 * @param string $compare Version compare string, defaults to >=
+		 *
+		 * @return bool
+		 */
+		public function is_plugin_version( $main_class, $version, $compare = '>=' ) {
+			return (
+				$this->is_plugin_active( $main_class ) &&
+				version_compare( $this->get_plugin_version( $main_class ), $version, $compare )
+			);
+		}
+
+
+		/**
+		 * Checks if each plugin is active and exceeds the specified version number
+		 *
+		 * @param array $plugins_required
+		 *
+		 * @return bool
+		 */
+		public function has_requisite_plugins( $plugins_required = array() ) {
+
+			foreach( $plugins_required as $class => $version ) {
+				// Return false if the plugin is not set or is a lesser version
+				if( ! $this->is_plugin_active( $class ) ) return false;
+				if( $version !== null && ! $this->is_plugin_version( $class, $version ) ) return false;
+			}
+
+			return true;
+		}
+
+	}
+
+}
--- a/src/Tribe/Plugin_Download_Notice.php
+++ b/src/Tribe/Plugin_Download_Notice.php
@@ -1,0 +1,72 @@
+<?php defined( 'WPINC' ) or die;
+
+if ( ! class_exists( 'Tribe__Plugin_Download_Notice' ) ) {
+
+	/**
+	 * Shows an admin notice telling users which requisite plugins they need to download
+	 */
+	class Tribe__Plugin_Download_Notice {
+
+		private $plugin_path;
+
+		private $plugins_required = array();
+
+		/**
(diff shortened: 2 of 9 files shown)


src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
+<?php defined( 'WPINC' ) or die;
```
bordoni: Add a Breakline befone `defined`.

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+
+if (! class_exists( 'Tribe__Plugin_Dependencies' ) ) {
```
bordoni: Bad spacing `(!`

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+	/**
+	 * Tracks which tribe plugins are currently activated
+	 */
+	class Tribe__Plugin_Dependencies {
```
bordoni: We could use `Tribe__Dependency`, right?

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+		 *
+		 * array(
+		 *  'class'   => 'main class name',
+		 *  'version' => 'version num',
+		 *  'path'    => 'Path to the main plugin/bootstrap file'
+		 * )
+		 */
+		protected $active_plugins = array();
+
+		protected static $instance;
```
bordoni: For the `singleton` patter, could you follow the following: https://github.com/moderntribe/the-events-calendar/blob/develop/src/Tribe/Aggregator/Cron.php#L30-L48

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+			$this->add_legacy_plugins();
+		}
+
+
+		/**
+		 * Registers older plugins that did not use this class
+		 */
+		private function add_legacy_plugins () {
+			// Version 4.2 and under of the plugins do not register themselves here, so we'll register them
+			// @TODO Consider removing this in 5.0
```
bordoni: I would move this to the docblock.

src/Tribe/Plugins_List.php
```diff
@@ -0,0 +1,126 @@
+<?php defined( 'WPINC' ) or die;
+
+if (! class_exists( 'Tribe__Plugins_List' ) ) {
+	/**
+	 * A list of Tribe's major plugins. Useful when encouraging users to download one of these.
+	 */
+	class Tribe__Plugins_List {
```
bordoni: Can we avoid this class as a whole by inserting our plugins on the `__construct` of the _Dependendecy_ class?

src/Tribe/Plugin_Download_Notice.php
```diff
@@ -0,0 +1,72 @@
+<?php defined( 'WPINC' ) or die;
+
+if ( ! class_exists( 'Tribe__Plugin_Download_Notice' ) ) {
+
+	/**
+	 * Shows an admin notice telling users which requisite plugins they need to download
+	 */
+	class Tribe__Plugin_Download_Notice {
```
bordoni: Same thing here, move these methods to the _Dependency_ class, and we also should be using the new `Tribe__Admin__Notice` class, with the helper `tribe_notice`.

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+
+		public function __construct() {
+			$this->add_legacy_plugins();
+		}
+
+
+		/**
+		 * Registers older plugins that did not use this class
+		 */
+		private function add_legacy_plugins () {
```
barryhughes: Space fix needed!

For functions, we don't add a space before the parenthesis. For control structures we do.

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+			// Version 4.2 and under of the plugins do not register themselves here, so we'll register them
+			// @TODO Consider removing this in 5.0
+
+			$tribe_plugins = new Tribe__Plugins_List();
+
+			foreach( $tribe_plugins->get_list() as $plugin ) {
```
barryhughes: Space fix needed!

Reverse of the above: we _do_ what a space before the parenthesis here, as this is a control structure rather than a function.

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+				if ( ! class_exists( $plugin[ 'class' ] ) ) continue;
```
barryhughes: Use of `{}` is required for the `if` statement body, no one liners I'm afraid!

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+				$this->is_plugin_active( $main_class ) &&
+				version_compare( $this->get_plugin_version( $main_class ), $version, $compare )
+			);
+		}
+
+
+		/**
+		 * Checks if each plugin is active and exceeds the specified version number
+		 *
+		 * @param array $plugins_required
```
barryhughes: Describe the expected format?

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+		 *
+		 * @return bool
+		 */
+		public function has_requisite_plugins( $plugins_required = array() ) {
+
+			foreach( $plugins_required as $class => $version ) {
```
barryhughes: Space needed before the parenthesis, ditto for the two `if` statements below (and they also ought to use `{}`).

src/Tribe/Plugins_List.php
elimn: @bordoni If you are wondering if a Plugin such as Events Calendar Pro could register this info when it activates: I wish, it would be a lot cleaner. But we need the list of plugins and their download URLs when those plugins are not active on the system. Did I understand correct?

src/Tribe/Plugin_Dependencies.php
```diff
@@ -0,0 +1,183 @@
...
+			// @TODO Consider removing this in 5.0
+
+			$tribe_plugins = new Tribe__Plugins_List();
+
+			foreach( $tribe_plugins->get_list() as $plugin ) {
+				if ( ! class_exists( $plugin[ 'class' ] ) ) continue;
+
+				$version = constant( $plugin[ 'class' ] . '::VERSION' );
+
+				$this->add_active_plugin( $plugin[ 'class' ], $version );
```
barryhughes: I could easily be missing something but it looks like we don't set the `path` for any of the active plugins.

Do we need it? If we don't need it/it's not always going to be set if we should say as much - perhaps in the docblock for `$this->active_plugins` at the top of the class.

src/Tribe/Plugins_List.php
bordoni: No no, not when the plugins are active, just avoid one more class by registering the list of plugins inside of the `Dependency` class, the one you created on this PR.

src/Tribe/Plugin_Download_Notice.php
```diff
@@ -0,0 +1,72 @@
...
+				'name'         => $name,
+				'thickbox_url' => $thickbox_url,
+			);
+		}
+
+		/**
+		 * Echoes the admin notice, attach to admin_notices
+		 */
+		public function show_inactive_plugins_alert() {
+			if ( ! current_user_can( 'activate_plugins' ) )	return;
```
barryhughes: No one liners :wink:

src/Tribe/Plugin_Download_Notice.php
```diff
@@ -0,0 +1,72 @@
...
+
+			$plugin_data = get_plugin_data( $this->plugin_path );
+
+			$req_plugins = array();
+
+			foreach ( $this->plugins_required as $req_plugin ) {
+
+				$item = esc_html( $req_plugin[ 'name' ] );
+
+				if( ! empty( $req_plugin[ 'thickbox_url' ] ) ){
```
barryhughes: Space before and after the parenthesis :stuck_out_tongue_winking_eye:

src/functions/utils.php
```diff
@@ -25,3 +25,36 @@ function tribe_array_merge_recursive( array &$array1, array &$array2 ) {
...
+	 * Checks if this plugin has permission to run, if not it notifies the admin
+	 *
+	 * @param string $file_path   Full file path to the base plugin file
+	 * @param string $main_class  The Main/base class for this plugin
+	 * @param string $version     The version
+	 * @param array  $classes_req Any Main class files/tribe plugins required for this to run
+	 *
+	 * @return bool Indicates if plugin should continue initialization
+	 */
+	function tribe_can_plugin_run( $file_path, $main_class, $version, $classes_req = array() ) {
```
barryhughes: I'm not completely sold on the name of the function. It doesn't just return true or false according to whether dependencies are met, it also registers the new plugin if appropriate. What about `tribe_register_plugin()` instead? Additionally, we should make it clear in the docblock that it handles registration.

src/functions/utils.php
```diff
@@ -25,3 +25,36 @@ function tribe_array_merge_recursive( array &$array1, array &$array2 ) {
...
+
+		if ( $tribe_plugins->has_requisite_plugins( $classes_req ) ) {
+			$tribe_plugins->add_active_plugin( $main_class, $version, $file_path );
+
+			return true;
+		} elseif ( is_admin() ) {
+			$tribe_plugins = new Tribe__Plugins_List();
+			$admin_notice  = new Tribe__Plugin_Download_Notice( $file_path );
+
+			foreach ( $classes_req as $class => $version ) {
```
barryhughes: It looks like we're going to notify admin users of all the required plugins (even if just one is missing). That may be fine - but I wanted to check it was intentional.

src/Tribe/Plugins_List.php
elimn: I got you. I was under the impression we were trying to make smaller classes, hence why there is 3. They are all strongly related though. If y'all would prefer one class that merged all of these I'm down.

bordoni: I mean, depends a lot of why we are breaking up the class.

elimn: That makes sense. I'd like to see Plugin_List stay separate. First because I could see us adding more info to it at some point that would make it more useful in other circumstances. Ideally we would only have one list of this sort of info to maintain. We already much of this info duplicated across our plugin suite in a few areas, so it would be awesome if we could start consolidating. Second, I could see a small possibility in the future where we make info like this info obtainable from the tec servers, and for that I am keen to keep that class separate with it's own get_list method.

src/Tribe/Plugin_Download_Notice.php
elimn: I love the idea of getting rid the notice class using the new stuff in 4.3. But this in intended for 4.2.7. What if I tagged this with a todo and ticketed removing it once 4.3 is released? That way we don't invest anymore time in something that's soon to go away.

src/Tribe/Plugin_Dependencies.php
elimn: Thanks for the feedback human linter. It made me reread our coding guidelines, which have been updated since the last time I read them. I'll get on these.

src/Tribe/Plugin_Dependencies.php
elimn: Path is not required, but will be added by any extension in the future which uses the run plugin function created below. If everything goes according to plan all of our extensions, and eventually all of our plugins should be using it and thus will be adding their path. There are a number of scenarios where it could be useful, for things such as plugin_info() and deactivate_plugins() which are very probably things that we will eventually be using as part of our communication with tec servers and keeping extensions up to date.

I added some documentation at the top as well indicating it's optional. But I'd still like to collecting that info wherever possible.

src/functions/utils.php
elimn: Great idea.

src/functions/utils.php
elimn: That is intentional, just because it's what our current plugin notices do. I had considered maybe bolding ones that were inactive or some other visual trick. I might leave enhancements like this to after we move over to tribe_notice() in 4.3.

elimn:
I accidentally selected develop as the branch to merge this into, it should be r/4.2.7. Going to close this and reopen with the proper branch.

PR #194, opened 2016-09-12
Feature/65890 authorize extensions

elimn:
Ticket: https://central.tri.be/issues/65890

This is a continuation of: https://github.com/moderntribe/tribe-common/pull/193 which was going to be merged into the wrong branch. I have applied all the feedback from that PR, but opened a new one against the proper branch.

> These libs are the basis for our new extension system. The goal is to create a function that each extension can check to see if it is authorized to run. For now the only reason a extension won't be authorized to run is if the plugins it requires are not activated. In the future there might be other times, such as when a new version of the extension is available, or the need for the extension has been removed.

### DIFF
--- a/lang/tribe-common.pot
+++ b/lang/tribe-common.pot
@@ -2,7 +2,7 @@
 # This file is distributed under the same license as the Tribe Common package.
 msgid ""
 msgstr ""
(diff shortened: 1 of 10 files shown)


bordoni:
👍

PR #195, opened 2016-09-13
Swap out the Cassowary with the Squid for the spirit animal in notice

borkweb:
See: https://central.tri.be/issues/66191


barryhughes:
Goodbye cassowary. Extinction beckons.

:octopus:

PR #197, opened 2016-09-13
get_log_entries() should always return an array, even if it's empty | spotfix

barryhughes:
If logging hasn't yet been enabled, null is returned where an array is expected. This corrects that to an empty array.


bordoni:
👍

PR #198, opened 2016-09-13
Fix tribe image resource URLs in app shop | spotfix

barryhughes:
Fix broken app shop images (pass the Tribe__Main object so it can figure out the paths from there).

### DIFF
--- a/src/admin-views/app-shop.php
+++ b/src/admin-views/app-shop.php
@@ -2,7 +2,7 @@
 
 	<div class="header">
 		<h1><?php esc_html_e( 'Tribe Event Add-Ons', 'tribe-common' ); ?></h1>
-		<a href="https://theeventscalendar.com/?utm_campaign=in-app&utm_source=addonspage&utm_medium=top-banner" target="_blank"><img src="<?php echo esc_url( tribe_resource_url( 'images/app-shop-banner.jpg', false, 'common' ) ); ?>" /></a>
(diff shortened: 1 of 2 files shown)


src/admin-views/app-shop.php
```diff
@@ -2,7 +2,7 @@
 
 	<div class="header">
 		<h1><?php esc_html_e( 'Tribe Event Add-Ons', 'tribe-common' ); ?></h1>
-		<a href="https://theeventscalendar.com/?utm_campaign=in-app&utm_source=addonspage&utm_medium=top-banner" target="_blank"><img src="<?php echo esc_url( tribe_resource_url( 'images/app-shop-banner.jpg', false, 'common' ) ); ?>" /></a>
+		<a href="https://theeventscalendar.com/?utm_campaign=in-app&utm_source=addonspage&utm_medium=top-banner" target="_blank"><img src="<?php echo esc_url( tribe_resource_url( '/images/app-shop-banner.jpg', false, null, $main ) ); ?>" /></a>
```
bordoni: Shouldn't we use the `Tribe__Main::$tec_url` here to compose this URL? Or maybe a `m.tri.be` link

barryhughes: We probably should - I think I left the same feedback in the PR that added those links - but in this case all I'm doing is correcting the resource URL for the images.

bordoni: As we talked I will ping Zach about these, on the On going discussion.

bordoni:
👍

PR #206, opened 2016-09-19
Add notices any time we save the key 

borkweb:
Additionally, clear notices when saving the key.

Related: https://github.com/moderntribe/the-events-calendar/pull/987

See: https://central.tri.be/issues/65378

### DIFF
--- a/src/Tribe/PUE/Checker.php
+++ b/src/Tribe/PUE/Checker.php
@@ -548,15 +548,12 @@ if ( ! class_exists( 'Tribe__PUE__Checker' ) ) {
 			} elseif ( isset( $plugin_info->api_expired ) && $plugin_info->api_expired == 1 ) {
 				$response['message'] = $this->get_license_expired_message();
 				$response['api_expired'] = true;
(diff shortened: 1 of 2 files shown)


bordoni:
🎱

src/Tribe/PUE/Checker.php
```diff
@@ -568,6 +571,8 @@ public function validate_key( $key ) {
 					}
 				}
 
+				$pue_notices->clear_notices( $plugin_name );
```
barryhughes: Just checking: are we positive we wish to clear notices at this point (during ajax validation, before a key has been committed/saved)?

borkweb: When a valid license key is entered and gets ajax verified, it actually _does_ get saved. The only time the ajax verification doesn't get saved is when the key is invalid, requires an upgrade, or expired.

barryhughes: You're totally right, I missed that completely :sleepy:

PR #209, opened 2016-09-20
Feature/46122 abstract activation screen

barryhughes:
Moves the activation page class to tribe-common (from The Events Calendar). Some restructuring, but there are also sizeable chunks which were not touched.

https://central.tri.be/issues/66742

### DIFF
--- a/src/Tribe/Admin/Activation_Page.php
+++ b/src/Tribe/Admin/Activation_Page.php
@@ -27,6 +27,7 @@ class Tribe__Admin__Activation_Page {
 	public function __construct( array $args = array() ) {
 		$this->args = wp_parse_args( $args, array(
 			'slug'                  => '',
+			'activation_transient'  => '',
 			'version'               => '',
 			'plugin_path'           => '',
 			'version_history_slug'  => '',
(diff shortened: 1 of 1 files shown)


src/Tribe/Admin/Activation_Page.php
```diff
@@ -0,0 +1,283 @@
...
+	/**
+	 * Prints the splash screen.
+	 *
+	 * @param string $context
+	 */
+	public function display_page( $context = 'welcome' ) {
+		do_action( 'tribe_settings_top' );
+
+		$context = isset( $_GET[ $this->welcome_slug ] ) ? 'welcome': 'update';
+		$title   = esc_html( @$this->args[ $context . '_page_title'] );
```
bordoni: Is it the best option using `@`?

barryhughes: It may not be the very best option :wink:

It's a precaution; in general and in our own internal usage it shouldn't be needed. As the method is public, though (has to be, it's a callback), I added it as a precaution.

We can add extra checks though, what do you think?

bordoni: I would rather have a few checks instead of `@`.

barryhughes: OK - will add :+1:

bordoni:
👍

PR #210, opened 2016-09-21
tribe-common should register its own deprecated classes with the autoloader

barryhughes:
Let tribe-common register its own deprecated classes, rather than have dependencies like TEC and ET take care of it. Apart from better separation of concerns, this should help avoid fatals upon updates where both those plugins are installed and activated.

https://central.tri.be/issues/66872


bordoni:
👍

PR #211, opened 2016-09-22
Fix context for splash screen (ie, whether 'update' or 'welcome') | #66742

barryhughes:
Plugs a gap in the splash screen setup, letting us display the correct (update or welcome) screen.

[#66742](https://central.tri.be/issues/66742)


bordoni:
👍

PR #213, opened 2016-09-22
added the `tribe_exit` function

lucatume:
Ticket: nope

Adds the filterable `tribe_exit` function to allow us to control `die` calls.


barryhughes:
:game_die:

PR #216, opened 2016-09-26
Reduce list of plugins to those that have been registered and *then* count them

barryhughes:
First check how many _registered_ plugins require a notice to be displayed, returning an empty value if there are none.

### DIFF
--- a/src/Tribe/PUE/Notices.php
+++ b/src/Tribe/PUE/Notices.php
@@ -360,11 +360,9 @@ class Tribe__PUE__Notices {
 
 		if ( 0 === $num_plugins ) {
 			return '';
-		}
-		elseif ( 1 === $num_plugins ) {
+		} elseif ( 1 === $num_plugins ) {
 			$html = current( $plugin_list );
-		}
-		elseif ( 1 < $num_plugins ) {
+		} elseif ( 1 < $num_plugins ) {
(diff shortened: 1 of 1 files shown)


src/Tribe/PUE/Notices.php
```diff
@@ -351,19 +351,20 @@ protected function find_your_key_text() {
...
 			return '';
 		}
 
 		$plugin_list = array_intersect( $this->registered, array_keys( $this->notices[ $group ] ) );
+		$num_plugins = count( $plugin_list );
 
-		if ( 1 === $num_plugins ) {
+		if ( 0 === $num_plugins ) {
+			return '';
+		}
```
bordoni: Can we remove the Line Break here?

src/Tribe/PUE/Notices.php
```diff
@@ -351,19 +351,20 @@ protected function find_your_key_text() {
...
+		elseif ( 1 === $num_plugins ) {
 			$html = current( $plugin_list );
 		}
-
-		if ( 1 < $num_plugins ) {
```
bordoni: _Ditto_

src/Tribe/PUE/Notices.php
barryhughes: Hah, yes we can. I did indeed deliberately break away from our established style here because I find this far more readable, but will adjust.

PR #219, opened 2016-09-28
Replace Facebook and iCal with Event Aggregator in the App Shop

borkweb:
I'm purposefully keeping the iCal image as it still seems to work for EA


trishasalas:
🐠

zbtirrell:
Late breaking feedback, can we adjust the order of these to be as such:

EA (wide slot)
PRO, ET+, Filterbar
Community, Comm. Tickets, Eventbrite

eliminates the 2nd wide slot, but leaves everything spaced evenly

trishasalas:
Another late breaking question...after thinking on this some I wonder if about using the m.tribe links like we did in the Upsell....just a thought.

borkweb:
I've re-ordered the products. For now, I'll keep the links as-is - I heard rumbling that the m.tri.be links mess with our tracking a bit due to how it redirects, but I don't know for sure. I'll match the current links for now.

borkweb:
As such, this needs another look-see :)

zbtirrell:
👍

PR #220, opened 2016-09-29
Don't try to call die or exit via call_user_func() | spotfix

barryhughes:
Language constructs like `exit` and `die` don't seem to be viable as callbacks. Solid catch by [Jeff](https://tribe.slack.com/archives/quality-assurance/p1475188920001344).


bordoni:
🤘🏻

PR #226, opened 2016-10-06
Prevent notices on JSON LD in month view when there are no posts in day

borkweb:


lucatume:
Makes sense

lucatume:
Makes sense

PR #233, opened 2016-10-14
Update autoload_52.php | spotfix

barryhughes:
Update to match actual classname within `composer/autoload_real_52.php`.


borkweb:
:+1:

PR #235, opened 2016-10-17
Handle packages during update to shorten filenames.

lucatume:
Ticket: https://central.tri.be/issues/62552

This tickets adds the `Tribe__PUE__Package_Handler` class to jump in during our plugin updates and shorten the temporary filename.
The logic is that we have no control over the temporary filename and that is generated during download creating it from the download URL and query args.  
The following operation, `WP_Upgrader::unpack_package`, would then use that temp file name as the name of the temp dir; that temp dir name is what makes the trouble (see ticket description); changing the temp file name we change the temp dir name.

Please check:
* I'm using the `pu_get_download` flag to identify a package as "ours", correct?
* I'm presuming all our downloads happen from a URL, correct?
* I'm assuming .org plugins too will go through PUE, correct?

Improvement: if we could have a whitelist of URLs for our updates another check could be put in place to make sure it's our update.


barryhughes:
Good to merge _but_ we probably want to add a changelog entry for this.

lucatume:
Ok, adding changelog not on TEC and Tickets per Slack discussion

PR #241, opened 2016-10-18
update di52  to version 1.4.3

lucatume:
Notable is only the change of some methods from snake_case to camelCase on di52 side.


borkweb:
:dance:

PR #242, opened 2016-10-18
Added Tribe__Extension

elimn:
This PR is here at the request of Gustavo. It will quite possible get closed. Discussion is in the ticket.

Some work in progress documentation on how to use this: https://github.com/theeventscalendar/example-extension/tree/extension-base-class

Ticket: https://central.tri.be/issues/68188

### DIFF
--- a/src/Tribe/Extension.php
+++ b/src/Tribe/Extension.php
@@ -95,7 +95,7 @@ class Tribe__Extension {
 	final private function __construct() {
 		$this->construct();
 
-		// The init() action/hook
+		// The init() action/hook.
 		$init_hook = $this->get_init_hook();
 
 		// Continue plugin run after $init_hook has fired.
@@ -118,7 +118,7 @@ class Tribe__Extension {
 	 * init() will not run.
 	 *
 	 * @param string      $main_class      The Main class for this Tribe plugin.
-	 * @param string|null $minimum_version Minimum acceptable version of plugin
+	 * @param string|null $minimum_version Minimum acceptable version of plugin.
 	 */
 	final protected function add_required_plugin( $main_class, $minimum_version = null ) {
 		$this->set( array( 'requires', $main_class ), $minimum_version );
@@ -127,7 +127,7 @@ class Tribe__Extension {
 	/**
 	 * Set the extension's tec.com URL
 	 *
-	 * @param string $url URL to the extension's page
+	 * @param string $url URL to the extension's page.
 	 */
 	final protected function set_url( $url ) {
 		$this->set( 'url', $url );
@@ -251,7 +251,7 @@ class Tribe__Extension {
 	final public function get_arg_or_plugin_data( $arg, $plugin_data_key ) {
 		$arg_value = $this->get( $arg, null );
 
-		// See if the arg is already set, if not get default from plugin data and set it
+		// See if the arg is already set, if not get default from plugin data and set it.
 		if ( null === $arg_value ) {
 			$pdata = $this->get_plugin_data();
 			$arg_value = isset( $pdata[ $plugin_data_key ] ) ? $pdata[ $plugin_data_key ] : null;
@@ -278,7 +278,7 @@ class Tribe__Extension {
 
 		// Multiple nested keys specified, iterate through each level.
 		foreach ( $key as $i ) {
-			// Ensure current array depth can have children set
+			// Ensure current array depth can have children set.
 			if ( ! is_array( $arg ) ) {
 				// $arg is set but is not an array. Converting it to an array
 				// would likely lead to unexpected problems for whatever first set it.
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Logical operator &quot;or&quot; is prohibited; use &quot;||&quot; instead | [Extension.php:2](https://github.com/moderntribe/tribe-common/blob/f6cd2a5df395dc42b922e2e28faac6ea9c8c4aec/src/Tribe/Extension.php#L2)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$error' | [Extension.php:291](https://github.com/moderntribe/tribe-common/blob/f6cd2a5df395dc42b922e2e28faac6ea9c8c4aec/src/Tribe/Extension.php#L291)
:x: | Space found before comma in function call | [Extension.php:362](https://github.com/moderntribe/tribe-common/blob/f6cd2a5df395dc42b922e2e28faac6ea9c8c4aec/src/Tribe/Extension.php#L362)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not ')' | [Extension.php:476](https://github.com/moderntribe/tribe-common/blob/f6cd2a5df395dc42b922e2e28faac6ea9c8c4aec/src/Tribe/Extension.php#L476)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not ')' | [Extension.php:489](https://github.com/moderntribe/tribe-common/blob/f6cd2a5df395dc42b922e2e28faac6ea9c8c4aec/src/Tribe/Extension.php#L489)

src/Tribe/Extension.php
```diff
@@ -0,0 +1,493 @@
...
+	 * If you directly modify this inside of a child class then you're arguably
+	 * an evil person.
+	 *
+	 * @var array {
+	 *      Each instance of an extension that extends this class
+	 *
+	 *      @type object $child_class_name instance
+	 * }
+	 */
+	private static $_instances = array();
```
barryhughes: We'd probably want to remove the leading `_` from this var name.

src/Tribe/Extension.php
```diff
@@ -0,0 +1,493 @@
...
+	 *                               resulting data is stored in this.
+	 * }
+	 */
+	protected $_args = array();
+
+	/**
+	 * The various extension instances
+	 *
+	 * If you directly modify this inside of a child class then you're arguably
+	 * an evil person.
```
elimn: I was thinking you could set it to protected in the child and then do nasty things. I did not test this though.

src/Tribe/Extension.php
```diff
@@ -0,0 +1,493 @@
...
+	final public function register() {
+		$is_plugin_authorized = tribe_register_plugin(
+			$this->get_plugin_file(),
+			$this->get( 'class', get_class( $this ) ),
+			$this->get_version(),
+			$this->get( 'requires', array() )
+		);
+
+		if ( $is_plugin_authorized ) {
+			$this->init();
```
barryhughes: Should we have an abstract or empty `init()` method? Or at least call out the need for subclasses to implement this via the inline docs?

src/Tribe/Extension.php
barryhughes: Ah ok. It wouldn't actually override the visibility in that case, though - it would define a new property of protected scope. The contents of the private array here in the base class would remain untouched.

src/Tribe/Extension.php
elimn: I wondered about that too. WE might could even have a an implements set in our extensions, and just implement an interface that contains init().

Either way they will currently get an error if they fail to declare it.

barryhughes: OK, just seemed odd because we have an empty `construct()` method.

elimn: The reason constuct is present is because it is truly optional. Some extensions may choose to not declare it. But if they don't declare init() they have no reason to exist.

src/Tribe/Extension.php
barryhughes: I get your line of thinking but, though I can find the reference, I'm pretty sure this style of marking `_private_properties` runs against our coding standards. Also, per other comments, private fields aren't accessible by child classes in any case.

src/Tribe/Extension.php
elimn: Awesome! I'll make the modifications.

src/Tribe/Extension.php
elimn: 👍

borkweb: We explicitly call it out about Methods, but not about other properties. That's an oversight in our engineering docs, though. We _do_ stay clear of underscore prefixing.

barryhughes: That would be my personal druthers :smiley:

borkweb: Helps if I include a link: http://moderntribe.github.io/products-engineering/php/#php-classes-methods

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Visibility must be declared on method &quot;init&quot; | [Extension.php:131](https://github.com/moderntribe/tribe-common/blob/10158ee71d29ddbf9003d31b8c808a422d0e84ff/src/Tribe/Extension.php#L131)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | The abstract declaration must precede the visibility declaration | [Extension.php:131](https://github.com/moderntribe/tribe-common/blob/e9f6b8ebdc74e6f8cdcb1cc9228acdcaada23114/src/Tribe/Extension.php#L131)

elimn:
Per our discussion in #products-extensions this should be closed. There are a few more tests we would like to do before calling it ready for release.

PR #245, opened 2016-10-19
Add the language map class and reCAPTCHA implementation

lucatume:
Ticket: https://central.tri.be/issues/45094

This PR adds the `Language_Map` interface in commons and the recaptcha implementation.
The idea is that we could re-use the recaptcha stuff and we could deal with other APIs defining their own language code list.
This way we centralize the WP > API language code mapping.


bordoni:
👍

PR #251, opened 2016-10-31
Makes sure our Logger Code is PHP 5.2 Compatible

bordoni:


src/Tribe/Log/File_Logger.php
```diff
@@ -212,8 +212,15 @@ public function list_available_logs() {
 		foreach ( new DirectoryIterator( $this->log_dir ) as $node ) {
 			$name = $node->getFilename();
 
+			// DirectoryIterator::getExtension() is only available on 5.3.6
+			if ( version_compare( phpversion(), '5.3.6', '>=' ) ) {
+				$ext = $node->getExtension();
+			} else {
+				$ext = pathinfo( $name, PATHINFO_EXTENSION );
```
barryhughes: Wonder if we should just use `pathinfo()` and ditch the if/else?

bordoni: The Pathinfo is slower, does one more Stat.

barryhughes:
:hamburger:

PR #254, opened 2016-11-01
tribe_update_option(): return a boolean indicating success or failure

barryhughes:
It would be useful to get an indication of whether a `tribe_update_option()` call was successful or not.


lucatume:
indeed, ok!

PR #255, opened 2016-11-02
Fix to_tz method name

lucatume:
Ticket: https://central.tri.be/issues/41324

While working on the ticket I've found out this issue where `to_tz` method name was changed to `jo_tz` which is a cool as it is not functional.


barryhughes:
merge-zebra

PR #259, opened 2016-11-06
Changes to styles, dependency.js, and adding dropdowns.js for Recurrence

borkweb:
These changes support the changes being made on the Recurrence UI in Pro.

See: https://central.tri.be/issues/66713


bordoni:
🎱

PR #262, opened 2016-11-07
Fix SSL problem on resources

bordoni:
_Ref:_ [C#68372](https://central.tri.be/issues/68372)


barryhughes:
Looks good - but should we add a changelog entry?

PR #263, opened 2016-11-08
Feature/68188 extension parent class

elimn:
Ticket: https://central.tri.be/issues/68188

We are adding a parent Extension class to common, so that we can quickly build powerful mini-plugins/extensions. The primary goal is to move as much common logic as possible for these extensions into Common. 

To that end we have come up with a new template for extensions. Below are a couple of examples the template. 

- https://github.com/theeventscalendar/extensions/tree/brooks-inital-extensions/tribe-ext-example
- https://github.com/theeventscalendar/extensions/tree/brooks-inital-extensions/tribe-ext-example2

As you can see the `Tribe__Extension_Loader` searches for some extra fields that it expects to find in the plugin header. When the `Extension File` field is present, it will include the specified file. This allows us to store the extension in a separate file and only load it after Tribe Common. When the `Extension Class` field is present, it will initalize the specified class. The Extension Loader searches for these header fields inside any plugin folder that has the `tribe-ext-` prefix.


src/Tribe/Extension.php
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	/**
+	 * Set the extension's tec.com URL
+	 *
+	 * @param string $url URL to the extension's page.
+	 */
+	final protected function set_url( $url ) {
+		$this->set( 'url', $url );
+		$this->add_meta_link( __( 'View Details' ), $url );
```
borkweb: Add the textdomain to `__( 'View Details' )`?

src/Tribe/Extension_Loader.php
```diff
@@ -0,0 +1,242 @@
...
+	 *
+	 * @see strpos()
+	 *
+	 * @param string       $haystack String to search in.
+	 * @param array|string $needles  Strings to search for.
+	 * @param int          $offset   Starting position of search.
+	 *
+	 * @return false|int Integer position of first needle occurrence.
+	 */
+	public static function strpos_array( $haystack, $needles, $offset = 0 ) {
```
borkweb: This may have uses beyond this class. I'm thinking this might have a nice home here: https://github.com/moderntribe/tribe-common/blob/develop/src/functions/utils.php

borkweb:
@elimn, this looks great! I'm pretty stoked about this addition. Nice work!

borkweb:
Could you add a readme.txt entry for this addition?

src/Tribe/Extension_Loader.php
elimn: That's a good idea. On second review it gave me some ideas. I wonder if I can break a few other things out:
- Tribe__Extension_Loader::get_plugin_data() could probably be made into tribe_get_plugin_data(), and used in Tribe__Extension::get_plugin_data() as well.
- Tribe__Extension::search_var() was handed to me by Gustavo, something very similar is used in a class he recently wrote. I wonder if that function, and a setting function based on Tribe__Extension::set(), would make sense somewhere? Perhaps in a a Tribe__Array_Utils class? And we could event put the strpos_array() in there if we wanted. What do you think?
- I think Tribe__Extension::filter_meta_links() and Tribe__Extension::add_meta_link() probably make sense inside of a Tribe__Plugin_Meta_Links class. 

I'd be happy to alter things a bit to make this happen. It might take an hour or two for something we technically have no budget for. But, if you approve I can squeeze this in before midday tomorrow so it can go through code review.

borkweb: @elimn - I like all of those ideas! If you have the time and the excitement, I think it would be a nice change.

elimn: Done!

src/Tribe/Utils/Plugins.php
```diff
@@ -0,0 +1,80 @@
...
+		if ( is_multisite() ) {
+			$full_list = array_merge( $full_list, wp_get_active_network_plugins() );
+		}
+
+		$filtered_list = array();
+
+		foreach ( $full_list as $plugin ) {
+			$base = plugin_basename( $plugin );
+
+			if ( Tribe__Utils__Array::strpos( $base, $prefix ) === 0 ) {
```
bordoni: Can you Yoda this guy?

bordoni:
💙🤘🏻

PR #264, opened 2016-11-08
removed transient setting

lucatume:

This PR removes the transient that would store the network activation state for subsite in favour of immediate feedback.


barryhughes:
:lion:

PR #265, opened 2016-11-10
Prevent checks on Plugins that doens't exist on the plugins folder

bordoni:
_Ref:_ [C#67981](https://central.tri.be/issues/67981)


borkweb:
:+1:

PR #270, opened 2016-11-11
make sure some data is carried over

lucatume:
Ticket: https://central.tri.be/issues/67143

This PR updates dropdowns.js to carry over some `data` attributes from elements after applying `select2` to it.
The reason being that @bordoni's excellent dependency automagical lib will require those `data` attributes on the top element and not on the hidden `select`

### DIFF
--- a/src/resources/js/dropdowns.js
+++ b/src/resources/js/dropdowns.js
@@ -190,14 +190,14 @@
 				};
 			}
 
-			$container = ($select.select2( args )).select2('container');
+			$container = ( $select.select2( args ) ).select2( 'container' );
 
 			if ( carryOverData.length > 0 ) {
 				carryOverData.map( function ( dataKey ) {
 					var attr = 'data-' + dataKey,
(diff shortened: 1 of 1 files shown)


src/resources/js/dropdowns.js
```diff
@@ -176,7 +190,20 @@
 				};
 			}
 
-			$select.select2( args );
+			$container = ($select.select2( args )).select2('container');
```
barryhughes: Spacing within parenthesis

src/resources/js/dropdowns.js
```diff
@@ -176,7 +190,20 @@
...
+
+			if ( carryOverData.length > 0 ) {
+				carryOverData.map( function ( dataKey ) {
+					var attr = 'data-' + dataKey,
+						val = $select.attr( attr );
+
+					if ( !val ) {
```
barryhughes: Space needed after `!`

lucatume:
spaces fixed

PR #272, opened 2016-11-14
allow for array not condition

lucatume:
Similarly to what is allowed for the `condition` condition allow for the `not-condition` to be an array in the format `["a", "b"]` (json format with *double* quotes.

	<span class="tribe-dependent" data-condition-not='["A", "B"]'>
		showing if not A and not B	
	</span>

@bordoni ideally you should review this.


borkweb:
This looks like a sexy improvement. 📟

PR #274, opened 2016-11-14
Add helper to safely retrieve post or URL query fields if set | #68051

barryhughes:
Adds a helper to retrieve post fields/URL query fields if set.

https://central.tri.be/issues/68051


src/functions/utils.php
```diff
@@ -123,3 +123,27 @@ function tribe_exit( $status = '' ) {
...
+	 * Post data takes priority over fields passed in the URL query. If the field is not
+	 * set then $default (null unless a different value is specified) will be returned.
+	 *
+	 * @param string $var
+	 * @param mixed  $default
+	 *
+	 * @return mixed
+	 */
+	function tribe_get_request_var( $var, $default = null ) {
+		if ( isset( $_POST[ $var ] ) ) {
```
bordoni: Could you use the `Utils__Array::get()` from: https://github.com/moderntribe/tribe-common/blob/release/4.3.3/src/Tribe/Utils/Array.php#L70

It's on our current MR, allows for deep search on an Array.

barryhughes: ...to expand on that, if we have `type[fruit]=apple&type[vegetable]=pear` I don't necessarily want to get back `apple` if I pass in `fruit` as the var name.

barryhughes: Actually, realizing I totally misunderstood how Utils__Array worked: yep we could use that, but it's not it develop yet (I can add a note to update once it is, though).

bordoni: It will be tonight, would you mind waiting to merge this after that is updated?

bordoni:
@barryhughes  just merged `release/4.3.3` into `develop` now we can use `Utils__Array`

borkweb:
:+1:

PR #276, opened 2016-11-15
conditionally fetch the parent slug

lucatume:
Ticket: https://central.tri.be/issues/69248

If TEC is active or not the parent slug should be different.


barryhughes:
Looks good, but should we add a changelog entry for this?

PR #277, opened 2016-11-15
Bump common version to avoid fatals

lucatume:
Ticket: https://central.tri.be/issues/69503 

When on `develop` common version (the constant `Tribe__Main::VERSION`) should always be updated to be > to the one set on `master`.


barryhughes:
:tiger:

PR #278, opened 2016-11-16
Handle plugin license visibility

lucatume:

This PR:

* prevents the network admin screen from showing if no "root" plugin is network activated (TEC or ET)
* handles visibility and editability of the license field to stick with user stories
* adds support to `dependency.js` for the `condition-checked` attribute
* to avoid larger specificity from blocking `dependency.js` show and hide operations it uses `show()` and `hide()` jQuery methods to show/hide stuff
* adds tests for relevant parts of code in Checker and Settings


src/Tribe/Settings.php
```diff
@@ -665,5 +681,45 @@ public function get_parent_slug() {
...
+			$root_plugin_is_mu_activated = array_sum( array_map( 'is_plugin_active_for_network', $this->root_plugins ) ) >= 1;
+
+			if ( ! $root_plugin_is_mu_activated ) {
+				return false;
+			}
+
+			if ( ! class_exists( 'Tribe__Events__Main' ) ) {
+				return true;
+			}
+
```
bordoni: Too much spacing.

src/resources/js/dependency.js
```diff
@@ -49,20 +50,22 @@
 						is_empty = $dependent.data( 'conditionEmpty' ) || $dependent.is( '[data-condition-empty]' ),
 						is_numeric = $dependent.data( 'conditionIsNumeric' ) || $dependent.is( '[data-condition-is-numeric]' ),
 						is_not_numeric = $dependent.data( 'conditionIsNotNumeric' ) || $dependent.is( '[data-condition-is-not-numeric]' ),
+						is_checked = $dependent.data('conditionChecked') || $dependent.is( '[data-condition-is-checked]'),
```
bordoni: Spacing.

src/resources/js/dependency.js
```diff
@@ -72,6 +75,7 @@
 						}
 					} else {
 						$dependent
+							.hide()
```
bordoni: No no, let's do it with classes please. To make it more flexible.

src/resources/js/dependency.js
```diff
@@ -49,20 +50,22 @@
...
 							|| ( is_not_empty && '' != value )
 							|| ( is_numeric && $.isNumeric( value ) )
+							|| ( is_checked && checked )
 							|| ( is_not_numeric && !$.isNumeric( value ) )
 							|| ( condition && ( _.isArray( condition ) ? -1 !== condition.indexOf( value ) : value == condition ) )
 							|| ( not_condition && ( _.isArray( not_condition ) ? -1 === not_condition.indexOf( value ) : value != not_condition ) )
 						) && ! is_disabled
 					) {
 						$dependent
+							.show()
```
bordoni: No no, let's do it with classes please. To make it more flexible.

src/resources/js/dependency.js
lucatume: It does not always work with classes.
We are styling the `.tribe-dependent` to `display: none`; that will not hide an element that is `display` styled with an heavier specificity like:

```html
<div class="wrapper">
    <span class="tribe-dependent some-class">Foo</span>
</div>
```

```css
.wrapper .some-class {
	display: inline-block;
}
```

The element specificity is higher than

```css
.tribe-dependent {
	display: none;
}
```

bordoni: Yeah, I know, the thing is, if we go with hiding it with JS, we will lose a lot of flexibility that the class gives us.

I mean, I would say that we could include that new `.wrapper .tribe-dependent` on the `.pcss`.

lucatume: That was an example; the case at hand has 3 classes nested in specificity.
We would end up having to take any possible use into account to use CSS to hide show, what's the problem with JS hide/show in this specific case? It's setting inline styles.

lucatume: and I've not removed the class, just use inline styles set by JS to make sure what is hidden is hidden and what should show shows

bordoni: So could you check if adding the class hides it or not `.is( ':hidden' )`? And based on that conditional use `.hide()/.show()`?

lucatume: this is a brilliant idea!

lucatume:
Here it is using `hide` and `show` if really necessary.

bordoni:
We might need to do the hide/show thing before the `select2( 'enable' )` because if you try to setup a Select2 while hidden it creates bugs.

lucatume:
I had no idea, thanks for sharing. I've updated the code.

bordoni:
👍

PR #279, opened 2016-11-17
add support for multiple conditions

lucatume:
Ticket: https://central.tri.be/issues/67143

Born in the context of the ticket this PR adds support for multiple conditions.  
By default the relation between the conditions will be the logic `or`:

```html
<span
    class="tribe-dependent"
    data-depends="#select"
    data-condition-empty
    data-condition="Foo"
>
```

The `span` **will show** if `#select` is empty OR has a value of 'Foo'.
To change the condition relation to the logic `and` one can use the `data-condition-relation="and"` attribute.

```html
<span
    class="tribe-dependent"
    data-depends="#select"
    data-condition-relation="and"
    data-condition-not-empty
    data-condition-not="Foo"
>
```

For the `span` **to show** `#select` should be not empty AND not have a value of 'Foo'.  
I'm setting the default condition to the logic `or` to follow along WordPress standard.


src/resources/js/dependency.js
```diff
@@ -44,26 +70,35 @@
...
+						return is_applicable;
+					} );
+
+					if ( condition_relation === 'and' ) {
+						passes = _.reduce( constraints, function ( passes, constraint, key ) {
+							return passes && constraint_conditions[key]( value, constraint, $field );
+						}, true );
+					} else {
+						passes = _.reduce( constraints, function ( passes, constraint, key ) {
+							return passes || constraint_conditions[key]( value, constraint, $field );
```
bordoni: spacing around `[key]`

src/resources/js/dependency.js
```diff
@@ -44,26 +70,35 @@
...
+						condition_relation = $dependent.data( 'relation' ) || 'and',
+						passes;
+
+					constraints = _.pick( constraints, function ( is_applicable ) {
+						return is_applicable;
+					} );
+
+					if ( condition_relation === 'and' ) {
+						passes = _.reduce( constraints, function ( passes, constraint, key ) {
+							return passes && constraint_conditions[key]( value, constraint, $field );
```
bordoni: spacing around `[key]`

src/resources/js/dependency.js
```diff
@@ -44,26 +70,35 @@
...
+						}, true );
+					} else {
+						passes = _.reduce( constraints, function ( passes, constraint, key ) {
+							return passes || constraint_conditions[key]( value, constraint, $field );
+						}, false );
+					}
+
+					if ( passes && !is_disabled ) {
```
bordoni: Spacing after `!`

lucatume:
changes in

bordoni:
👍

PR #280, opened 2016-11-17
Auto-size dropdowns, style optgroups, and avoid dupe dependency triggers

borkweb:
See: R125 findings sheet


src/resources/js/dependency.js
```diff
@@ -33,7 +33,7 @@
 				}
 
 				// Fetch dependent elements
-				var $dependents = $document.find( '[data-depends="' + selector + '"]' );
+				var $dependents = $document.find( '[data-depends="' + selector + '"]' ).not( '[id^="s2id_autogen"]' );
```
bordoni: it would be a _"faster"_ check if we looked for `.select2-container`

borkweb:
Switched.

bordoni:
👍

PR #282, opened 2016-11-17
keep network check in mind when checking key

lucatume:

Avoid storing the license network key in subsites.


barryhughes:
:dolphin:

PR #284, opened 2016-11-21
add support for checked condition

lucatume:
Ticket: none

Carry over the `condition-is-checked` data attribute when "select2izing" `select`s


barryhughes:
:train:

PR #285, opened 2016-11-22
Make our common dropdowns methods publicly accessible

barryhughes:
Make dropdown methods accessible from other code via `tribe_dropdowns`.

I'm assuming all of the current `obj.*` methods can be public here, but actually all I really need is `obj.dropdown` - if the remainder ought to be made 'private' that's doable.

[#69686/R125 findings](https://central.tri.be/issues/69686)


borkweb:
:+1:

PR #286, opened 2016-11-23
Re-organized Select + Sticky Search

bordoni:
_Ref:_ [C#38129](https://central.tri.be/issues/38129)


src/resources/js/dropdowns.js
```diff
@@ -14,307 +14,361 @@ var tribe_dropdowns = tribe_dropdowns || {};
...
-				if ( ! result && 'undefined' !== typeof args.tags ){
-					var possible = _.where( args.tags, { text: text } );
-					if ( args.tags.length > 0  && _.isObject( possible ) ){
-						var test_value = obj.search_id( possible[0] );
-						result = test_value.toUpperCase().indexOf( term.toUpperCase() ) == 0;
-					}
-				}
+		// Don't Remove HTML elements or escape elements
+		if ( $select.is( '[data-allow-html]' ) ) {
+			args.escapeMarkup = obj.allow_html_markup;
```
barryhughes: Here and a few other places it looks like we're assigning the function itself (to `args.escapeMarkup` in this case) rather than the function result - is that deliberate or something that needs to be fixed?

src/resources/js/dropdowns.js
```diff
@@ -14,307 +14,361 @@ var tribe_dropdowns = tribe_dropdowns || {};
...
-			if ( ! $select.is( '[multiple]' ) ){
-				return;
-			}
-			if ( ! $select.is( '[data-source]' ) ){
-				return;
-			}
+			// By default only send the source
+			args.ajax.data = function( search, page ) {
+				return {
+					action: 'tribe_aggregator_dropdown_' + source,
```
barryhughes: Should be `tribe_*` rather than `tribe_aggregator_*`?

src/resources/js/dropdowns.js
bordoni: Yes That is what we want for sure.

src/resources/js/dropdowns.js
barryhughes: This (piece of JS as a whole, if not this specific part of it) is already used by more than just aggregator - which makes sense as it's in tribe-common. Isn't it a recipe for trouble if in future we want other components to listen out for an act on this?

bordoni:
Rdy for another pair of 👀

barryhughes:
:bird:

PR #287, opened 2016-11-23
remove top margin from field first child

lucatume:
Ticket: https://central.tri.be/issues/62031#note-16

Remove top margin from first child or tribe field wrap content.


bordoni:
😎🤘🏻

PR #288, opened 2016-11-24
Add tribe_is_truthy() for truthiness comparisons | spotfix

barryhughes:
General purpose is-this-thing-truthy function to avoid per-class replication of the same logic.


bordoni:
👍

PR #289, opened 2016-11-26
Before using any Core only Classes we need to check if it's available

bordoni:
_Ref:_ [C#69860](https://central.tri.be/issues/69860)


tr1b0t:

PR #290, opened 2016-11-29
explicitly set the display attribute

lucatume:
Ticket: https://central.tri.be/issues/69857

Any browser will cause Select2 container inline `style` attribute to be set to `inline-block`. But no Firefox.
Let's be explicit about it.


src/resources/js/dropdowns.js
```diff
@@ -268,6 +268,8 @@ var tribe_dropdowns = tribe_dropdowns || {};
 		// Save data on Dropdown
 		$select.data( 'dropdown', args );
 
+		args.containerCss = { 'display': 'inline-block' };
```
bordoni: Would you mind moving it up next to: `args.dropdownCss`?

lucatume:
Not at all, there it is.

PR #295, opened 2016-12-07
Search recursively for preselected items when setting up dropdowns

barryhughes:
https://central.tri.be/issues/70351


src/resources/js/dropdowns.js
```diff
@@ -83,38 +83,76 @@ var tribe_dropdowns = tribe_dropdowns || {};
...
+	 * object within the haystack that matches one of those two properties will be returned.
+	 *
+	 * If objects contain an array named 'children', then that array will also be searched.
+	 *
+	 * @param {Object} description
+	 * @param {Array}  haystack
+	 *
+	 * @return {Object|boolean}
+	 */
+	function find_item( description, haystack ) {
```
bordoni: Can you make it an prop of `obj`?

src/resources/js/dropdowns.js
```diff
@@ -83,38 +83,76 @@ var tribe_dropdowns = tribe_dropdowns || {};
...
+	 * @param {object}   $select
+	 * @param {function} make_selection
+	 */
+	obj.init_selection = function( $select, make_selection ) {
+		var is_multiple    = $select.is( '[data-multiple]' ),
+		    options        = $select.data( 'dropdown' ),
+		    current_values = $select.val().split( options.regexSplit ),
+		    selected_items = [];
+
+		$( current_values ).each( function() {
```
bordoni: So I was thinking we could use `_.reduce(list, iteratee, [memo], [context])` to apply a method somewhat like `find_item`.

bordoni: Avoiding having a loop that loops again.

src/resources/js/dropdowns.js
barryhughes: I thought about that. Is there a good reason to prefer that over essentially keeping it private/protected, though?

bordoni: Actually, I don't have a very good reasoning in mind. Just following a habit to store everything publicly to make it _easier_ to use it later by our own extensions.

barryhughes: Hmm, well definitely two different ways we can go here; my own feeling is we should avoid a large public surface area, same as in our server-side code where we default to private/protected levels of visibility.

Also, if we see it as a good general purpose function, maybe the dropdown lib isn't the ideal home for it anyway - else non-dropdown-specific code needs to include this to utilize the function).

It could possibly go in `tribe-common.js` if we wanted to do that, but I'm not sure a) if we need to do that now b) if it will make sense following resolution of your other note re using a reducer.

bordoni: Yeah correct, you make good points here. I think we can improve on this later.

bordoni:
👍

PR #297, opened 2016-12-09
Remove stray characters

Aetles:
These two stray characters showed up on page when using The Events Calendar 4.3.4.


bordoni:
_Ref:_ [C#70536](https://central.tri.be/issues/70536)

PR #301, opened 2016-12-13
Allow 'forced enqueuing' of registered assets

barryhughes:
At the moment, assets may be registered such that they are enqueued during a specific action and/or if a specific set of conditionals are met.

Sometimes, though, this prohibits reuse by other plugins in cases that weren't originally anticipated (or just where the base plugin doesn't have any knowledge of the dependents that might require its assets). 

This change allows `Tribe__Assets::enqueue()` to be called directly with one or more slugs which should be enqueued even if the original conditions haven't been met. The advantage of this over a direct call to `wp_enqueue_script()` is that any localization data set up at point of registration will also be handled.

https://central.tri.be/issues/70623

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -77,15 +77,31 @@ class Tribe__Assets {
 		}
 	}
 
-	public function enqueue() {
+	/**
+	 * Enqueues registered assets.
+	 *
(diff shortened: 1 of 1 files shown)


borkweb:
:+1:

PR #302, opened 2016-12-13
Fix/70623 common script data

barryhughes:
* Makes the `tribe_l10n_datatables` object available independently of whether `tribe-common.js` or any other single asset is enqueued
* Adds a helper class to achieve that (ie, add data to the page without localizing individual scripts)

https://central.tri.be/issues/70682

### DIFF
--- a/src/Tribe/Asset/Script_Data.php
+++ b/src/Tribe/Asset/Script_Data.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Handles adding script data to the page in cases where localizing a
+ * specific script is not suitable.
+ *
+ * Should generally be accessed via tribe( 'tribe.asset.script-data' )
+ * rather than via direct instantiation.
+ */
+class Tribe__Asset__Script_Data {
+	/**
+	 * Container for any JS data objects that should be added to the page.
+	 *
+	 * @var array
+	 */
+	protected $objects = array();
+
+	/**
+	 * Hooks up the method used to actually render the JSON data.
+	 */
+	public function hook() {
+		if ( is_admin() ) {
+			add_action( 'admin_footer', array( $this, 'render_json' ) );
+		} else {
+			add_action( 'wp_footer', array( $this, 'render_json' ) );
+		}
+	}
+
+	/**
+	 * Adds the provided data to the list of objects that should be available
+	 * to other scripts.
+	 *
+	 * @param string $object_name
+	 * @param array  $data
+	 */
+	public function add( $object_name, array $data ) {
+		$this->objects[ $object_name ] = $data;
+	}
+
+	/**
+	 * Outputs the
+	 * @internal
+	 */
+	public function render_json() {
+		if ( empty( $this->objects ) ) {
+			return;
+		}
+
+		echo '<script type=\'text/javascript\'> /* <![CDATA[ */';
+
+		foreach ( $this->objects as $object_name => $data ) {
+			echo "var $object_name = " . wp_json_encode( $data ) . ';';
+		}
+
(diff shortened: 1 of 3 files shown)


src/Tribe/Asset/Script_Data.php
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	/**
+	 * Adds the provided data to the list of objects that should be available
+	 * to other scripts.
+	 *
+	 * @param string $object_name
+	 * @param array  $data
+	 */
+	public function add( $object_name, array $data ) {
```
bordoni: Could data also be an Object right? I would allow anything and cast it as an `Array`.

src/Tribe/Asset/Script_Data.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * Handles adding script data to the page in cases where localizing a
+ * specific script is not suitable.
+ *
+ * Should generally be accessed via tribe( 'tribe.asset.script-data' )
+ * rather than via direct instantiation.
+ */
+class Tribe__Asset__Script_Data {
```
bordoni: I would rename it to `Tribe__Asset__Data` since no CSS would have data attached to it.

src/Tribe/Asset/Script_Data.php
```diff
@@ -0,0 +1,56 @@
...
+	/**
+	 * Outputs the
+	 * @internal
+	 */
+	public function render_json() {
+		if ( empty( $this->objects ) ) {
+			return;
+		}
+
+		echo '<script type=\'text/javascript\'> /* <![CDATA[ */';
```
bordoni: Can we render each variable in a Single JS script, and add an ID Attribute like `id="tribe-asset-data-$object_name"`

src/Tribe/Asset/Script_Data.php
```diff
@@ -0,0 +1,56 @@
...
+
+		foreach ( $this->objects as $object_name => $data ) {
+			echo "var $object_name = " . wp_json_encode( $data ) . ';';
```
bordoni: Escape the Object name in some way?

src/Tribe/Asset/Script_Data.php
barryhughes: True - will change

src/Tribe/Asset/Script_Data.php
barryhughes: I guess it could - will change

src/Tribe/Asset/Script_Data.php
barryhughes: We can - are you thinking there will be cases where we want to manipulate it other than by interacting directly with the object itself, though?

src/Tribe/Asset/Script_Data.php
barryhughes: We can - I was actually taking my lead here from how WP does things (the name of the data to localize is not escaped, which makes sense since its entirely under our control) but extra safety isn't a bad thing

src/Tribe/Asset/Script_Data.php
bordoni: I can see this been a need on Customizer. To replace the contents of one of these scripts of remove then from the page.

barryhughes: Doesn't seem like that's the right way to go; once this has been sent to the browser the object is in memory even if you subsequently remove the script element.

We'd need to overwrite either before it hits the browser or define a new object of the same name - separate script elements wouldn't really help.

PR #307, opened 2016-12-24
Add a new Util for buttonset using JS

bordoni:
As per conversation with @zbtirrell and @leahkoerper, I've applied parts of the Mock from our new Designer on to recurrence.

This code is to enable the cool new checkbox like a button.

screen shot 2016-12-24 at 05 43 17

_Rel:_ https://github.com/moderntribe/events-pro/pull/539 https://github.com/moderntribe/the-events-calendar/pull/1233


src/resources/js/buttonset.js
```diff
@@ -0,0 +1,89 @@
...
+		buttonset: '.tribe-buttonset',
+		button: '.tribe-button-field',
+		input: '.tribe-button-input',
+		active: '.tribe-active'
+	};
+
+	obj.ready = function( event ) {
+		obj.$body = $( 'body' );
+		obj.$body.on( 'click.tribe_buttonset', obj.selector.button, obj.click );
+		obj.$body.on( 'change.tribe_buttonset', obj.selector.input, obj.change ).find( obj.selector.input ).trigger( 'change' );
```
barryhughes: Optional - could move the `.find().trigger()` portion to a new line - as together they form a distinct operation in terms of setting up the initial UI vs the preparatory step of setting up the event listener.

Plus it would shorten the line a little, it exceeds our soft limit of 120chars atm.

src/resources/js/buttonset.js
```diff
@@ -0,0 +1,89 @@
+var tribe_buttonset = tribe_buttonset || {};
+
+( function( $, obj ) {
+	'use strict';
+
+	obj.$body;
```
barryhughes: Could this be a local var rather than a 'public' `obj` property?

borkweb:
This looks good (other than Barry's thoughts). Adding hold so it waits until the first MR after 4.4

bordoni:
After a bunch of cleaning up, and work here I think it's good for a new Code Review. Do this with care, there is a lot of changes here.

borkweb:
:+1:

PR #310, opened 2017-01-09
Make container registration overwrite previous registrations.

lucatume:
Ticket: https://central.tri.be/issues/70530

To keep things easy let's make sure we replace any binding when binding stuff on the container.


bordoni:
Why include composer?

lucatume:
We need some of its libraries for the autoloader.

PR #317, opened 2017-01-19
Resolve the Circular Microtime bug from Windows servers

bordoni:
_Ref:_ [C#71912](https://central.tri.be/issues/71912)


lucatume:
Probably a changelog entry on TEC?
The fix is in there, good to go!

PR #323, opened 2017-02-01
Feature/67759 add filters to get start/end dates

niconerd:
Adjusted filter names to match convention. As the code is on 'common' the filter prefix should be 'tribe_' and not 'tribe_events'.

https://central.tri.be/issues/67759


barryhughes:
✔️

PR #326, opened 2017-02-04
print Customizer styles in head instead of footer

cliffordp:
re: https://central.tri.be/issues/42754

### DIFF
--- a/src/Tribe/Customizer.php
+++ b/src/Tribe/Customizer.php
@@ -119,7 +119,7 @@ final class Tribe__Customizer {
 		// Hook the Registering methods
 		add_action( 'customize_register', array( $this, 'register' ), 15 );
 
-		add_action( 'wp_print_footer_scripts', array( $this, 'print_css_template' ), 15 );
+		add_action( 'wp_head', array( $this, 'print_css_template' ), 15 );
 
 		add_filter( "default_option_{$this->ID}", array( $this, 'maybe_fallback_get_option' ) );
 	}


tr1b0t:

bordoni:
The problem that loading it on the `wp_head` it will create tons of problems with a few of the rules. At least it was creating when I initially create this Customizer code. It would be safer for this to be tested with all customizer fields before going forward.

cliffordp:
I tested with just changing Filter Bar color and that's it, since that's what the support ticket was regarding.
  

  
Feel free to QA it of course. :)
  
  

  
>

bordoni:
@cliffordp yeah but on this case it's important to test all of them because each field has different types of CSS rules, and they depend on when they are applied.

It is important to not only fix the problem from the support but also for the developer (person in charge) to take a look and test to see if the change doesn't break anything that was previously was working.

cliffordp:
Gotcha. I misread your previous comment.   
    
  
 If you want me to do the testing, let me know. Otherwise I'll leave it to you and whomever else.   

  
  
  

  
  
>

bordoni:
Yeah it's important for the developer to test all possible scenarios before passing it to the QA team, because if we can avoid the time lost sending it back to QA and then QA sending it back to us, it's a time well spent.

barryhughes:
Placing on hold (note added to [#42754](https://central.tri.be/issues/42754))

cliffordp:
Snippet to still accomplish this if desired:

https://gist.github.com/cliffordp/13c7a2e3f91158d5c9d0fca075b121fb

barryhughes:
Going to close this to reduce noise in our list of items waiting on review. Can re-open as needed, probably once we pull [#42754](https://central.tri.be/issues/42754) into a future MR.

PR #330, opened 2017-02-10
Stop using Tribe__PUE__Checker where not needed

sc0ttkclark:
Tribe__PUE__Checker hooks into things, does many other things. However, this case absolutely does not warrant the use of the whole class.

### DIFF
--- a/src/Tribe/Support.php
+++ b/src/Tribe/Support.php
@@ -380,11 +380,19 @@ if ( ! class_exists( 'Tribe__Support' ) ) {
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
 
 			$url   = $url ? $url : urlencode( str_replace( array( 'http://', 'https://' ), '', get_site_url() ) );
-			$pue   = new Tribe__PUE__Checker( 'https://theeventscalendar.com/', 'events-calendar' );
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
This allows us to test the `tribe_system/v2/customer-info/` endpoint on local TEC instances, so we can properly test changes and future work done with for SysInfo support work.

src/Tribe/Support.php
```diff
@@ -380,11 +380,19 @@ public static function ajax_sysinfo_optin() {
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
 
 			$url   = $url ? $url : urlencode( str_replace( array( 'http://', 'https://' ), '', get_site_url() ) );
-			$pue   = new Tribe__PUE__Checker( 'https://theeventscalendar.com/', 'events-calendar' );
-			$query = $pue->get_pue_update_url() . 'wp-json/tribe_system/v2/customer-info/' . $optin_key . '/' . $url . $remove;
+
+			$teccom_url = 'https://theeventscalendar.com/';
+
+			if ( defined( 'PUE_TEC_URL' ) ) {
```
borkweb: Let's switch this define to `TEC_URL`.

sc0ttkclark: Updated

PR #332, opened 2017-02-14
When fetching from post transients, if no data set, return false

borkweb:
In order to properly react to the state of post transients and deciding whether or not to perform logic, we need to know if there were values found in get_post_meta for the transient. If nothing was found, we should return `false` to match the state when the transient has expired.

Related: https://github.com/moderntribe/event-tickets/pull/340

See: https://central.tri.be/issues/70485


bordoni:
👍

PR #333, opened 2017-02-14
Add an int validator

borkweb:
Related: https://github.com/moderntribe/the-events-calendar/pull/1297

See: https://central.tri.be/issues/70497

### DIFF
--- a/src/Tribe/Validate.php
+++ b/src/Tribe/Validate.php
@@ -167,7 +167,7 @@ if ( ! class_exists( 'Tribe__Validate' ) ) {
 				$this->result->error = sprintf( esc_html__( '%s must contain numbers, letters and dots only', 'tribe-common' ), $this->label );
 			}
 		}
-		
+
 		/**
 		 * Validates a field as a string containing only letters,
 		 * numbers, dashes and underscores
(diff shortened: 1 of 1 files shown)


src/Tribe/Validate.php
```diff
@@ -227,6 +227,20 @@ public function positive_int() {
 		}
 
 		/**
+		 * validates a field as being an integer
+		 *
+		 * @return stdClass validation result object
+		 */
+		public function int() {
```
bordoni: If it checks it as a string it should be `numeric` right?

borkweb: I was trying to adhere to the naming convention of the other validation method: `positive_int()`. I just assume keep it as `int()`

barryhughes: I don't want to hold this up unnecessarily, so I'm going to drop this in as a musing more than anything.

To me, 'int' is a loaded term. For example, 17179869184 is clearly an integer but it would never be a 'true' integer in terms of the PHP type system if a 32-bit runtime is in use.

Could it be worth renaming the function - or easier yet, expanding on this slightly in the docblock?

bordoni: Exactly what @barryhughes mentioned, I mean. Thinking about the scope with 32bit systems and the fact that this uses a Regular Expression to check so It would validate any numeric value, not only valid integers.

borkweb: I've opted to add a comment that justifies the name.

barryhughes: ✔️

PR #335, opened 2017-02-16
allow returning the container

lucatume:
Return the container instance when using `tribe()` with no arguments and register service providers using `tribe_register_provider`


bordoni:
👍

PR #341, opened 2017-02-27
refine options of server mocker

lucatume:
This PR:

* makes the cleaner options available even if server mocking is disabled
* removes the setting to soft-set the ea server domain to avoid confusion with the `wp-config.php` set one


borkweb:
:+1:

PR #343, opened 2017-02-27
improved language for blank EA license message

zbtirrell:
See: https://central.tri.be/issues/73019

I also fixed some escaping and spacing for standards

### DIFF
--- a/src/Tribe/PUE/Checker.php
+++ b/src/Tribe/PUE/Checker.php
@@ -437,9 +437,6 @@ if ( ! class_exists( 'Tribe__PUE__Checker' ) ) {
 					'network_option'  => true,
 				);
 			} elseif ( $this->should_show_subsite_editable_license() ) {
-
-
-
 				$to_insert[ $this->pue_install_key ] = array(
 					'type'            => 'license_key',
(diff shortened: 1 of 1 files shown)


src/Tribe/PUE/Checker.php
```diff
@@ -413,36 +413,44 @@ public function do_license_key_fields( $fields ) {
...
+					'tooltip'         => $no_license_tooltip,
 					'parent_option'   => false,
 					'network_option'  => true,
 				);
-			} elseif($this->should_show_subsite_editable_license()){
-				$to_insert[$this->pue_install_key ] = array(
+			} elseif ( $this->should_show_subsite_editable_license() ) {
+
+
+
```
borkweb: I would recommend 12 more line breaks or 2 fewer.

borkweb:
But it wouldn't be horrid to have those extra line breaks - so I'm not going to hold up this PR for that suggestion.

src/Tribe/PUE/Checker.php
zbtirrell: That's where I was nagging my head against the keyboard (apparently)

PR #345, opened 2017-03-15
http should be https in Coordinates_Provider

trishasalas:
https://central.tri.be/issues/74930

_see https://github.com/moderntribe/the-events-calendar/pull/1323 for changelog entry_

### DIFF
--- a/src/Tribe/Utils/Coordinates_Provider.php
+++ b/src/Tribe/Utils/Coordinates_Provider.php
@@ -106,7 +106,7 @@ class Tribe__Utils__Coordinates_Provider {
 	 * @return null|WP_Http
 	 */
 	public function get_http() {
-		return $this->http;
+		return $this->https;
 	}
 
 	protected function get_google_api_base() {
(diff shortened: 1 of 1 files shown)


src/Tribe/Utils/Coordinates_Provider.php
```diff
@@ -31,7 +31,7 @@ class Tribe__Utils__Coordinates_Provider {
 	/**
 	 * @var WP_Http
 	 */
-	private $http;
+	private $https;
```
barryhughes: This is really just a small quibble, but I'd probably have left the property name as `http` despite the change in default protocol to `https://` on line 14, just because that's consistent with the object type (`WP_Http`).

It's also slightly inconsistent with the name of existing method `get_http()` (which we can't really change without deprecating and replacing it).

trishasalas: Totally see your point here, reverted.

barryhughes:
Great! Also, can we get a changelog entry in here? (Apologies for not highlighting this first time round)

PR #351, opened 2017-03-21
change http urls to https

zbtirrell:
See: https://central.tri.be/issues/72273

I also fixed a few URLs that were redirecting


bordoni:
👍

PR #357, opened 2017-04-12
Fix/68437 updates available message

elimn:
Ticket: https://central.tri.be/issues/68437

Fixed two things

1. The pue-slug does not always match the WP slug that gets generated on the plugins.php page, so switched to using the plugin file which always matches.
2. We are completely missing an update available message on the PUE server responses, and within our own plugin. So I added one.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | String &quot;&lt;/td&gt;&lt;/tr&gt;&quot; does not require double quotes; use single quotes instead | [Checker.php:825](https://github.com/moderntribe/tribe-common/blob/29b638285a94c3b9f8b371539bd1b8afb6c7f1f9/src/Tribe/PUE/Checker.php#L825)

src/Tribe/PUE/Checker.php
```diff
@@ -780,23 +780,58 @@ public function maybe_display_json_error_on_plugins_page( $page ) {
...
-				'slug' => $this->get_slug(),
-				'message_row_html' => "
-					<tr class='plugin-update-tr active'> <td colspan='3' class='plugin-update'>
-						<div class='update-message notice inline notice-warning notice-alt'>
-							{$state->update->license_error}
-						</div>
-					</td> </tr>
-				",
-			);
+			if ( ! empty( $messages ) ) {
```
borkweb: I'd love to see this changed to:

```
if ( empty( $messages ) ) {
  return;
}
```

That way the meat of what we are expecting to have happen, isn't nested in.

borkweb:
Other than that one change, this looks pretty darn solid :+1:

The change I mentioned isn't _necessary_ - just desirable. Proceed as you see fit.

elimn:
Good idea o @borkweb . Done. Thanks!

PR #359, opened 2017-04-14
rename var

ggwicz:
this fixes some notices that came out of QAing image widget plus


borkweb:
So complex! The most complex code change you've ever seen! Nobody makes complex changes like this.

PR #362, opened 2017-04-17
Updating spirit animal to a wild goat

borkweb:


barryhughes:
Seems slightly odd that there are datatable and datepicker CSS changes (without matching .pcss changes), but probably just a side-effect of recompiling.

borkweb:
Yup. Just a side effect of a recompilation.

PR #376, opened 2017-04-27
Added check if the log file is writable

elimn:
Per [this forum topic](https://theeventscalendar.com/support/forums/topic/error-when-activating-events-calendar-pro/page/2/#post-1231901) let's check if the log file is writable as well as readable. In some server configs this will prevent errors.

### DIFF
--- a/src/Tribe/Log/File_Logger.php
+++ b/src/Tribe/Log/File_Logger.php
@@ -64,7 +64,12 @@ class Tribe__Log__File_Logger implements Tribe__Log__Logger {
 			touch( $this->log_file );
 		}
 
-		if ( is_readable( $this->log_file ) && is_writable( $this->log_file ) ) {
+		// Bail if we're attempting to write but don't have permission.
+		if ( 'r' !== $this->context && ! is_writable( $this->log_file ) ) {
+			return;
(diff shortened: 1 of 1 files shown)


src/Tribe/Log/File_Logger.php
```diff
@@ -64,7 +64,7 @@ protected function obtain_handle() {
 			touch( $this->log_file );
 		}
 
-		if ( is_readable( $this->log_file ) ) {
+		if ( is_readable( $this->log_file ) && is_writable( $this->log_file ) ) {
```
barryhughes: What about something like:

```
$is_writable_if_needed = 'r' !== $this->context && is_writable( $this->log_file );

if ( is_readable( $this->log_file ) && $is_writable_if_needed ) {
    /* ... existing code */
}
```

That variable name may need some work, but my thinking is that if the log file is opened in _read_ mode, rather than _append,_ we don't actually care about writability and in fact we'd want to open it even if it is not writable.

elimn: That's a great idea. I'll enhance.

bordoni:
👍

PR #378, opened 2017-05-02
Backslash the sprintf var names (which use double quotes)

sc0ttkclark:
Fixes issues like:

```
Notice: Undefined variable: s in wp-content/plugins/the-events-calendar/common/src/Tribe/PUE/Notices.php on line 222
Notice: Undefined variable: sfrom in wp-content/plugins/the-events-calendar/common/src/Tribe/PUE/Notices.php on line 222
Notice: Undefined variable: s in wp-content/plugins/the-events-calendar/common/src/Tribe/PUE/Notices.php on line 222
```


bordoni:
👍

PR #380, opened 2017-05-02
change textdomain from tribe-common to the-events-calendar

sebastienserre:


bordoni:
Hi @sebastienserre thanks for wanting to contribute to our `tribe-common` repository.
I don't think I follow what you were trying to accomplish by changing the Text Domain for those strings. Could you share a little bit of information as to why you needed that?

We are inside of the `tribe-common` project so that seems that the correct domain for those strings. But I might not be seen something here.

sebastienserre:
Hello, 
I'll try to explain:
Here's a translation string in GlotPress:
https://translate.wordpress.org/projects/wp-plugins/the-events-calendar/stable/fr/default?filters%5Bterm%5D=The+license+key+you+received+when+completing+your+purchase&filters%5Buser_login%5D=&filters%5Bstatus%5D=either&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc

This one is always in english even if we have translated it as the text domain is "tribe-common" : 

`esc_html__( 'The license key you received when completing your purchase from %1$s will grant you access to support and updates until it expires. You do not need to enter the key below for the plugins to work, but you will need to enter it to get automatic updates. %3$sFind your license keys at %2$s%4$s.', 'tribe-common' ),`

but WordPress is waiting after the-events-calendar textdomain, not tribe-common.

So we're transating but:
https://dl.dropboxusercontent.com/1/view/sdis332qc23jedn/Apps/Shutter/capture_125.jpg

text is allways in english... 


I've the problem as French but it's the case in all the locale I tink.

barryhughes:
We appreciate you taking the time to submit this, @sebastienserre - but I'm afraid we cannot accept it: the correct text domain for this project is `'tribe-common'` and not `'the-events-calendar'` (which is just one of the plugins it is used within).

You are of course welcome to help with the translation effort, though, which is [handled here](http://translations.theeventscalendar.com/projects/tribe-common).

Thanks!

sebastienserre:
ok but as the textDomain isn't the right one for WordPress, the strings are not translated...
Why don't you separate tribe-common from The Events Calendar and force download of Tribe-common on TEC activation. We'll translate in our locale tribe-common on wp.org and it will be  translated in all WP install!

zbtirrell:
WordPress.org guidelines do not allow plugins like our tribe-common to be listed there, so, we cannot do those translations via WordPress.org.

We do however have the translations available here: http://translations.theeventscalendar.com/projects/tribe-common

We would welcome your contributions there!

PR #386, opened 2017-05-10
#35822 - Enable slugs instead of ids for post tags

jesseeproductions:
_Ref:_ [C#35822](https://central.tri.be/issues/35822)

Post tags were saving as the selected id  and create a new tag each time with that id as the tag name.

https://central.tri.be/issues/35822#note-32

wp_update_post  uses the slug to save and not the id like event categories. 

This PR along with this one:

https://github.com/moderntribe/events-community/pull/230

Changes that.


borkweb:
:+1:

PR #393, opened 2017-05-17
Fix what broken tests i've found

lucatume:
Ticket: https://central.tri.be/issues/78582


bordoni:
Haha interesting finding

PR #396, opened 2017-05-19
Revert "Fix tests and update di52 to 2.0.6"

borkweb:
1) This shouldn't have been merged into master (I missed it in the Code Review)
2) This is causing fatals because most of the methods in the di52 Container class are duplicated

Reverts moderntribe/tribe-common#384


borkweb:
cc @lucatume

elimn:
The MR 17.9 branch has the same problem. We should merge master into it after this is accepted.

elimn:
Oh my bad this is merged. We just need to fast forward Common in the ET master and MR branches.

PR #399, opened 2017-05-22
Add ArrayAccess support to the Tribe_Cache class

lucatume:
Ticket: https://central.tri.be/issues/67581

This PR modifies the `Tribe__Cache` class implementation adding support for an `ArrayAccess` API and array cache like behaviour.
It also fixes the non persistent caching behaviour: it  would always return the cache as non-existent.

I'm PRing to `master` but am open to suggestion about the correct branch to use.


borkweb:
:+1:

PR #403, opened 2017-05-26
Avoid querying the DB when priming Meta Chunker cache

lucatume:
Ticket: https://central.tri.be/issues/79593

As correctly reported in the ticket and on the support forum querying the DB on each run would kill larger DBs and defeat the purpose.
This PR stores the post IDsa and meta keys in an option to avoid performance issues.


src/Tribe/Meta/Chunker.php
```diff
@@ -93,19 +94,18 @@ public function hook() {
...
+
+		$chunked_keys = get_option( $this->chunked_keys_option_name );
+
+		if ( empty( $chunked_keys ) ) {
+			return;
+		}
+
+		foreach ( $chunked_keys as $post_id => $keys ) {
+			foreach ( $keys as $key ) {
+				$this->chunks_cache[ $this->get_key( $post_id, $key ) ] = null;
```
barryhughes: I'm probably missing something, but why do this? It seems like any time we're going to work with this array, we do an array key exists check in any case. Is there a further benefit to populating this array pre-emptively each request?

src/Tribe/Meta/Chunker.php
```diff
@@ -34,6 +34,7 @@
  * or filter the `tribe_meta_chunker_post_types` filter.
  */
 class Tribe__Meta__Chunker {
+	public $chunked_keys_option_name = '_tribe_chunker_chunked_keys';
```
barryhughes: Does this need to remain public (we have a getter already - we could add a setter if required, too)?

src/Tribe/Meta/Chunker.php
lucatume: This test https://github.com/moderntribe/tribe-common/blob/9d8c8c75789d697ce91d2da1e8b89b86920f9956/tests/wpunit/Tribe/Meta/ChunkerTest.php#L281-L281 clarifies it.
If nothing is being tagged as chunkable in the context of the request the cache will never be primed and any meta chunked, and registered as chunkable, in a previous request would not be treated as such.

barryhughes: Still not sure I follow: doesn't The Events Calendar create an instance of this class every request? Every time that happens, unless `$this->chunked_keys_option_name` is empty, the cache will be primed - regardless of whether any chunking takes place, right?

lucatume: Yes, an instance is created on each run.
Request 1: post ID `2` and meta key `foo` are set as chunkable hence the option is updated to store hte `2+foo` couple; large meta is added via `update_post_meta` and chunked.
Request 2: different request flow, the code registering the `2+foo` as chunkable is not hit hence the cache is not primed (removing the priming) hence a `get_post_meta(2,'foo');` request would yield broken meta as the chunker is not gluing stuff.

lucatume:
This should make it

barryhughes:
You are both a mensch and a doyen.

lucatume:
I hope this is not an insult.

barryhughes:
Never :-)

PR #404, opened 2017-05-29
Meta Chunker: update the option on post deletion

lucatume:
Ticket: https://central.tri.be/issues/79593

This PR makes sure proper cleanup is in place when a post is deleted

### DIFF
--- a/src/Tribe/Meta/Chunker.php
+++ b/src/Tribe/Meta/Chunker.php
@@ -857,7 +857,7 @@ class Tribe__Meta__Chunker {
 		$this->prime_chunks_cache();
 
 		foreach ( $this->chunks_cache as $key => $value ) {
-			if ( 0 === strpos( $key, '' . $post_id ) ) {
+			if ( 0 === strpos( $key, (string) $post_id ) ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Meta/Chunker.php
```diff
@@ -846,4 +847,25 @@ public function get_key_option_name() {
...
+	/**
+	 * Removes the entries associated with a deleted post from the cache and the database option.
+	 *
+	 * @param int $post_id A post ID
+	 */
+	public function remove_post_entry( $post_id ) {
+		$this->prime_chunks_cache();
+
+		foreach ( $this->chunks_cache as $key => $value ) {
+			if ( 0 === strpos( $key, '' . $post_id ) ) {
```
bordoni: Why append an empty thing instead of casting it as string?

lucatume: faster usually but really makes no difference here

bordoni: Up to you ;)

lucatume: Let's go with clarity (your option)

PR #412, opened 2017-06-08
#45569 | Add proposed tribe_events_set_notice filter

ggwicz:
_Ref:_ [C#45569](http://central.tri.be/issues/45569)

Lets you more easily customize the inner text of `Tribe__Notice` notices — current filters and methods require customizing much of the wrapping HTML in addition to that text.

An example of how you could use this filter is as follows:

```
add_filter( 'tribe_events_set_notice', 'tribe_45569_example', 10, 2 );

function tribe_45569_example( $notice, $key ) {

	if ( 'event-search-no-results' === $key ) {

                $example_url = 'something'; // Just an example here.

		$notice = sprintf( __( 'Oh no, nothing was found! <a href="%s">Back to the main events page</a>.', 'example' ), $example_url );
	}

	return $notice;
}
```

### DIFF
--- a/src/Tribe/Notices.php
+++ b/src/Tribe/Notices.php
@@ -16,6 +16,17 @@ class Tribe__Notices {
 	 * @return bool
 	 */
 	public static function set_notice( $key, $notice ) {
+
+		/**
+		 * Make notices filterable.
+		 *
+		 * @param string $notice The notice text.
+		 * @param string $key The key of the notice being filtered.
+		 *
+		 * @return string.
+		 */
+		$notice = apply_filters( 'tribe_events_set_notice', $notice, $key );
+
 		self::instance()->notices[ $key ] = $notice;
 
 		return true;
--- a/src/Tribe/Notices.php
+++ b/src/Tribe/Notices.php
@@ -18,7 +18,9 @@ class Tribe__Notices {
 	public static function set_notice( $key, $notice ) {
 
 		/**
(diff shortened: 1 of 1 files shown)


src/Tribe/Notices.php
```diff
@@ -16,6 +16,17 @@ class Tribe__Notices {
 	 * @return bool
 	 */
 	public static function set_notice( $key, $notice ) {
+
+		/**
+		 * Make notices filterable.
```
barryhughes: The addition of an `apply_filters()` call does indeed make the notices filterable but (maybe this is just me) it seems like a slightly odd description to use in the docblock.

Thinking of, say, a 3rd party dev reading this not in the source code but via our technical docs, might a description such as _"Provides an opportunity to alter the text of admin notices"_ provide a little more context?

barryhughes:
Left a note but it's not a blocker to merge: feel free to review and address (or not) as you see fit then merge :+1:

src/Tribe/Notices.php
bordoni: Add a `@since M17.11`.

barryhughes: (Though, we know the version number so you could add it directly rather than using the M17.11 placeholder)

bordoni: You are right! ;)

ggwicz:
Thanks guys, have implemented the feedback.

PR #417, opened 2017-06-14
Don't default to Meta Chunking post post_types

borkweb:
Instead, let's default to nothing.

See: https://central.tri.be/issues/80857


borkweb:
The Chunker tests continue to pass with this change. Merging.

PR #418, opened 2017-06-17
When testing if a file exists, ensure it's a local file on the filesystem

barryhughes:
Avoids a potential notice level error where the HTTP(S)/FTP stream wrapper isn't available by simplifying a check.

If `$package` is empty, or is not a remote URL, returns false. Seems there's need to do a further confirmation check by testing if the file exists (which tries to use stream wrappers when available).

:ticket: [#80492](https://central.tri.be/issues/80492)

### DIFF
--- a/src/Tribe/PUE/Package_Handler.php
+++ b/src/Tribe/PUE/Package_Handler.php
@@ -70,9 +70,9 @@ class Tribe__PUE__Package_Handler {
 	 * @return bool
 	 */
 	protected function is_mt_package( $package ) {
-		if ( empty( $package )
-		     || ! preg_match( '!^(http|https|ftp)://!i', $package )
-		     || ( is_file( $package ) && file_exists( $package ) )
+		if (
+			empty( $package )
+			|| ! preg_match( '!^(http|https|ftp)://!i', $package )
 		) {
 			return false;
 		}
--- a/src/Tribe/PUE/Package_Handler.php
+++ b/src/Tribe/PUE/Package_Handler.php
@@ -72,7 +72,7 @@ class Tribe__PUE__Package_Handler {
 	protected function is_mt_package( $package ) {
 		if ( empty( $package )
 		     || ! preg_match( '!^(http|https|ftp)://!i', $package )
-		     || file_exists( $package )
+		     || ( is_file( $package ) && file_exists( $package ) )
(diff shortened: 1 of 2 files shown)


src/Tribe/PUE/Package_Handler.php
```diff
@@ -70,9 +70,9 @@ public function filter_upgrader_pre_download( $reply, $package, WP_Upgrader $upg
 	 * @return bool
 	 */
 	protected function is_mt_package( $package ) {
-		if ( empty( $package )
-		     || ! preg_match( '!^(http|https|ftp)://!i', $package )
-		     || file_exists( $package )
```
borkweb: The removal of this line gives me pause. Your PR description reads:

> If `$package` is empty, or is not a remote URL, returns false. Seems there's need to do a further confirmation check by testing if the file exists (which tries to use stream wrappers when available).

Shouldn't we continue to handle the `file_exists()` case then?

barryhughes: Well, we've got three tests here and because of lazy evaluation we're only going to reach the `file_exists()` test if `$package` is a non-empty string that looks like a URL (because if it _doesn't_ look like a URL, we'll already have bailed).

But do we really want to perform a stat on a remote URL? I don't think we do, I think the intention was to check if `$package` exists in the local filesystem (yet we already know it is a remote URL by this point, else we'd have bailed out already).

Also ... my understanding - and local testing bears this out - is that stat-like operations aren't supported by the HTTP(S) stream wrappers, so if we _do_ want to make an HTTP request and determine if `$package` returns a 200 OK we'd want something more suited to that purpose in here, I'd have thought.

If you've still got reservations though I'm happy to rework: we could capture the match from the regex then see if that is in the list of returned stream wrappers _prior_ to calling `file_exists()`.

borkweb: Well...you can actually ignore me completely. I totally missed the `!` preceding the `preg_match( '!^(http|https|ftp)://!i', $package )`. I read that as: empty || is URL || file exists.

Sorry to make you type all of that - though it was a good read :)

barryhughes: Hah - glad you liked it :+1:

PR #422, opened 2017-06-20
Fixed App Shop page, and tribe_resource_url()

elimn:
Ticket: https://central.tri.be/issues/77975

tribe_resource_url() was returning a path in place of a URL everywhere it was being used.It was primarily being used on the App Shop page, which completely broke without anyone noticing.


src/Tribe/App_Shop.php
```diff
@@ -77,11 +79,28 @@ public function add_toolbar_item() {
...
+		/**
 		 * Enqueue the styles and script
 		 */
 		public function enqueue() {
-			wp_enqueue_style( 'app-shop', tribe_resource_url( 'app-shop.css', false, null, Tribe__Main::instance() ), array(), apply_filters( 'tribe_events_css_version', Tribe__Main::VERSION ) );
-			wp_enqueue_script( 'app-shop', tribe_resource_url( 'app-shop.js', false, null, Tribe__Main::instance() ), array(), apply_filters( 'tribe_events_js_version', Tribe__Main::VERSION ) );
+
+
+
+			wp_enqueue_style( 'app-shop-css' );
```
barryhughes: Is this method / are these `wp_enqueue_*()` calls required now we've switched to `tribe_assets()` - and if they are can we kill all the extra whitespace above them?

elimn: Who doesn't love extra whitespace randomly sprinkled about?

These assets will not get enqueued without those calls. That said, is there a way to make tribe_asset() automatically enqueue something on a given action? If there is, I couldn't quite figure out how to make it work.

borkweb: Yessir! Here's the docblock for [`tribe_asset()`](https://github.com/moderntribe/tribe-common/blob/master/src/functions/template-tags/general.php#L518). The 5th parameter lets you specify an action.

src/Tribe/App_Shop.php
```diff
@@ -77,11 +79,25 @@ public function add_toolbar_item() {
...
+		 * Registers the plugin assets
+		 */
+		protected function register_assets() {
+			tribe_assets(
+				Tribe__Main::instance(),
+				array(
+					array( 'app-shop-css', 'app-shop.css' ),
+					array( 'app-shop-js', 'app-shop.js' ),
+				),
+				'admin_print_styles-' . $this->admin_page
```
barryhughes: I _think_ if we switch this out for `admin_enqueue_scripts` we'll be laughing all the way to the bank and can strip out the calls in the `enqueue()` method/remove that method entirely

barryhughes:
Needs changelog entry :memo:

src/Tribe/App_Shop.php
borkweb: Swapping that out for `admin_enqueue_scripts` would cause the styles to be added to every page, which is what Brook was attempting to avoid, I think.

borkweb: @elimn, @barryhughes mentioned in Slack that `$this->admin_page` is null. If that was set to the correct value, perhaps all would be right in the world?

elimn:
Thanks Barry. This is ready for a second review.

barryhughes:
:+1:

PR #424, opened 2017-06-21
Spotfix/tribe asset tweaks

elimn:
Spotfix that stemmed from a discussion with Gustavo.

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -141,8 +141,6 @@ class Tribe__Assets {
 			/**
 			 * Allows developers to hook-in and prevent an asset from been loaded
 			 *
-			 * Note: When you pass callables on the `$asset->conditionals` argument this will be hooked here
-			 *
 			 * @param bool   $enqueue If we should enqueue or not a given asset
 			 * @param object $asset   Which asset we are dealing with
 			 */
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -82,7 +82,7 @@ class Tribe__Assets {
 
 			// Enqueue the registered assets at the appropriate time
 			if ( did_action( $asset->action ) > 0 ) {
-				call_user_func( array( $this, 'enqueue' ) );
+				$this->enqueue();
 			} else {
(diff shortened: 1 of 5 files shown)


src/Tribe/Assets.php
```diff
@@ -72,8 +72,12 @@ public function register_in_wp( $assets = null ) {
 				continue;
 			}
 
-			// Now add an action to enqueue the registered assets
-			add_action( $asset->action, array( $this, 'enqueue' ), $asset->priority );
+			// Enqueue the registered assets at the appropriate time
+			if ( did_action( $asset->action ) > 0 ) {
+				call_user_func( array( $this, 'enqueue' ) );
```
barryhughes: Why call this through `call_user_func()` rather than directly?

elimn: I just really like to complicate things.

src/Tribe/Assets.php
```diff
@@ -154,13 +174,17 @@ public function enqueue( $forcibly_enqueue = null ) {
 			} else {
 				wp_enqueue_style( $asset->slug );
 			}
+
+			$this->assets[ $key ]->already_enqueued = true;
```
barryhughes: Could simplify to `$asset->already_enqueued = true;`

elimn: This method gets called many times, each time the `$this->assets` will be passed to the loop. In order for a change to one of the assets to persist it looks like I would need to make `$asset` a reference for that to work. Do you think that's preferable? It would be more concise, but sometimes people forget they're working with a reference. I could go either way on this one.

http://php.net/manual/en/control-structures.foreach.php

barryhughes: It's an object - so you're acting on a reference/alias in any case.

bordoni: Ohh good catch @barryhughes we should use `$asset` it's a reference to the object.

elimn: 👍 Done. Didn't realize objects were references in a foreach assignment. Thanks for the tip.

PR #430, opened 2017-06-23
Data required by tribe-common.js needs to be provided to tribe-common…

barryhughes:
`tribe-common.js` expects and needs the `tribe_system_info` object - so let's localize that script, instead of a bunch of other scripts which don't truly need that object.

:ticket: [#81456](https://central.tri.be/issues/81456)


ggwicz:
no changelog entry required, as noted by barry [here](https://tribe.slack.com/archives/C3XGB48QK/p1498250233125127)

PR #435, opened 2017-06-28
Alter the range of admin screens in which the expired notices message…

barryhughes:
To avoid having to enqueue our assets on all admin screens, let's limit where the invalid key notice displays (to our own screens and to the plugin admin screen).

Can review and revise in the next MR.

:ticket: [#81260](https://central.tri.be/issues/81260)

### DIFF
--- a/src/Tribe/PUE/Notices.php
+++ b/src/Tribe/PUE/Notices.php
@@ -212,6 +212,15 @@ class Tribe__PUE__Notices {
 			return;
 		}
 
+		// Only show our invalid key messaging on the plugin admin screen and on our own admin screens
+		// @todo review and revise in MR 17.13
+		if (
+			'plugins.php' !== $pagenow
+			&& ! Tribe__Admin__Helpers::instance()->is_screen()
+		) {
+			return;
+		}
+
 		$plugin_names = $this->get_formatted_plugin_names( self::INVALID_KEY );
 
 		if ( empty( $plugin_names ) ) {


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task &quot;review and revise in MR 17.13&quot; | [Notices.php:216](https://github.com/moderntribe/tribe-common/blob/4bb6db983fa301111d2b4b719b03f5259a7ac34d/src/Tribe/PUE/Notices.php#L216)

src/Tribe/PUE/Notices.php
```diff
@@ -212,6 +212,15 @@ public function render_invalid_key() {
 			return;
 		}
 
+		// Only show our invalid key messaging on the plugin admin screen and on our own admin screens
+		// @todo review and revise in MR 17.13
+		if (
+			'plugins.php' !== $pagenow
```
bordoni: Can't we use the `is_screen` for that? By passing the page id you can use that helper class for this condition

barryhughes: This is consistent with the preceding lines and is temporary; can explore it if you think it's really better though

barryhughes: ... it also avoids two method calls to the same method

bordoni: We used it in the last PR, `is_screen( 'plugins' )`, I think it was by Brook.

bordoni: But ok.

PR #438, opened 2017-07-02
Remove width property from 'Buy addon' buttons

abrain:
Due to the fixed width the button could not adapt to the content. The german caption is longer than the english one (surprise ;) ) and does not fit into these 123px. The same goes for the disabled state of the button and its 100px. 

Before and after:

bildschirmfoto 2017-07-02 um 11 22 16

bildschirmfoto 2017-07-02 um 11 24 14

That's why I propose to remove the width property and let the browser define the width of the button.

P.S.: The `gulp` task brought up some more changes (mainly removal of `-webkit` properties in other CSS files which are unrelated to this change), which I didn't submit here.

See: https://central.tri.be/issues/88868


ggwicz:
Thanks a lot for submitting these changes, @abrain. We have merged them into Tribe Common, and as can be seen over in #537, we note the change in our changelog and thank you for submitting it.

This will all eventually be officially released some time in the next two or three weeks.

Cheers,
George

PR #439, opened 2017-07-03
Helper to retrieve anonymous objects registered as callbacks

barryhughes:
Sometimes callbacks belonging to anonymous objects are registered, leaving us in a tricky position if we want to unhook them. This helper provides a means of grabbing the needed object reference.

:ticket: [#74938](https://central.tri.be/issues/74938)

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -262,3 +262,49 @@ if ( ! function_exists( 'tribe_is_error' ) ) {
 		return ( $thing instanceof Tribe__Error || is_wp_error( $thing ) );
 	}
 }
+
+if ( ! function_exists( 'tribe_retrieve_object_by_hook' ) ) {
+	/**
+	 * Attempts to find and return an object of the specified type that is associated
+	 * with a specific hook.
+	 *
+	 * This is useful when third party code registers callbacks that belong to anonymous
+	 * objects and it isn't possible to obtain the reference any other way.
+	 *
+	 * @since TBD
+	 *
+	 * @param string   $class_name
+	 * @param string   $hook
+	 * @param int      $priority
+	 *
+	 * @return object|false
+	 */
+	function tribe_retrieve_object_by_hook( $class_name, $hook, $priority ) {
+		global $wp_filter;
+
+		// Not such hook/no callbacks registered for this hook?
+		if ( ! isset( $wp_filter[ $hook ] ) ) {
+			return false;
+		}
+
+		// Otherwise iterate through the registered callbacks at the specified priority
+		foreach ( $wp_filter[ $hook ]->callbacks[ $priority ] as $callback ) {
+			// Skip if this callback isn't an object method
+			if (
+				! is_array( $callback['function'] )
+				|| ! is_object( $callback['function'][ 0 ] )
+			) {
+				continue;
+			}
+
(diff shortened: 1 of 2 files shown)


src/functions/utils.php
```diff
@@ -262,3 +262,49 @@ function tribe_is_error( $thing ) {
...
+	function tribe_retrieve_object_by_hook( $class_name, $hook, $priority ) {
+		global $wp_filter;
+
+		// Not such hook/no callbacks registered for this hook?
+		if ( ! isset( $wp_filter[ $hook ] ) ) {
+			return false;
+		}
+
+		// Otherwise iterate through the registered callbacks at the specified priority
+		foreach ( $wp_filter[ $hook ]->callbacks[ $priority ] as $callback ) {
```
bordoni: Can we check if the given priority exists?

src/functions/utils.php
```diff
@@ -262,3 +262,49 @@ function tribe_is_error( $thing ) {
...
+			// Skip if this callback isn't an object method
+			if (
+				! is_array( $callback['function'] )
+				|| ! is_object( $callback['function'][ 0 ] )
```
bordoni: We don't need spaces around the `0` here.

src/functions/utils.php
```diff
@@ -262,3 +262,49 @@ function tribe_is_error( $thing ) {
...
+		// Not such hook/no callbacks registered for this hook?
+		if ( ! isset( $wp_filter[ $hook ] ) ) {
+			return false;
+		}
+
+		// Otherwise iterate through the registered callbacks at the specified priority
+		foreach ( $wp_filter[ $hook ]->callbacks[ $priority ] as $callback ) {
+			// Skip if this callback isn't an object method
+			if (
+				! is_array( $callback['function'] )
```
bordoni: Do we want just to bail if it's not an array? Would make it easier to read this whole foreach.

src/functions/utils.php
```diff
@@ -262,3 +262,49 @@ function tribe_is_error( $thing ) {
...
+				|| ! is_object( $callback['function'][ 0 ] )
+			) {
+				continue;
+			}
+
+			// If this isn't the callback we're looking for let's skip ahead
+			if ( $class_name !== get_class( $callback['function'][ 0 ] ) ) {
```
bordoni: Usage around `get_class` can be different on versions 5.2 and 7.0, can we double check it works as expected on both?

src/functions/utils.php
barryhughes: Good call - will add that :white_check_mark:

src/functions/utils.php
barryhughes: Good note - wasn't aware of that - but I don't believe it will be a problem here (it would only be an issue if we passed null)

src/functions/utils.php
barryhughes: Not sure I'm following - do you mean, only test if it's not an array and drop the `is_object()` test?

src/functions/utils.php
barryhughes: (Added, above)

src/functions/utils.php
borkweb: This foreach seems readable enough to me.

PR #442, opened 2017-07-04
Allow empty multichoice fields

barryhughes:
If an option has been _saved,_ don't assume the default value should be returned just because the saved value is null or empty.

:ticket: [#73183](https://central.tri.be/issues/73183)


elimn:
This has bugged me in the past. 👍

PR #443, opened 2017-07-06
Adds ability to test for radio is_checked (by ID)

Camwyn:
Allow us to test for checked state of radio buttons too - it was only allowing checkboxes.

Happy to add this to the MR?


barryhughes:
> Happy to add this to the MR?

:100:

bordoni:
Yey! Nice.

PR #444, opened 2017-07-06
Fix - radios don't trigger change when deselected

Camwyn:
This corrects the issue with the radios not triggering a change on being deselected when another is checked.

Ticket: https://central.tri.be/issues/82510

Expands on https://github.com/moderntribe/tribe-common/pull/443


src/resources/js/dependency.js
```diff
@@ -57,7 +76,7 @@
 						return ! $.isNumeric( val );
 					},
 					'is_checked': function ( _, __, $field ) {
-						return ( $field.is( ':checkbox' ) || $field.is( ':radio' ) ) ? $field.is( ':checked' ) : false;
+						return ( $field.is( ':checkbox' ) || $field.is( ':radio' ) ) ? $field.prop( 'checked' ) : false;
```
bordoni: We should convert this to boolean when returning `prop`. Because if it's not set, it can return `underfined`.

barryhughes:
Do we need a ticket/is any QA required?

src/resources/js/dependency.js
```diff
@@ -32,6 +33,20 @@
...
+				/**
+				 * If we're hooking to a radio, we need to make sure changing
+				 * any similarly _named_ ones trigger verify on all of them.
+				 * The base code only triggers on direct interations.
+				 * @since TBD
+				 */
+				if ( $field.is( ':radio' ) ) {
+					var $radios = $( "[name='" + $field.attr( 'name' ) + "']" );
+
+			    $radios.not( selectors.linked ).on( 'change', function() {
```
borkweb: Tabbing is off here.

borkweb:
@Camwyn can you create a Central ticket for this in the MR? Testing to make sure the field visibility in the Recurrence Rule form with the many rule combinations we support should be sufficient for this fix as that implementation of `dependency.js` is pretty thorough.

borkweb:
Oh! Can you also add a changelog entry in `readme.txt`?

src/resources/js/dependency.js
```diff
@@ -32,6 +33,20 @@
...
+				 * If we're hooking to a radio, we need to make sure changing
+				 * any similarly _named_ ones trigger verify on all of them.
+				 * The base code only triggers on direct interations.
+				 * @since TBD
+				 */
+				if ( $field.is( ':radio' ) ) {
+					var $radios = $( "[name='" + $field.attr( 'name' ) + "']" );
+
+					$radios.not( selectors.linked ).on( 'change', function() {
+							$radios.trigger( 'verify.dependency' );
```
barryhughes: Indented by one tab too many?

PR #446, opened 2017-07-07
Allow passing multiple localize-scripts to tribe-assets.

Camwyn:
Don't output a localized script more than once.

Ticket: https://central.tri.be/issues/81644

This was originally done in the `ticket-refresh bucket`, moving to the MR so others can benefit before that goes out.

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -25,6 +25,13 @@ class Tribe__Assets {
 	private $assets = array();
 
 	/**
+	 * Stores the localized scripts for reference
+	 *
+	 * @var array
+	 */
+	private $localized = array();
+
+	/**
 	 * Static Singleton Factory Method
 	 *
 	 * @since 4.3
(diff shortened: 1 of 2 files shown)


src/Tribe/Assets.php
```diff
@@ -169,7 +176,23 @@ public function enqueue( $forcibly_enqueue = null ) {
 
 				// Only localize on JS and if we have data
 				if ( ! empty( $asset->localize ) ) {
-					wp_localize_script( $asset->slug, $asset->localize->name, $asset->localize->data );
+					/**
+					 * check to ensure we haven't already localized it before
+					 * @since TBD
```
borkweb: Since this is now going out in the MR, you can update this TBD to `4.5.8`

src/Tribe/Assets.php
```diff
@@ -370,11 +393,17 @@ public function register( $origin, $slug, $file, $deps = array(), $action = null
 
 		// If you are passing localize, you need `name` and `data`
 		if ( ! empty( $asset->localize ) && ( is_array( $asset->localize ) || is_object( $asset->localize ) ) ) {
-			$asset->localize = (object) $asset->localize;
+			$asset->localize = (object) $asset->localize;if ( is_array( $asset->localize ) && empty( $asset->localize['name'] )  ) {
```
borkweb: Looks like the `if` was accidentally concatenated with the previous line.

PR #448, opened 2017-07-13
Modify how Verified field gets checked

bordoni:
_Ref:_ [C#82243](http://central.tri.be/issues/82243)
---


src/Tribe/Validator/Base.php
```diff
@@ -33,7 +33,30 @@ public function is_string( $value ) {
 	 * @return bool
 	 */
 	public function is_time( $value ) {
-		return is_numeric( $value ) || ( is_string( $value ) && strtotime( $value ) );
+		// If it is time we just know it's time
```
lucatume: Did you run the tests for this?
Rather than modifying an existing, and tested, method I'd create one extending it, e.g.
```php
public function is_datepicker_time( $value ) {
    if( $this->is_time( $value ) ) {
          return true;
    }

    // here make the checks for the datepicker specific format.
}
```

bordoni: Yeah I like that approach.

bordoni:
Ready for another 👁 👁

PR #455, opened 2017-07-21
Avoid noise if $_SERVER isn't populated | Spotfix

barryhughes:
Scratches a personal itch and avoids noise during WP CLI requests (end result is the same, we just eliminate the notice).


borkweb:
Add a changelog entry?

barryhughes:
Changelog entry added, good call

PR #457, opened 2017-07-21
Cleanup country list logic, avoid translation faux pas | #72113

barryhughes:
Stops us from translating a variable, which is a translation no-no. In this case, it was also unnecessary as the string had already been passed through `__()`.

:ticket: [#72113](https://central.tri.be/issues/72113)

### DIFF
--- a/src/Tribe/View_Helpers.php
+++ b/src/Tribe/View_Helpers.php
@@ -295,8 +295,8 @@ if ( ! class_exists( 'Tribe__View_Helpers' ) ) {
 				if ( $defaultCountry && $defaultCountry[0] != '' ) {
 					$selectCountry = array_shift( $countries );
 					asort( $countries );
-					$countries = array( $defaultCountry[0] => __( $defaultCountry[1], 'tribe-common' ) ) + $countries;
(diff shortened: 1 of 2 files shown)


src/Tribe/View_Helpers.php
```diff
@@ -295,8 +295,8 @@ public static function constructCountries( $postId = '', $useDefault = true ) {
 				if ( $defaultCountry && $defaultCountry[0] != '' ) {
 					$selectCountry = array_shift( $countries );
 					asort( $countries );
-					$countries = array( $defaultCountry[0] => __( $defaultCountry[1], 'tribe-common' ) ) + $countries;
-					$countries = array( '' => __( $selectCountry, 'tribe-common' ) ) + $countries;
+					$countries = array( $defaultCountry[0] => $defaultCountry[1] ) + $countries;
```
bordoni: should we move into using `array_merge`? For both.

barryhughes: Hmm, there's differences between the two (`array_merge()` vs an array union using `+`) though I guess in this case both would have the same result ... mostly though I just wanted to resolve the i18n issue and not change any other logic.

I'm happy to make the change but isn't `+` preferred? It's far more concise.

PR #460, opened 2017-07-27
Fixed select2 search returns exact matches only.

fabianmarz:
**Problem**:
The backend select2 search for venue/organizer returns exact matches only. This requires users to know the exact wording including special characters like quotes, apostrophes etc. which isn't very user friendly.

**Proposed solution:**
Change the checking to return matches which contains the term anywhere in the result not just at the beginning.

PS: There is no `develop` branch anymore. How is the current workflow for PRs?


barryhughes:
Thanks, @fabianmarz!

> PS: There is no develop branch anymore. How is the current workflow for PRs?

That's right - so targeting `master` as you've done is fine (and we'll modify the base branch as needed).

barryhughes:
Ticketed: #84496

PR #465, opened 2017-08-03
Enhance integration with plugin API/add new plugin | #82223

barryhughes:
Replaces a bunch of null values with reasonable defaults. Avoids zero star, last-updated-48-years ago scenarios.

:ticket: [#82223](https://central.tri.be/issues/82223)

### DIFF
--- a/src/Tribe/Plugins_API.php
+++ b/src/Tribe/Plugins_API.php
@@ -367,7 +367,7 @@ class Tribe__Plugins_API {
 	protected function fetch_current_stats() {
 		$response = wp_remote_get( $this->stats_url );
 
-		if ( is_wp_error( $response ) || '200' != $response['response']['code'] ) {
+		if ( is_wp_error( $response ) || 200 !== $response['response']['code'] ) {
 			return;
 		}
 
--- a/src/Tribe/Plugins_API.php
+++ b/src/Tribe/Plugins_API.php
@@ -102,7 +102,7 @@ class Tribe__Plugins_API {
 		$count = 0;
 		$products_to_inject = array();
 		foreach ( $products as $key => $product ) {
-			$products[ $key ] = $product = $this->build_product_data( $product, $key );
+			$products[ $key ] = $product = $this->build_product_data( $product );
 
 			// if the product is already installed, skip it
 			if ( $product['is_installed'] ) {
--- a/src/Tribe/Plugins_API.php
+++ b/src/Tribe/Plugins_API.php
@@ -2,32 +2,6 @@
 
 class Tribe__Plugins_API {
 	/**
-	 * Transient key used to store premium plugin stats.
-	 *
-	 * @since TBD
-	 *
-	 * @var string
-	 */
-	protected $cached_stats_key = '_tribe_plugins_api_stats';
-
-	/**
-	 * Our current (or most recent available) plugin statistics, including
-	 * ratings, download count etc.
-	 *
-	 * @var array
-	 */
-	protected $current_stats = array();
-
-	/**
-	 * Endpoint for discovering current plugin statistics.
-	 *
-	 * @since TBD
-	 *
-	 * @var string
-	 */
-	protected $stats_url = 'https://m.tri.be/plugin-stats';
-
-	/**
 	 * Static Singleton Factory Method
 	 *
 	 * @since 4.5.3
@@ -163,11 +137,10 @@ class Tribe__Plugins_API {
 	 * @since 4.5.3
 	 *
 	 * @param array  $product_data
-	 * @param string $product_slug
 	 *
 	 * @return array
 	 */
-	public function build_product_data( $product_data, $product_slug ) {
+	public function build_product_data( $product_data ) {
 		$defaults = array(
 			'name'                     => null,
 			'slug'                     => null,
@@ -202,8 +175,7 @@ class Tribe__Plugins_API {
 
 		$product = array_merge(
 			$defaults,
-			$product_data,
-			$this->get_sanitized_plugin_stats( $product_slug )
+			$product_data
 		);
 
 		if ( ! empty( $product['title'] ) && empty( $product['name'] ) ) {
@@ -233,8 +205,6 @@ class Tribe__Plugins_API {
 	 * @return array
 	 */
 	public function get_products() {
-		$this->get_current_stats();
-
 		$products = array(
 			'the-events-calendar' => array(
 				'title' => __( 'The Events Calendar', 'tribe-common' ),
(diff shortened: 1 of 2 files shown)


src/Tribe/Plugins_API.php
```diff
@@ -307,4 +340,67 @@ public function get_products() {
...
+	 * Fetches current plugin statistics from our remote endpoint.
+	 *
+	 * @since TBD
+	 *
+	 * @return array|mixed
+	 */
+	protected function fetch_current_stats() {
+		$response = wp_remote_get( $this->stats_url );
+
+		if ( is_wp_error( $response ) || '200' != $response['response']['code'] ) {
```
bordoni: Use `!==` here.

src/Tribe/Plugins_API.php
```diff
@@ -136,27 +162,28 @@ public function filter_api_result( $result, $action, $args ) {
...
 		$defaults = array(
 			'name'                     => null,
 			'slug'                     => null,
 			'version'                  => null,
 			'author'                   => '<a href="http://m.tri.be/19o3">Modern Tribe, Inc.</a>',
 			'author_profile'           => null,
 			'requires'                 => '3.9',
-			'tested'                   => '4.7.5',
-			'rating'                   => null,
+			'tested'                   => $GLOBALS['wp_version'],
```
borkweb: Isn't this the "tested up to" field? Shouldn't this be a hardcoded version of what we've tested up to rather than the current installed WP version?

borkweb: Or pulled from the readme, maybe?

src/Tribe/Plugins_API.php
```diff
@@ -2,6 +2,32 @@
...
+	protected $current_stats = array();
+
+	/**
+	 * Endpoint for discovering current plugin statistics.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected $stats_url = 'https://m.tri.be/plugin-stats';
```
zbtirrell: we can't do a call out to our domain without prior approval, it is against the rules.

barryhughes: I wondered about that.

barryhughes: If I strip out that logic, so we aren't connecting to our own domain, do the revised defaults seem reasonable to you? Some of them are clearly fictitious unless we randomize them, and I'm not too sure that's a great idea either.

zbtirrell: we could drop a JSON file in http://plugins.svn.wordpress.org/the-events-calendar/assets and reference that (without a link shortner)

src/Tribe/Plugins_API.php
barryhughes: Support only TEC is active and we're injecting PRO, which isn't active: it probably also isn't installed so I'm not going to be able to parse its readme.

I'd be be reluctant to hardcode from within the get_products, method too, because of the potential for things falling out of date depending on the customers update habits but that's something we could consider?

barryhughes: It's also just a default we can overwrite when we pull data in.

src/Tribe/Plugins_API.php
barryhughes: If that's a valid way to go, let's do that - sounds like the best of both worlds to me.

src/Tribe/Plugins_API.php
borkweb: Our TEC readme's tested-up-to entry often matches that of our premium plugins so if we wanted it to be more dynamic than an outright hardcode, we could use that. Your point on on things falling out of date is a good one. Perhaps we should version_compare() `$GLOBALS['wp_version']` and the readme date and use whichever is higher?

Though...maybe I'm just overcomplicating something that doesn't really matter in the end. An end user just wants to know if the plugins are compatible with the version of WP that they are running.

barryhughes: Actually that's not a bad thought ... that said, if we can confirm that Zach's suggestion is valid this may mostly be moot (since we'll then generally be able to override this default with the correct and current information)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:248](https://github.com/moderntribe/tribe-common/blob/83bf537bd6e2e1e6cf42541d138783561b3ec65e/src/Tribe/Plugins_API.php#L248)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:262](https://github.com/moderntribe/tribe-common/blob/83bf537bd6e2e1e6cf42541d138783561b3ec65e/src/Tribe/Plugins_API.php#L262)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:282](https://github.com/moderntribe/tribe-common/blob/83bf537bd6e2e1e6cf42541d138783561b3ec65e/src/Tribe/Plugins_API.php#L282)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:248](https://github.com/moderntribe/tribe-common/blob/a63e8f215c56db67b4ca9ef679c65b5e8a7c10f2/src/Tribe/Plugins_API.php#L248)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:262](https://github.com/moderntribe/tribe-common/blob/a63e8f215c56db67b4ca9ef679c65b5e8a7c10f2/src/Tribe/Plugins_API.php#L262)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:282](https://github.com/moderntribe/tribe-common/blob/a63e8f215c56db67b4ca9ef679c65b5e8a7c10f2/src/Tribe/Plugins_API.php#L282)

barryhughes:
Code that calls home is stripped out, what we have instead are not completely accurate placeholder values that look a whole lot better than what we had before.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:248](https://github.com/moderntribe/tribe-common/blob/44fda2161dfdc2bb04d0d7d6bd095f5c2db5ebcd/src/Tribe/Plugins_API.php#L248)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:262](https://github.com/moderntribe/tribe-common/blob/44fda2161dfdc2bb04d0d7d6bd095f5c2db5ebcd/src/Tribe/Plugins_API.php#L262)
:x: | Each line in an array declaration must end in a comma | [Plugins_API.php:282](https://github.com/moderntribe/tribe-common/blob/44fda2161dfdc2bb04d0d7d6bd095f5c2db5ebcd/src/Tribe/Plugins_API.php#L282)

PR #467, opened 2017-08-03
Added support to tribe_asset() for altering plugin dir or using WPMU …

elimn:
Ticket: https://central.tri.be/issues/82809

Per conversation with @bordoni we decided to remove the ability to pass a filepath to this method. It is doubtful anyone was doing this, none of our code is. It simplifies our logic quite a bit to just accept URLs, and it's not hard to convert a path to a URL before passing it.

Original bug report noted that if the plugins dir is not inside the wp-content dir, this method fails. I noted the same thing if the wpmu plugins dir isn't inside wp-content. This should address both.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [4.5.10] TBD =
 
-
+* Fix - Added support to tribe_asset() for altering plugin dir or using WPMU plugins. Thanks to @squirrelandnnuts for reporting this [82809]
 
 = [4.5.9] 2017-007-26 =
 
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -218,14 +218,18 @@ class Tribe__Assets {
 
 		if ( 0 === strpos( $url, WPMU_PLUGIN_URL ) ) {
 			// URL inside WPMU plugin dir.
-			$plugins_dir = wp_normalize_path( WPMU_PLUGIN_DIR );
-			$base_url   = WPMU_PLUGIN_URL;
+			$base_dir = wp_normalize_path( WPMU_PLUGIN_DIR );
+			$base_url = WPMU_PLUGIN_URL;
 		} elseif ( 0 === strpos( $url, WP_PLUGIN_URL ) ) {
 			// URL inside WP plugin dir.
-			$plugins_dir = wp_normalize_path( WP_PLUGIN_DIR );
-			$base_url   = WP_PLUGIN_URL;
+			$base_dir = wp_normalize_path( WP_PLUGIN_DIR );
+			$base_url = WP_PLUGIN_URL;
+		} elseif ( 0 === strpos( $url, WP_CONTENT_URL ) ) {
+			// URL inside WP content dir.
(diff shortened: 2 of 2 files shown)


src/Tribe/Assets.php
```diff
@@ -207,46 +207,52 @@ public function enqueue( $forcibly_enqueue = null ) {
...
+		if ( 0 === strpos( $url, WPMU_PLUGIN_URL ) ) {
+			// URL inside WPMU plugin dir.
+			$plugins_dir = wp_normalize_path( WPMU_PLUGIN_DIR );
+			$base_url   = WPMU_PLUGIN_URL;
+		} elseif ( 0 === strpos( $url, WP_PLUGIN_URL ) ) {
+			// URL inside WP plugin dir.
+			$plugins_dir = wp_normalize_path( WP_PLUGIN_DIR );
+			$base_url   = WP_PLUGIN_URL;
+		} else {
+			// Resource needs to be inside a plugins dir, if not we can't check if it exists.
```
borkweb: I know it is a super weird edge case, but some folks bake plugins inside of themes. I've had to do that when hosting a site on VIP.

readme.txt
```diff
@@ -4,7 +4,7 @@
 
 = [4.5.10] TBD =
 
-
+* Fix - Added support to tribe_asset() for altering plugin dir or using WPMU plugins. Thanks to @squirrelandnnuts for reporting this [82809]
```
barryhughes: Props on adding props!

elimn: And props to you for propping me on propping squirrelandnuts.

barryhughes: More props to you for giving me props for my acknowledgement of the props you gave to squirrelandnuts

src/Tribe/Assets.php
elimn: I am not an enormous fan of this method anyways. It does things a bit backwards.

It asks that you convert a filepath to a URL and pass it here, we then immediately try to reverse engineer the URL and determine the original filepath, then we check if the file exists. But accurately reversing the URL is not easy due to all the possible configurations of WordPress, in fact it's impossible to do outside of certain directories.

A much better method would be one that asks you for a filepath, and then returns the possibly minified URL. This is easy, WordPress provides an API for it. Best of all it would be trivial to rejigger the `Tribe__Assets` to make full use of this, as `Tribe__Assets` converts from a filepath to a URL immediately before calling this method. And `Tribe__Assets` is the only place we even use this method.

IMO what we should do is ticket a new method that does what I am describing. We don't even have to deprecate this one, we could just stop using it. It's static and harmless, just limited in functionality due to what it's trying to do. What do you think @borkweb ? In the mean time can I merge this PR, which has the same limitation as the previous code in this regard.

elimn:
Added WP_CONTENT as a fallback. Per Slack conversation we still have some edge cases we can't account for. URLs are not truly reversible, we are just doing our best here.

In the future we could consider adding a new method that only accepts filepaths. It would be able to leverage the WP API and work anywhere WP does. It would be a lot simpler than the above. Until then, this works.

PR #471, opened 2017-08-08
Fix image upload REST API issue

lucatume:
Tickets:
* https://central.tri.be/issues/76729
* https://central.tri.be/issues/76730

### DIFF
--- a/src/Tribe/Image/Uploader.php
+++ b/src/Tribe/Image/Uploader.php
@@ -100,6 +100,9 @@ class Tribe__Image__Uploader {
 		);
 
 		$id = wp_insert_attachment( $attachment, $upload['file'] );
+
+		require_once( ABSPATH . 'wp-admin/includes/image.php' );
+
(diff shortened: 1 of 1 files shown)


src/Tribe/Image/Uploader.php
```diff
@@ -100,6 +100,9 @@ protected function upload_file( $file_url ) {
 		);
 
 		$id = wp_insert_attachment( $attachment, $upload['file'] );
+
+		require_once( ABSPATH . 'wp-admin/includes/image.php' );
```
bordoni: Do a conditional; to see `wp_generate_attachment_metadata` doesn't already exists.

lucatume: OK, here is my reasoning for writing it like this (and I thought about your approach): if the function exists the file has been already included and require once, at an engine level, will make a faster check; the difference is between checking for a function among hundreds and then among files or just among less than 100 files in a C engine.
This was the reasoning, how does it sound? Plus: see [in the Codex](https://codex.wordpress.org/Function_Reference/media_handle_upload) how it's done (in the example)

PR #474, opened 2017-08-08
inArray() needs to be tested against -1, 0 is not false!

Camwyn:
Also, `.data()` was failing and giving us objects instead of text


barryhughes:
Branch `release/M17.16` is available if we want this to go out as part of 4.5.11.

PR #478, opened 2017-08-10
i18n: Simpler translation string

ramiy:
Related: 

https://github.com/moderntribe/the-events-calendar/pull/1481

See: https://central.tri.be/issues/88865


ramiy:
First we move the HTML tags out of the translation string.

And then we replace the functions name and the URL with `%s` placeholders.

barryhughes:
:ticket: [#88982](https://central.tri.be/issues/88982)

PR #482, opened 2017-08-12
Allows Data param for tribe_asset() localization to be a Callback

bordoni:
When I was moving us away from some of the Old Asset classes I found that having the Data set early is not always possible, so I've added a way for us to only run it very close to when the asset is actually loaded.


borkweb:
That being said, I don''t quite follow the `enqueue_group()` thing. I'd love to see an example of its usage.

bordoni:
So the idea of the grouping is to allow multiple scripts to be enqueued by a single function call.
Example; script and styles required for Weekview; can live in the `week-view` group and be enqueued all in one go

PR #485, opened 2017-08-14
Massaged tribe_resource_url until it worked with symlinks

elimn:
Ticket: https://central.tri.be/issues/74413

Altered this to use `plugin_basename()` instead of `basename()`. `plugin_basename()` has some lovely bits to account for symlinked plugin dirs, and it powers `plugins_url()` as well. By using it here we are giving `plugins_url()` a path it always understands, even if that path is actually symlinked to an atypical location.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [4.5.11] TBD =
 
-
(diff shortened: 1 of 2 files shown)


elimn:
This change is not needed the more I look into this. I think the users error just stems from them not deactivating and reactivating their plugins when the symlinks change. I am going to close this PR as it breaks when links are relative to the common dir, rather than the plugin dir.

PR #493, opened 2017-08-17
Add WP-CLI support to the logger

lucatume:
Ticket: https://central.tri.be/issues/81639

This PR moves the WP-CLI based logging from the Cron class found in EA client code to common to have logging on CLI everywhere on the `tribe( 'logger' )` object.


src/Tribe/Log.php
```diff
@@ -131,14 +133,71 @@ public function log_error( $entry, $src ) {
...
+				WP_CLI::error( $entry );
+				break;
+			case self::WARNING:
+				WP_CLI::warning( $entry );
+				break;
+			case self::SUCCESS:
+				WP_CLI::success( $entry );
+				break;
+			case self::DEBUG:
+				WP_CLI::debug( $entry, 'aggregator' );
```
borkweb: I'm guessing that this reference to `aggregator` is just copy pasta?

lucatume: yep

PR #494, opened 2017-08-17
Add More Validation methods + camelCase Utils

bordoni:
_Ref:_ [C#82243](http://central.tri.be/issues/82243)
---


src/resources/js/validation.js
```diff
@@ -37,12 +37,250 @@ tribe.validation = {};
...
+			return constraint == value;
+		},
+		isNotEqualTo: function( value, constraint, $field ) {
+			var type = $field.data( 'validationType' );
+			var $constraint = null;
+
+			// If it's not Numeric we treat it like a Selector
+			if ( ! _.isNumber( constraint ) ) {
+				$constraint = $( constraint );
+				constraint = $constraint.val();
```
Camwyn: I'd stick a check in here just in case we get a garbage selector - either bail or treat as 0? 
Probably before the `val()`
Same for the previous methods.

src/resources/js/validation.js
```diff
@@ -37,12 +37,250 @@ tribe.validation = {};
 	 * @type   {object}
 	 */
 	obj.conditions = {
-		required: function( val ) {
-			return '' == val;
+		isRequired: function( value ) {
+			return '' != value;
```
Camwyn: Value won't work for a checkbox or radio?

src/resources/js/validation.js
```diff
@@ -217,6 +497,29 @@ tribe.validation = {};
...
+		$listItem.data( 'validationField', $field );
+
+		// Add which notice item is related to this error field
+		$field.data( 'validationNoticeItem', $field );
+
+		$list.append( $listItem );
+	};
+
+	/**
 	 * Hooks to the submit and if invalid prevents submit to happen
```
Camwyn: "prevents submit from completing"? Same below (538) "prevent the submit from completing"

src/resources/js/validation.js
bordoni: For now, no. But this is only a MVP of the Module, when we have the need for it we can implement it.

PR #495, opened 2017-08-18
Ensure that minified asset detection is SSL-friendly.

boonebgorges:
I ran into a problem with The Events Calendar where JS assets were not loading on post-new.php?post_type=tribe_events. After a great deal of debugging, I traced it down to the problem that was previously quite thoroughly explained in this wordpress.org ticket by @earnjam : https://wordpress.org/support/topic/use-of-wp_content_url-breaks-plugin-on-admin-ssl-only-sites/ As he notes there, the problem arises on `FORCE_SSL_ADMIN`, but it's not only there: any time `get_option( 'home' )` is http:// but the current URL is https://, the problem arises. (SSL-optional sites, server-level rewrites/redirects, etc.)

Basically, when performing string comparisons on URL, you must ensure that your needle shares the scheme with the current haystack URL. There's a number of ways to do this. For greatest consistency and compatibility with mu-plugins, I chose to run everything through `set_url_scheme()` in this PR.

If for some reason you choose not to implement something like this, please provide a filter that I can use to work around the problem without forking the plugin :-)

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -216,18 +216,22 @@ class Tribe__Assets {
 	public static function maybe_get_min_file( $url ) {
 		$urls = array();
 
-		if ( 0 === strpos( $url, WPMU_PLUGIN_URL ) ) {
+		$wpmu_plugin_url = set_url_scheme( WPMU_PLUGIN_URL );
+		$wp_plugin_url = set_url_scheme( WP_PLUGIN_URL );
+		$wp_content_url = set_url_scheme( WPMU_CONTENT_URL );
+
+		if ( 0 === strpos( $url, $wpmu_plugin_url ) ) {
 			// URL inside WPMU plugin dir.
 			$base_dir = wp_normalize_path( WPMU_PLUGIN_DIR );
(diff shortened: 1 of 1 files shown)


barryhughes:
Thanks @boonebgorges - thanks for the PR!

What do you think of the change [made here](https://github.com/moderntribe/tribe-common/blob/release/M17.16/src/Tribe/Assets.php#L205-L268) (the release/M17.16 branch is going to be the basis of our next release), which were put in place for similar reasons?

As a related aside, a filter could be useful here even so.

boonebgorges:
Oh! https://github.com/moderntribe/tribe-common/commit/3d7aedf8c198b5e623225a996e9078954c11ffd2 is the same as this PR. Sorry for not seeing this - I didn't know how your branch system worked.

barryhughes:
No problem - we'll have a think about ways to make that a little clearer (we recently re-structured the way we use branches).

As a heads-up, though I can't guarantee anything, we _should_ see this change ship next week.

Thanks again!

boonebgorges:
Many thanks for the quick follow-up!

PR #502, opened 2017-08-28
pass the requested URL, not the mock, to the callback

lucatume:
Ticket (context): https://central.tri.be/issues/84695

Small fix to hte HTTP API class


barryhughes:
:bulb: this was merged into `release/M17.16` which already shipped (as 4.5.11, last week) ... flagging because we probably want to resubmit and target `release/M17.17` (or cherry-pick it across directly).

PR #510, opened 2017-08-30
Spotfix - add filter to customize list of states

cliffordp:
Tweak - Added a filter to customize the list of states in the USA that are available to drop-downs when creating or editing venues.

Use case documented at https://theeventscalendar.com/support/forums/topic/restrict-location-options/

### DIFF
--- a/src/Tribe/View_Helpers.php
+++ b/src/Tribe/View_Helpers.php
@@ -318,7 +318,7 @@ if ( ! class_exists( 'Tribe__View_Helpers' ) ) {
 		 * @return array The states array.
 		 */
 		public static function loadStates() {
-			return array(
+			$states = array(
 				'AL' => esc_html__( 'Alabama', 'tribe-common' ),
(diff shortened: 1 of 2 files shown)


src/Tribe/View_Helpers.php
```diff
@@ -371,6 +371,8 @@ public static function loadStates() {
 				'WI' => esc_html__( 'Wisconsin', 'tribe-common' ),
 				'WY' => esc_html__( 'Wyoming', 'tribe-common' ),
 			);
+
+			return apply_filters( 'tribe_get_state_options', $states );
```
elimn: Could you add a docblock for this?

https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

cliffordp:
@elimn, please re-review. tyvm!

elimn:
Thanks! Typically we would do `@since TBD` and then as part of our release we would assign the version. But, at this point I think it's safe to assume the version number.

PR #514, opened 2017-09-08
Fix some issues in the Field API

lucatume:
Ticket (context): https://central.tri.be/issues/79975

Some fixes to the Field API and the addition of an `id` attribute to the `dropdown` `select` element


lucatume:
I'm the one that found it out while working on 79975; as per the changelog: it seems so internal that I would not even know how to report it.

borkweb:
Ohhhhh - I see. How about this for a changelog entry:

```
* Tweak - Added an id attribute to dropdowns generated by the Fields API [spotfix]
```

lucatume:
OK, changelog entry added.

PR #515, opened 2017-09-11
Feature/78023 remove metropolitan france

niconerd:
https://central.tri.be/issues/78023


niconerd:
Props added @borkweb

niconerd:
Also, do you think this need to be exposed in The Events Calendar changelog as well ?

borkweb:
Exposing in the TEC changelog might not be a bad idea!

PR #523, opened 2017-09-15
Safer filter deprecation | spotfix

barryhughes:
The `deprecated_filter_message()` method needs to pass through the filtered value, else it will potentially break any code that still uses the deprecated hook (unless it so happens to expect or tolerate `null`).

[Example of breakage.](https://tribe.slack.com/archives/C3XGB48QK/p1505512878000238)


src/Tribe/Deprecation.php
```diff
@@ -117,8 +117,14 @@ public function deprecated_action_message() {
...
 	 * Triggers a deprecation notice if there is any callback hooked on a deprecated filter.
+	 *
+	 * @since TBD the filtered value is passed through unchanged
+	 *
+	 * @param mixed $value
+	 *
+	 * @return mixed
 	 */
-	public function deprecated_filter_message() {
+	public function deprecated_filter_message( $value ) {
```
bordoni: Add a Default, to prevent problems when attaching with `0` as the params

barryhughes: good call, done

PR #526, opened 2017-09-19
Update inputs loop for REST API Tester to support checkboxes (checked vs not checked)

sc0ttkclark:


lucatume:
How would the UI come to have checkboxes? Does this PR relates to a PR I missed?

sc0ttkclark:
Yes that was PR #501 that got merged in for boolean field types. If we see a swagger spec with boolean, we use a checkbox with value=1, but when we introduced that it would submit all checkboxes no matter what. This PR just limits to those with checked prop true. We also now support number inputs as a result of that PR. It works awesome!

lucatume:
The only issue I have with that is that we have just killed the plugin capacity to send a "wrong" value for a boolean field: this is a legit testing scenario.

sc0ttkclark:
If that's the case, I can back the checkboxes/number inputs out.

lucatume:
I say we merge this and, maybe, revert your stuff later.
Better yet: provide a default mode, with UI helps like the ones you put in place, and a test/dev mode to be able to send anything on any field.

sc0ttkclark:
Link next to the checkbox / number input "Custom input" which then converts the input to a normal input type="text" -- that wouldn't be too hard

lucatume:
that's a brilliant idea; noted and good for another PR

sc0ttkclark:
Needs official review approval first

PR #531, opened 2017-09-22
Ticket Editor Validation

bordoni:
_Ref:_ [C#80632](http://central.tri.be/issues/80632)
---


src/Tribe/Main.php
```diff
@@ -249,8 +253,19 @@ public function load_assets() {
...
+	 }
+
+	/**
+	 * Load All localization data create by `asset.data`
+	 *
+	 * @since  TBD
+	 *
+	 * @return void
+	 */
+	public function load_localize_data() {
```
Camwyn: Nitpicky and not a blocker, but should this be `load_localized_data`?

bordoni: That was how it was on RER, I would rather not change to avoid merge conflicts if possible

Camwyn: That makes sense 👍

PR #532, opened 2017-09-22
Fetch the defaultCurrencySymbol for any case where symbol is falsey

borkweb:
We want to default to the `defaultCurrencySymbol` when formatting currency _any time_ the currency symbol is `null`/`false`/`0` regardless of the state of `$post_id`.

Adjusted tests to account for the presence/absence of `$post_id`.

See: https://central.tri.be/issues/89218


barryhughes:
:cactus:

PR #540, opened 2017-09-28
Fix/68472 country list performance

elimn:
Ticket: https://central.tri.be/issues/68472

Originally I did not realize that Filter Bar calls the `constructCountries()` method 10 times in a row, seemingly out of necessity. This results in a ~.4 second delay on my system. If we cache it the first time in memory we can drastically help page loads when subsequent calls are needed. All of this work does that.


elimn:
@lucatume I made some modifications to the `Tribe__Cache` class that I figured you would like to review. I wanted a method that get the value from the cache. If the value was not in the cache, then it would retrieve data from a callback and store that in the cache. Originally I had thought about adding a second method in addition to `get()` that does this, but that second method would require all the same arguments as `get()` plus a few extra. In the end it made more sent to me to add some optional args to `get()`.

src/Tribe/Cache.php
```diff
@@ -58,15 +57,27 @@ public function set_transient( $id, $value, $expiration = 0, $expiration_trigger
...
+	 * @param string       $id                 The key for the cached value.
+	 * @param string       $expiration_trigger Optional. Hook to trigger cache invalidation.
+	 * @param string|array $callback_if_empty  Optional. Sets the cache value to the return of this callback if no value was already set.
+	 * @param mixed        $callback_args      Optional. Args passed to callback.
+	 * @param int          $expiration         Optional. Value of expiration if we have to set the cache value.
 	 *
 	 * @return mixed
 	 */
-	public function get( $id, $expiration_trigger = '' ) {
+	public function get( $id, $expiration_trigger = '', $callback_if_empty = null, $callback_args = array(), $expiration = 0 ) {
```
lucatume: Rather than `callback_if_empty` here I would use `default`; I'd add a check to see if `$default` `is_callable()` and either use that or use the return value of the callback. I would also flip the `expiration` and `callback_args` parameters to have a signature like this:
```php
public function get( $id, $expiration_trigger = '', $default = null,  $expiration = 0, $args = array())
```
The reason being that we might more frequently use the `expiration` parameter than the `args` (I'd rename those to decouple the parameter from the concept of a callback).

src/Tribe/Languages/Locations.php
```diff
@@ -0,0 +1,371 @@
...
+	 *
+	 * Adds array to object cache to speed up subsequent retrievals.
+	 *
+	 * @return array {
+	 *      List of countries
+	 *
+	 *      @type string $country_code Country name.
+	 * }
+	 */
+	public function countries() {
```
lucatume: Rename this to `get_countries`

src/Tribe/Languages/Locations.php
```diff
@@ -0,0 +1,371 @@
...
+	 *
+	 * Adds array to object cache to speed up subsequent retrievals.
+	 *
+	 * @return array {
+	 *      List of countries
+	 *
+	 *      @type string $country_code Country name.
+	 * }
+	 */
+	public function us_states() {
```
lucatume: rename to `get_us_states`

src/Tribe/Cache.php
lucatume: Bonus points if you add a test for this to
* test that the `default` parameter will take and use scalar values
* test that the `default` parameter will take and use callbacks (called only once)
* test that the `args` array, if present, is passed to the `default` callback
* test that if the `default` callback throws an exception we handle the exception and do not fill the cache

elimn:
All of @lucatume's feedback has been address, except the ["bonus points"](https://github.com/moderntribe/tribe-common/pull/540#discussion_r141896470) tests he was thinking could be added. I would like to add those, but I will do so in a followup PR that might come in after the QA deadline. If anyone else can rereview this I would appreciate it, Luca is now out for the weekend..

PR #541, opened 2017-09-28
Use URL functions instead of undocumented constants

earnjam:
`WP_CONTENT_URL` and `WP_PLUGINS_URL` can cause issues when sites use domain mapping or have http front-end and https admin. These functions prevent those issues by running them through the proper filters and scheme checks.

See: https://wordpress.org/support/topic/support-for-domain-mapping/

You could probably also support mu-plugins in the same way, but there isn't a function that wraps `WPMU_PLUGIN_URL` neatly like there is for the other two constants.

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -216,8 +216,8 @@ class Tribe__Assets {
 	public static function maybe_get_min_file( $url ) {
 		$urls = array();
 		$wpmu_plugin_url = set_url_scheme( WPMU_PLUGIN_URL );
-		$wp_plugin_url = set_url_scheme( WP_PLUGIN_URL );
-		$wp_content_url = set_url_scheme( WP_CONTENT_URL );
+		$wp_plugin_url = set_url_scheme( plugins_url() );
+		$wp_content_url = set_url_scheme( content_url() );
 
 		if ( 0 === strpos( $url, $wpmu_plugin_url ) ) {
 			// URL inside WPMU plugin dir.


barryhughes:
Thanks, @earnjam!

The release we're currently working on is in the `release/M17.20` branch so I've updated the base branch accordingly. However, your work conflicts with some other changes made in that same function. 

Can you review and if you still feel your change is needed update your own branch to avoid conflicts? Thanks and sorry for any extra work here :-)

earnjam:
Just glancing over it, I _think_ it should be fine. The change that @boonebgorges committed is a fallback if the `WP_PLUGIN_URL` constant doesn't match the url (such as a mixed http/https site or a multisite network using a domain mapping plugin). Seems like it would be more efficient to just get the correct URL to begin with using the `plugins_url()` or `content_url()` functions, rather than comparing to the constant first, and then later to the result of the function you've already executed. 

I guess this would prevent any breakage from someone incorrectly using the `plugins_url` filter, but I'm pretty sure that would break a lot more than this plugin.

barryhughes:
OK - well we're still open to making a further change here, but due to the conflict with those other changes submitted by Boone we'd need you to resolve that first (whether by updating your own branch with the latest changes from `release/M17.20` or manually).

barryhughes:
4.6.2 has shipped, changing base branch so we can hopefully progress in our next release.

barryhughes:
:bulb: this was closed automatically once we shut down release/M17.21 (our 4.6.3 release). Please do feel free to re-open/create a new pull request targeting release/M17.22 if you still see a need for this work :+1:

PR #549, opened 2017-10-13
#82286 Add Function for Checking if WPML is Active

ggwicz:
**Ticket:** [**#82286**](http://central.tri.be/issues/82286)

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -336,10 +336,6 @@ if ( ! function_exists( 'tribe_is_wpml_active' ) ) {
 	 * @return boolean
 	 */
 	function tribe_is_wpml_active() {
-		if ( class_exists( 'SitePress' ) && defined( 'ICL_PLUGIN_PATH' ) ) {
-			return true;
-		}
-
-		return false;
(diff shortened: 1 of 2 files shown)


src/functions/utils.php
```diff
@@ -325,4 +325,21 @@ function tribe_retrieve_object_by_hook( $class_name, $hook, $priority ) {
...
+if ( ! function_exists( 'tribe_is_wpml_active' ) ) {
+	/**
+	 * A unified way of checking if WPML is activated.
+	 *
+	 * @since TBD
+	 *
+	 * @return boolean
+	 */
+	function tribe_is_wpml_active() {
+		if ( class_exists( 'SitePress' ) && defined( 'ICL_PLUGIN_PATH' ) ) {
```
zbtirrell: you could just directly return this:
```
return ( class_exists( 'SitePress' ) && defined( 'ICL_PLUGIN_PATH' ) );
```

elimn: This function could be shortened to:
`return class_exists( 'SitePress' ) && defined( 'ICL_PLUGIN_PATH' );`

ggwicz: true! done and ready for another review

PR #556, opened 2017-10-24
Prevent Fail on Validation for Empty Datepicker format

bordoni:
_Ref:_ [C#91563](http://central.tri.be/issues/91563)
---


zbtirrell:
seems fair 👍

PR #558, opened 2017-10-25
Fix/72127 select2 ie issue

elimn:
Ticket: https://central.tri.be/issues/72924

I purposefully broke this up into two commits to make it easier to review. Initial commit https://github.com/moderntribe/tribe-common/commit/8f2a4954921dd57449685f2dbe45a7d00b595704 is for removing the submodule, and adding the files back as a modifiable vendor lib. This is just a temporary measure until we upgrade to selectWoo, then we might be able to switch back to a submodule, though it will have a different name.

The second commit https://github.com/moderntribe/tribe-common/commit/6007fa6d071f8a069248e2210443ab8746d0de3f is where I modified the vendor lib. The fix is pretty small.


elimn:
I forgot about adding this to the package whitelist. The last commit https://github.com/moderntribe/tribe-common/pull/558/commits/ae4a1636b92b3108c0ede6629d8a5704c5c98acb fixes that, it's the only change since @borkweb approved it.

PR #561, opened 2017-10-27
Groundwork for consistent filtering of JSON LD objects | #89801

barryhughes:
Retains the same filter, but moves it to its own method for easier reuse in subclasses (since many or all subclasses override the `get_data()` method, we need an easy way to let each of those overrides apply the same filter for consistency).

@sc0ttkclark - flagging in case you have any concerns (pretty sure it's safe and won't pose any problems, but it does modify work you completed in [PR 527](https://github.com/moderntribe/tribe-common/pull/527) - please see also [TEC PR 1647](https://github.com/moderntribe/the-events-calendar/pull/1647) for context).

:ticket: [#89801](https://central.tri.be/issues/89801)


sc0ttkclark:
Appreciate you keeping me in the loop on this, wasn't expecting to be included, but I'm glad to see this tweak!

PR #566, opened 2017-11-03
#78621 Better Accommodate 24-hr Time Format in Options

ggwicz:
**Ticket:** [**#78621**](http://central.tri.be/issues/78621)

**Related PR:** https://github.com/moderntribe/the-events-calendar/pull/1658

---

### Code Review Note

- I created this function because I originally thought that checking for the 24-hour time directly would be a good solution for #78621.
- But then I noticed that the site's `time_format` option wasn't being respected for the "End of Day Cutoff" option _period_, not **just** when it was a 24-hour time format, so I switched approaches and used `date_i18n()` to get the right format regardless of what it was. [(_See the TEC PR for more info._)](https://github.com/moderntribe/the-events-calendar/pull/1658)
- **I still think this function could be useful though**, so submitted this PR here anyways; **is that cool, or should I pull this because it's not yet being used?**

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.6.4] TBD =
+
+* New - Added new `tribe_is_site_using_24_hour_time()` function to easily check if the site is using a 24-hour time format [78621]
+
 = [4.6.3] 2017-11-02 =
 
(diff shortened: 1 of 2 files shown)


ggwicz:
Fixed some `readme.txt` conflicts; mind giving this another review, @barryhughes?

PR #572, opened 2017-11-10
check if meta key is string before logic

lucatume:
Tciket: https://central.tri.be/issues/91434

Handle the case where the meta key is not a string


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M17.22*.

PR #573, opened 2017-11-10
Resolved the problem on Select2 created on the F17.8 release

bordoni:
On Ticket Editor Refresh (F17.8), I've introduced a `tribe-select2` to `tribe-dependency` which made some Third Party Plugins like ACF break due to the usage of Select2 v4. 

Modified the code on `dependency.js` here to make sure it doesn't throw any warnings for not having Select2, and use it correctly when available.


barryhughes:
:ticket: [♯93017](https://central.tri.be/issues/93017)

PR #574, opened 2017-11-10
#61880 Allow Non-Latin URL Slugs

ggwicz:
**Ticket:** [**#61880**](http://central.tri.be/issues/61880)

**Related PR:** https://github.com/moderntribe/the-events-calendar/pull/1680

### DIFF
--- a/src/Tribe/Validate.php
+++ b/src/Tribe/Validate.php
@@ -254,12 +254,9 @@ if ( ! class_exists( 'Tribe__Validate' ) ) {
 		public function slug() {
 			$maybe_valid_value = esc_url_raw( $this->value );
 
-			// While esc_url_raw is handy and does the work of validating for us, it returns the
-			// checked string with a prepended URL protocol; let's remove that.
-			$maybe_valid_value = wp_kses_bad_protocol_once( $maybe_valid_value, array() );
-			$maybe_valid_value = str_replace( '//', '', $maybe_valid_value );
-
-			if ( $maybe_valid_value == $this->value ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Validate.php
```diff
@@ -252,7 +252,14 @@ public function int() {
...
 		public function slug() {
-			if ( preg_match( '/^[a-zA-Z0-9-_]+$/', $this->value ) ) {
+			$maybe_valid_value = esc_url_raw( $this->value );
+
+			// While esc_url_raw is handy and does the work of validating for us, it returns the
+			// checked string with a prepended URL protocol; let's remove that.
+			$maybe_valid_value = wp_kses_bad_protocol_once( $maybe_valid_value, array() );
+			$maybe_valid_value = str_replace( '//', '', $maybe_valid_value );
+
+			if ( $maybe_valid_value == $this->value ) {
```
Camwyn: Just a thought: can we skip the juggling by just doing a `stripos`?

ggwicz: @Camwyn

we _could_, but only by explicitly naming the protocols 

my thinking is that, since this validation method could in theory be used anywhere, in lots of diff contexts, we want to make sure there's absolutely no protocol there, and i wanted to cover anything beyond `http://` and `https://` so we are *certainly* dealing with a slug only

Camwyn: Gotcha.

ggwicz: @Camwyn and i are discussing this choice some more over Slack DM. may change approach here.

Camwyn:
That looks way cleaner and should do the trick!

PR #580, opened 2017-11-29
Better enforced Datepicker Styles to overwrite for our Datepicker

bordoni:
_Ref:_ [C#94161](http://central.tri.be/issues/94161)
---


src/resources/postcss/datepicker.pcss
```diff
@@ -274,6 +285,8 @@
 	color: #23282d;
 }
 
+
+
```
Camwyn: Three empty lines? :trollface:

PR #584, opened 2017-12-03
fix PHP Warning for `count()` as $search can be null

afragen:
I'm seeing a PHP warning for `count()` on line 204 because $seach can be null.

screenshot_01


afragen:
Anytime Barry.

PR #585, opened 2017-12-05
#71121 - move method to common for getting timezone string

jesseeproductions:
_Ref:_ [C#71121](https://central.tri.be/issues/71121)

https://github.com/moderntribe/events-pro/pull/742

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -1167,6 +1167,34 @@ if ( ! class_exists( 'Tribe__Date_Utils' ) ) {
 			// Why so simple? Let's handle other cases as those come up. We have tests in place!
 			return str_replace( '\\\\', '\\', $date_format );
 		}
+
+		/**
+		 * Gets the timezone string associated with an event.
+		 *
+		 * Will fall back to WP timezone string and to system timezone string in this order.
+		 *
+		 * @since TBD
+		 *
+		 * @param $event_id
+		 *
+		 * @return mixed|string|void
+		 */
+		public static function get_event_timezone_string( $event_id ) {
+			$event_timezone_string = get_post_meta( $event_id, '_EventTimezone', true );
+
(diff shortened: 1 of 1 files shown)


src/Tribe/Date_Utils.php
```diff
@@ -1167,6 +1167,34 @@ public static function unescape_date_format( $date_format ) {
...
+		 *
+		 * @return mixed|string|void
+		 */
+		public static function get_event_timezone_string( $event_id ) {
+			$event_timezone_string = get_post_meta( $event_id, '_EventTimezone', true );
+
+			if ( empty( $event_timezone_string ) ) {
+				$event_timezone_string = get_option( 'timezone_string', false );
+				if ( empty( $event_timezone_string ) ) {
+					$event_timezone_string = date_default_timezone_get();
```
ggwicz: non-blocker for sure, just commenting: would love to see this part reduced to one line:

`return date_default_timezone_get();`

src/Tribe/Date_Utils.php
```diff
@@ -1167,6 +1167,34 @@ public static function unescape_date_format( $date_format ) {
...
+		 *
+		 * Will fall back to WP timezone string and to system timezone string in this order.
+		 *
+		 * @since TBD
+		 *
+		 * @param $event_id
+		 *
+		 * @return mixed|string|void
+		 */
+		public static function get_event_timezone_string( $event_id ) {
```
barryhughes: Got a couple of thoughts on this:

1) Here in Tribe Common we should avoid event-specific stuff - that ought to go in TEC
2) What's the advantage of this method over `Tribe__Events__Timezones::get_event_timezone_string()`? If there's a difference, can it be reconciled and can we just alter the existing method of the same name?

jesseeproductions: closing this pr to use an existing method.

PR #589, opened 2017-12-08
Event Cost no longer forced to numeric

cliffordp:
https://central.tri.be/issues/95403

Note that '17,000,100.53' is non-numeric (comma-separated thousands) and therefore will be passed as the raw Event Cost string instead of displaying with a currency symbol in front of it.

I believe this is acceptable.

### DIFF
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -321,8 +321,14 @@ class Tribe__Cost_Utils {
 
 		foreach ( $costs as &$cost ) {
 			// Get the required parts
+			$cost_wo_separators = $cost;
+
+			foreach ( $this->get_separators() as $separator ) {
+				$cost_wo_separators = str_replace( $separator, '', $cost_wo_separators );
+			}
+
 			if (
-				is_numeric( $cost )
+				is_numeric( $cost_wo_separators )
 				&& preg_match_all( '/' . $price_regex . '/', $cost, $matches )
 			) {
 				$cost = reset( $matches );
--- a/src/Tribe/Cost_Utils.php
+++ b/src/Tribe/Cost_Utils.php
@@ -321,7 +321,10 @@ class Tribe__Cost_Utils {
 
 		foreach ( $costs as &$cost ) {
 			// Get the required parts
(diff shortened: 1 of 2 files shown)


src/Tribe/Cost_Utils.php
```diff
@@ -321,7 +321,10 @@ public function parse_cost_range( $costs, $max_decimals = null, $sort = true ) {
 
 		foreach ( $costs as &$cost ) {
 			// Get the required parts
-			if ( preg_match_all( '/' . $price_regex . '/', $cost, $matches ) ) {
+			if (
+				is_numeric( $cost )
+				&& preg_match_all( '/' . $price_regex . '/', $cost, $matches )
```
barryhughes: I'm not completely clear on why test here to see if `$cost` is numeric ... the `$price_regex` is designed to accommodate and work with strings such as `1,234,567.89` which would fail that test.

barryhughes: Let me re-word ... if we only accept strings and numbers that pass the `is_numeric()` test, why parse with the regex?

src/Tribe/Cost_Utils.php
```diff
@@ -321,8 +321,14 @@ public function parse_cost_range( $costs, $max_decimals = null, $sort = true ) {
 
 		foreach ( $costs as &$cost ) {
 			// Get the required parts
+			$cost_wo_separators = $cost;
+
+			foreach ( $this->get_separators() as $separator ) {
```
barryhughes: Non-essential, but we could simplify this a little and dispose of the `foreach()` loop by doing something like:

```php
$cost_wo_separators = str_replace( $this->get_separators(), '', $cost_wo_separators );
```

barryhughes:
Not sure we want to proceed with this just yet - it breaks existing functionality in some cases (expected currency formatting won't be applied to cost ranges) and there are possibly other things to think on, too.

Placing on hold and I've left a question for strategy input [over here](https://central.tri.be/issues/95403#note-4).

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M17.25*.

PR #590, opened 2017-12-11
Fix the license validation colors

lucatume:
Ticket (context): https://central.tri.be/issues/91890

Without this fix the AJAX validation of keys would not mutate the message color, no matter the result of the validation; the correct color would be applied after a page refresh: https://cl.ly/2f1w3k0v0d07

This fixes it: https://cl.ly/1s0E1v0o1N1i


barryhughes:
Probably deserves a changelog entry

PR #592, opened 2017-12-15
Fixes from test restore work

lucatume:
Ticket: n/a

Related to: https://github.com/moderntribe/the-events-calendar/pull/1725

This PR:

* makes the ``tribe_beginning_of_day` and `tribe_end_of_day` accept and handle timestamps and not just string dates


barryhughes:
It looks good, but I have basically the same question [that I asked here](https://github.com/moderntribe/the-events-calendar/pull/1725#issuecomment-352027466) (ie, it's not ticketed and that means it won't go in front of QA - is that safe, should we fix that and ticket this work - and is it essential we get this in M17.25 since we're very close to release at this point).

lucatume:
Reopening this PR with `master` base

PR #595, opened 2018-01-04
Test to see if var is array before using count() | #95527

barryhughes:
As of PHP 7.2, we need to be more careful about our use of `count()` if it's possible the var is not going to be an array or countable object.

:ticket: [#95527](https://central.tri.be/issues/95527)

### DIFF
--- a/src/Tribe/Customizer/Section.php
+++ b/src/Tribe/Customizer/Section.php
@@ -206,7 +206,7 @@ abstract class Tribe__Customizer__Section {
 			return $settings;
 		}
 
-		if ( count( $search ) === 1 ) {
+		if ( is_array( $search ) && count( $search ) === 1 ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Customizer/Section.php
```diff
@@ -206,7 +206,7 @@ public function filter_settings( $settings, $search ) {
 			return $settings;
 		}
 
-		if ( count( $search ) === 1 ) {
+		if ( is_array( $search ) && count( $search ) === 1 ) {
```
mitogh: I think this one should be fine but we can always cast the variable as well.

barryhughes: Hmm, but if we cast `$search` to an array the special handling in the else block won't be applied.

PR #596, opened 2018-01-05
Fix the nonce issues

lucatume:
Ticket: https://central.tri.be/issues/89404 and https://central.tri.be/issues/95916#note-10

This PR fixes the nonce validation issue that was first only present when using CE and was then always present on WP 4.9 and above rendering the test plugin useless.

Changes:
* fix the nonce issues
* in so doing semplify the logic for the nonce spoofing
* remove the Webpack dependency manually including, the WP way, the `renderjson` file.

Work for ticket https://central.tri.be/issues/88265 will come in another PR (generalize the plugin to allow using with all REST APIs)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M18.01*.

PR #597, opened 2018-01-08
fix validation callback behaviour

lucatume:
Ticket: n/a

Make sure our code does not return any value from `__construct` and that `validation_callback` is correctly used and supported.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Tribe Commerce*, but this was pull requested against *release/M18.01*.

barryhughes:
Looks good - but we should also remove or correct the `@return` tag in the constructor docblock (my vote being we just remove it).

lucatume:
RIght, thanks. I've removed it.

PR #598, opened 2018-01-09
Remove missing meta field when post is removed

mitogh:
Makes sure to clean up the tracker meta field as the priority the meta
is not deleted by WordPress however after the action delete_post is
fired we can do another manual cleanup.

See https://central.tri.be/issues/96136

### DIFF
--- a/src/Tribe/Tracker.php
+++ b/src/Tribe/Tracker.php
@@ -436,9 +436,10 @@ class Tribe__Tracker {
 	 *
 	 * @since TBD
 	 *
-	 * @param int Post ID
+	 * @param int  Post ID
+	 * @return bool
 	 */
 	public function cleanup_meta_fields( $post_id ) {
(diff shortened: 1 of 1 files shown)


src/Tribe/Tracker.php
```diff
@@ -426,4 +429,16 @@ public function set_tracked_taxonomies( array $tracked_taxonomies ) {
...
+	/**
+	 * Make sure to remove the changed field if the event is deleted to ensure there are no left meta fields when
+	 * the event is deleted.
+	 *
+	 * @since TBD
+	 *
+	 * @param int Post ID
+	 */
+	public function cleanup_meta_fields( $post_id ) {
+		delete_post_meta( (int) $post_id, self::$field_key );
```
bordoni: Can we return the result of the `delete_post_meta()` and add the return to the Doc Block?

mitogh: Changes has been updated.

PR #599, opened 2018-01-17
Make sure to merge the settings with the defaults

mitogh:
This will make sure if any section has settings merge those with the default values of each section in the customizer.

See https://central.tri.be/issues/96821

PR #601, opened 2018-01-17
Report label has been updated based on feedback

mitogh:
Central - https://central.tri.be/issues/68687


mitogh:
**Note** changelog entry is going to be added after #599 is merged.

PR #603, opened 2018-01-18
#94725 Validate PHP Date Formats

ggwicz:
**Ticket:** [**#94725**](http://central.tri.be/issues/94725)

There are no "props" for this ticket.

### DIFF
--- a/src/Tribe/Validate.php
+++ b/src/Tribe/Validate.php
@@ -182,7 +182,7 @@ if ( ! class_exists( 'Tribe__Validate' ) ) {
 			if ( empty( $this->value) ) {
 				$this->result->valid = false;
 				$this->result->error = sprintf( esc_html__( '%s must not be empty', 'tribe-common' ), $this->label );
-			} else 
+			} else {
 				$this->result->valid = true;
 			}
 		}
--- a/src/Tribe/Validate.php
+++ b/src/Tribe/Validate.php
@@ -172,6 +172,26 @@ if ( ! class_exists( 'Tribe__Validate' ) ) {
 		}
 
 		/**
+		 * Validates a field as a string acceptable for use as the $format parameter in PHP's date()
+		 * function: a-z, A-Z, and a variety of punctuation (hyphen, em dash, en dash, colon, semicolon
+		 * comma, period, vertical bar, left and right brackets, @ symbol, ampersand, /, ', and ").
+		 *
+		 * @since TBD
+		 *
+		 * @return stdClass validation result object
+		 */
+		public function php_date_format_string() {
+			$this->value = trim( $this->value );
+
+			if ( preg_match( '/^[a-zA-Z ,.:;|\'"<>_—–@&\/\-]+$/', $this->value ) ) {
+				$this->result->valid = true;
+			} else {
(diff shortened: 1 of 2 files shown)


src/Tribe/Validate.php
```diff
@@ -171,6 +171,26 @@ public function alpha_numeric_with_dashes_and_underscores() {
...
+		 * comma, period, vertical bar, left and right brackets, @ symbol, ampersand, /, ', and ").
+		 *
+		 * @since TBD
+		 *
+		 * @return stdClass validation result object
+		 */
+		public function php_date_format_string() {
+			$this->value = trim( $this->value );
+
+			if ( preg_match( '/^[a-zA-Z ,.:;|\'"<>_—–@&\/\-]+$/', $this->value ) ) {
```
bordoni: Doesn't it allow anything if escaped?

ggwicz: Can you give an example of what you mean?

My goal is to allow letters and certain punctuation—in my testing this regex worked well, but I'm not super proficient with regex so may be missing something.

barryhughes: The regex covers a lot of possibilities yet is also a little restrictive ... it won't let through digits or international chars like `ē` or other things such as `»` and I can't think of a good reason to block them out of hand.

Looking at the Central ticket for this work, the problem was that we were accepting _blank_ strings ... is it possible there's a simpler and less restrictive solution?

bordoni: Chaining it to only accept string that contain the PHP date modifiers would be nice, instead of check for what it can't be, check for at least one of the modifiers (without a leading `\`).

ggwicz: Sorry guys, I'm a little confused to be honest. 

The problem is that blank values were allowed, yes—so **we _could_ just make it so that blank values are not accepted.**

> The regex covers a lot of possibilities yet is also a little restrictive ... it won't let through digits or international chars like ē or other things such as » and I can't think of a good reason to block them out of hand.

When it comes to letters and international chars, I only allowed a-z and A-Z because they're the only ones allowed for [the `$format` parameter for PHP's `date()` function](https://secure.php.net/manual/en/function.date.php) — as for the symbols/punctuation, `»` would be good to add, that's a good call.

However, maybe for now just presenting a blank value is the best?

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space before the last parenthesis | [Validate.php:182](https://github.com/moderntribe/tribe-common/blob/a221444944e44705e9e19b2287dd6b7a5b8b6311/src/Tribe/Validate.php#L182)
:x: | Whitespace found at end of line | [Validate.php:185](https://github.com/moderntribe/tribe-common/blob/a221444944e44705e9e19b2287dd6b7a5b8b6311/src/Tribe/Validate.php#L185)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space before the last parenthesis | [Validate.php:182](https://github.com/moderntribe/tribe-common/blob/60d3fb735d033cf07173eb6cec7fc5ef12bd65ee/src/Tribe/Validate.php#L182)

PR #608, opened 2018-02-07
Add Regular Expressions for Validation.js

bordoni:
_Ref:_ [C#81831](http://central.tri.be/issues/81831)
---
Also added Tooltipster for later on down the line add Inline errors for Validation JS


src/resources/js/validation.js
```diff
@@ -565,6 +600,8 @@ tribe.validation = {};
 			var $field = $( field );
 			var message = $field.data( 'validationError' );
 
+			console.log( message );
```
mitogh: Is this one un purpose ?

mitogh: And following `console.log`

PR #609, opened 2018-02-07
Update function declarations to use no conflict version of underscore

mitogh:
Ticket: https://central.tri.be/issues/92205

Update references to the new underscore variable and fallback to the global if not defined.


mitogh:
Waiting for https://github.com/moderntribe/the-events-calendar/pull/1797 if we need to include the code from there in this one.

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,45 @@
+<?php
+/**
+ * Class used to hook into the assets being loaded
+ *
+ * @since TBD
+ */
+class Tribe__Load_Assets {
```
ggwicz: i know @bordoni recommended this class name, but i personally find `load_assets` a bit confusing, since _all_ of our `_asset` classes technically "load assets"

to me something more specific like `modify_assets` or `filter_assets` or even just `assets_loaded` would make more sense, to help clarify that this class and its methods — thoughts?

barryhughes: 🤔 I think that's a great point. We should probably ask how we see this class developing in the future: if it's only going to be concerned with underscore compatibility, perhaps it ought to follow the same model and naming as our integration classes (in TEC).

barryhughes: ...if we legit see this as a place to handle all kinds of script conflicts, on the other hand, maybe the name should reflect that?

bordoni: It's mostly a place where we will load all the Assets for Common maybe. Any other suggestions?

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	public function prevent_underscore_conflict( $tag, $handle ) {
+		if ( is_admin() || ! defined( 'TRIBE_EVENTS_FILE' ) ) {
+			return $tag;
+		}
+
+		if ( 'underscore' === $handle ) {
+			$path = plugin_dir_path( TRIBE_EVENTS_FILE ) . 'common/src/resources/js';
+			$dir = plugin_dir_url( TRIBE_EVENTS_FILE ) . 'common/src/resources/js';
+			if ( file_exists( $path . '/underscore-before.js' ) ) {
```
bordoni: No need to check for file exists. It will always do.

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @return string The <script> tag.
+	 */
+	public function prevent_underscore_conflict( $tag, $handle ) {
+		if ( is_admin() || ! defined( 'TRIBE_EVENTS_FILE' ) ) {
+			return $tag;
+		}
+
+		if ( 'underscore' === $handle ) {
+			$path = plugin_dir_path( TRIBE_EVENTS_FILE ) . 'common/src/resources/js';
+			$dir = plugin_dir_url( TRIBE_EVENTS_FILE ) . 'common/src/resources/js';
```
bordoni: Use the base URL for common, because it might be loaded from Event Tickets and this would fatal.

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,45 @@
...
+			if ( file_exists( $path . '/underscore-before.js' ) ) {
+				$tag = "<script type='text/javascript' src='{$dir}/underscore-before.js'></script>\n" . $tag;
+			}
+			if ( file_exists( $path . '/underscore-after.js' ) ) {
```
bordoni: No need to check for file exists. It will always do.

src/Tribe/Main.php
```diff
@@ -549,6 +550,7 @@ public function bind_implementations() {
 		tribe_singleton( 'settings', 'Tribe__Settings', array( 'hook' ) );
 		tribe_singleton( 'ajax.dropdown', 'Tribe__Ajax__Dropdown', array( 'hook' ) );
 		tribe_singleton( 'assets', 'Tribe__Assets' );
+		tribe_singleton( 'assets.listen_load', 'Tribe__Load_Assets', array( 'hook' ) );
```
bordoni: It should use the same name as the class `tribe_singleton( 'load-assets', ...`

src/Tribe/Load_Assets.php
mitogh: Perfect, but there's a check above to prevent errors if constant is not defined it will not be fired. Thanks.

src/Tribe/Load_Assets.php
mitogh: How about: `Assets_Pipeline`  or `Assets_Sniffer` –– thoughts? Thanks

src/Tribe/Main.php
```diff
@@ -103,6 +103,8 @@ public function __construct( $context = null ) {
 
 		$this->plugin_url  = plugins_url( $parent_plugin_dir === $this->plugin_dir ? $this->plugin_dir : $parent_plugin_dir );
 
+		define( 'TRIBE_COMMON_PARENT_PLUGIN_URL', $this->plugin_url );
```
bordoni: No need to define a Constant here.

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,41 @@
...
+	 * taking place on the global namespace if lodash is present.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $tag The <script> tag for the enqueued script.
+	 * @param string $handle The script's registered handle.
+	 * @return string The <script> tag.
+	 */
+	public function prevent_underscore_conflict( $tag, $handle ) {
+		if ( is_admin() || ! defined( 'TRIBE_COMMON_PARENT_PLUGIN_URL' ) ) {
```
bordoni: Avoid this constant

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,41 @@
...
+			return $tag;
+		}
+
+		if ( 'underscore' === $handle ) {
+			$dir = plugin_dir_url( TRIBE_COMMON_PARENT_PLUGIN_URL ) . 'src/resources/js';
```
bordoni: use `$dir = Tribe__Main::instance()->plugin_url . 'src/resources/js';`

src/Tribe/Load_Assets.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @param string $handle The script's registered handle.
+	 * @return string The <script> tag.
+	 */
+	public function prevent_underscore_conflict( $tag, $handle ) {
+		if ( is_admin() || ! defined( 'TRIBE_COMMON_PARENT_PLUGIN_URL' ) ) {
+			return $tag;
+		}
+
+		if ( 'underscore' === $handle ) {
+			$dir = TRIBE_COMMON_PARENT_PLUGIN_URL . 'src/resources/js';
```
bordoni: use `$dir = Tribe__Main::instance()->plugin_url . 'src/resources/js';`

src/Tribe/Load_Assets.php
bordoni: I like Pipeline.

PR #610, opened 2018-02-07
Fix and improve some Field and Settings related functionalities

lucatume:
Ticket: n/a

Context: #81216

Please note that I've not added a changelog entry as pretty much all the work is internal and very specific.


src/Tribe/Post_Transient.php
```diff
@@ -9,211 +9,206 @@
...
-
-				// if there aren't any values, communicate that it did not fetch data from post transient
-				if ( ! is_array( $value ) || 0 === count( $value ) ) {
-					return false;
-				}
-
-				// grab the first value, because that's all we care about
-				$value = current( $value );
+		return
+			has_filter( 'tribe_post_meta_transient_' . $transient )
```
bordoni: no need to break the line between return and `has_filter`

lucatume: makes sense. I blame auto-formatting.

PR #611, opened 2018-02-09
Create a way for fields to be Displayed but Disabled by dependency

bordoni:
_Ref:_ [C#94708](http://central.tri.be/issues/94708)
---


src/resources/js/dependency.js
```diff
@@ -124,7 +124,8 @@
 
 					var active_class       = selectors.active.replace( '.', '' );
 					// Allows us to check a disabled dependency
-					var allowDisabled     = $dependent.data( 'dependencyCheckDisabled' ) || $dependent.is( '[data-dependency-check-disabled]' );
+					var allowDisabled      = $dependent.data( 'dependencyCheckDisabled' ) || $dependent.is( '[data-dependency-check-disabled]' );
+					var alwaysVisible      = $dependent.data( 'dependencyAlwaysVisibile' ) || $dependent.is( '[data-dependency-always-visible]' );
```
ggwicz: use `snake_case` instead of `camelCase` (unless there's a hyper-specific reason not to?)

bordoni: On JavaScript we are going with `camelCase`.

ggwicz: Oh damn—was not familiar with that standard change. Approving this PR, then!

PR #612, opened 2018-02-09
Prevent DST on UTC conversion of date times

mitogh:
Ticket => https://central.tri.be/issues/69784

Split dates before a UTC conversion in order to avoid a change on the
initial time.

### DIFF
--- a/src/Tribe/Timezones.php
+++ b/src/Tribe/Timezones.php
@@ -279,10 +279,29 @@ class Tribe__Timezones {
 		$new_datetime = date_create( $datetime, $local );
 
 		if ( $new_datetime ) {
-			$new_datetime->setTimezone( $utc );
-			$format = ! empty( $format ) ? $format : Tribe__Date_Utils::DBDATETIMEFORMAT;
-
-			return $new_datetime->format( $format );
+			if ( empty ( $format ) ) {
+				/**
+				 * In order to avoid problems with dates that fall into DST for local timeszones we need to split into: date & time
+				 * doing this we convert to UTC each in order to avoid a DST conversion based on date and time.
+				 *
+				 * This is important specially for recurring events as some of them might fall inside or outside of DST periods and as the conversion
+				 * might fall into timezones with DST PHP will automatically convert an hour from normal to DST or viceversa
+				 * in most cases the event that was supposed to be started at 12:00pm should remain at the same time
+				 * regardless if is DST or not.
+				 */
+				$date = date_create( $new_datetime->format( Tribe__Date_Utils::DBDATEFORMAT ), $local );
+				$time = date_create( $new_datetime->format( Tribe__Date_Utils::DBTIMEFORMAT ), $local );
+				$date->setTimezone( $utc );
+				$time->setTimezone( $utc );
+				return sprintf(
+					'%s %s',
+					$date->format( Tribe__Date_Utils::DBDATEFORMAT ),
+					$time->format( Tribe__Date_Utils::DBTIMEFORMAT )
+				);
+			} else {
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [Timezones.php:282](https://github.com/moderntribe/tribe-common/blob/cfee1274a19f8927147952232411921b752284d9/src/Tribe/Timezones.php#L282)

src/Tribe/Timezones.php
```diff
@@ -279,10 +279,29 @@ public static function to_utc( $datetime, $tzstring, $format = null ) {
...
+				/**
+				 * In order to avoid problems with dates that fall into DST for local timeszones we need to split into: date & time
+				 * doing this we convert to UTC each in order to avoid a DST conversion based on date and time.
+				 *
+				 * This is important specially for recurring events as some of them might fall inside or outside of DST periods and as the conversion
+				 * might fall into timezones with DST PHP will automatically convert an hour from normal to DST or viceversa
+				 * in most cases the event that was supposed to be started at 12:00pm should remain at the same time
+				 * regardless if is DST or not.
+				 */
+				$date = date_create( $new_datetime->format( Tribe__Date_Utils::DBDATEFORMAT ), $local );
```
bordoni: Does `date_create()` throw warnings when is not a good format?
If so, we want to wrap this on `try() catch`

mitogh: Right! I think we control the format but it might be wise to do it anyways. BTW do you know the error of the sniff above not sure what that really means because the value tested on empty is null maybe?

bordoni: That is a false positive, dunno why it does that.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [Timezones.php:282](https://github.com/moderntribe/tribe-common/blob/9ce65c346687cecbef9dc2ba338d85dbbe79a351/src/Tribe/Timezones.php#L282)
:x: | Don't leave error_log lying around | [Timezones.php:303](https://github.com/moderntribe/tribe-common/blob/9ce65c346687cecbef9dc2ba338d85dbbe79a351/src/Tribe/Timezones.php#L303)

src/Tribe/Timezones.php
```diff
@@ -279,10 +279,34 @@ public static function to_utc( $datetime, $tzstring, $format = null ) {
...
+					$time = date_create( $new_datetime->format( Tribe__Date_Utils::DBTIMEFORMAT ), $local );
+					$date->setTimezone( $utc );
+					$time->setTimezone( $utc );
+					return sprintf(
+						'%s %s',
+						$date->format( Tribe__Date_Utils::DBDATEFORMAT ),
+						$time->format( Tribe__Date_Utils::DBTIMEFORMAT )
+					);
+				} catch( Exception $e ) {
+					error_log( 'There was an error while date_create() was called.' );
```
bordoni: I would avoid the `error_log()` we have notice classes on common that we can use if we want.

mitogh: Okay perfect I'm going to remove it ;)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [Timezones.php:282](https://github.com/moderntribe/tribe-common/blob/2691aca93d24f30bc4d2a0f09881908701fb0ced/src/Tribe/Timezones.php#L282)

mitogh:
Ready for another set of 👀

PR #614, opened 2018-02-09
Add eventbrite origin and type for Global ID handling

sc0ttkclark:
https://central.tri.be/issues/98988


bordoni:
By the way this is supposed to go into `master` since we want to be able to sync it into EA Service.

PR #620, opened 2018-02-14
more robust checks on the context

lucatume:
Ticket: #99046

This PR:
1. adds the `Tribe__Utils__Array::get_in_any` method to get the specified indexes in any one of the specified values
2. updates the code of the `Tribe__Context` class to avoid notices in those cases where the post or post type might not be as expected

### DIFF
--- a/src/Tribe/Context.php
+++ b/src/Tribe/Context.php
@@ -44,7 +44,7 @@ class Tribe__Context {
 		}
 
 		if ( null !== $post_or_type ) {
-			$current_post = get_post();
+			$current_post = Tribe__Utils__Array::get( $_REQUEST, 'post', get_post() );
 
 			if ( is_numeric( $post_or_type ) ) {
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Context.php
```diff
@@ -44,25 +44,29 @@ public function is_editing_post( $post_or_type = null ) {
...
 			if ( count( array_filter( $post_types, 'is_numeric' ) ) === count( $post_types ) ) {
 				return ! empty( $post ) && in_array( $post->ID, $post_types );
 			}
 
-			if ( $is_post ) {
+			if ( $is_post && $post instanceof WP_Post ) {
 				$post_type = $post->post_type;
 			} else {
-				$post_type = Tribe__Utils__Array::get( $_GET, 'post_type', 'post' );
+				$post_type = Tribe__Utils__Array::get_in_any( $lookup, 'post_type', 'post' );
```
Camwyn: Out of curiosity - any reason other than tidiness to not just use the `$lookup` array here since it's small?

lucatume: the idea is that we might want to change that in the future and changing it in one place only, on line 47, seems like the most maintainable solution

PR #626, opened 2018-02-22
Fix the order of run for cron schedule/hook code

sc0ttkclark:
https://central.tri.be/issues/99680

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -330,7 +330,7 @@ class Tribe__Main {
 		add_action( 'wp_footer', array( $this, 'toggle_js_class' ) );
 
 		// Schedule cron if needed
-		add_action( 'admin_init', array( tribe( 'cron' ), 'schedule' ) );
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -329,7 +329,8 @@ public function add_hooks() {
 		add_filter( 'body_class', array( $this, 'add_js_class' ) );
 		add_action( 'wp_footer', array( $this, 'toggle_js_class' ) );
 
-		add_filter( 'cron_schedules', array( tribe( 'cron' ), 'filter_cron_schedules' ) );
+		// Schedule cron if needed
+		add_action( 'admin_init', array( tribe( 'cron' ), 'schedule' ) );
```
sc0ttkclark: Need feedback on which action this should run on, is admin_init good enough?

WP Codex says crons should be scheduled on activation stuff, but this is tribe-common so that's not necessarily an option.

bordoni: I like the `admin_init`, can we use `tribe_callback( 'cron', 'schedule' )`

PR #628, opened 2018-02-27
100199 Add Function for "Safely" Accessing the $wp_query Global

ggwicz:
**Ticket:** [**#100199**](http://central.tri.be/issues/100199)

**Related PRs:**
- _The Events Calendar:_ https://github.com/moderntribe/the-events-calendar/pull/1841
- _Events Calendar PRO:_ https://github.com/moderntribe/events-pro/pull/801

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -158,6 +158,30 @@ if ( ! function_exists( 'tribe_get_request_var' ) ) {
 	}
 }
 
+if ( ! function_exists( 'tribe_get_global_query_object' ) ) {
+	/**
+	 * Grabs the $wp_query global in a safe way with some fallbacks that help prevent fatal errors
+	 * on sites where themes or other plugins directly manipulate the $wp_query global.
+	 *
+	 * @since TBD
+	 *
+	 * @return object|boolean The $wp_query, the $wp_the_query if $wp_query empty, false otherwise.
+	 */
+	function tribe_get_global_query_object() {
+		global $wp_query;
+
+		if ( ! empty( $wp_query ) ) {
+			return $wp_query;
+		}
+
+		if ( ! empty( $_GLOBALS['wp_the_query'] ) ) {
+			return $_GLOBALS['wp_the_query'];
+		}
+
+		return false;
+	}
+}
+
 if ( ! function_exists( 'tribe_is_truthy' ) ) {
 	/**
 	 * Determines if the provided value should be regarded as 'true'.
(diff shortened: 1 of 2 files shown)


src/functions/utils.php
```diff
@@ -158,6 +158,30 @@ function tribe_get_request_var( $var, $default = null ) {
...
+
+		if ( ! empty( $wp_query ) ) {
+			return $wp_query;
+		}
+
+		if ( ! empty( $_GLOBALS['wp_the_query'] ) ) {
+			return $_GLOBALS['wp_the_query'];
+		}
+
+		return false;
```
mitogh: I think in this case `null` might be a better default as `wp_query` is an object I think if the object is not present `null` should be the default representation of an object.

src/functions/utils.php
```diff
@@ -158,6 +158,30 @@ function tribe_get_request_var( $var, $default = null ) {
...
+	/**
+	 * Grabs the $wp_query global in a safe way with some fallbacks that help prevent fatal errors
+	 * on sites where themes or other plugins directly manipulate the $wp_query global.
+	 *
+	 * @since TBD
+	 *
+	 * @return object|boolean The $wp_query, the $wp_the_query if $wp_query empty, false otherwise.
+	 */
+	function tribe_get_global_query_object() {
+		global $wp_query;
```
barryhughes: I don't have a strong feeling either way but since it caught my eye ... why not also import the `$wp_the_query` global, instead of mixing and matching styles (for one global you're doing it this way, for the other you're pulling from the `$GLOBALS` array)?

src/functions/utils.php
```diff
@@ -158,6 +158,30 @@ function tribe_get_request_var( $var, $default = null ) {
...
+
+		if ( ! empty( $wp_query ) ) {
+			return $wp_query;
+		}
+
+		if ( ! empty( $_GLOBALS['wp_the_query'] ) ) {
```
barryhughes: Shouldn't this be `$GLOBALS`?

barryhughes: (ie, no leading _)

ggwicz: Yes, I made a dumb mistake. Will fix!

src/functions/utils.php
ggwicz: Sure thing, will do.

ggwicz:
Addressed conflicts and implemented @barry and @crisoforo's feedback—this one's ready for another review! 🔍

mitogh:
I'm not sure if we should merge this one tho just because of the `hold` on this project @ggwicz

ggwicz:
Barry has greenlit this and a few other specific changes for the MR despite that hold—so, good note on calling it out Crisoforo, but I can confirm that this PR is allowed 👍

PR #631, opened 2018-03-01
Make sure JSON strings are single lines

mitogh:
**Ticket** https://central.tri.be/issues/99089

JSON Strings should be a single line string otherwise when there are multiple
carry returns in the middle will fail to parse the string to JSON.


barryhughes:
Note: fairly safe change where we're making an exception to the block on merging into Common (for M18.04).

PR #648, opened 2018-03-21
Allow skipping activation page welcome redirection with an option

lucatume:
Ticket: n/a

This PR adds a check for the `tribe_skip_welcome` option in the activation page code to avoid redirecting after a plugin activation; the non-skippable redirection breaks all acceptance tests and appending the `tribe-skip-welcome` query var to *any* request would be an exercise in futility. Adding the set of the option *once* before the tests is, instead, way more practical and maintainable.

### DIFF
--- a/src/Tribe/Admin/Activation_Page.php
+++ b/src/Tribe/Admin/Activation_Page.php
@@ -47,7 +47,7 @@ class Tribe__Admin__Activation_Page {
 	 * Listen for opportunities to show update and welcome splash pages.
 	 */
 	public function hooks() {
-		if ( true === ( (bool) get_option( 'tribe_skip_welcome' ) ) ) {
+		if ( tribe_is_truthy( get_option( 'tribe_skip_welcome' ) ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Admin/Activation_Page.php
```diff
@@ -47,6 +47,10 @@ public function __construct( array $args = array() ) {
 	 * Listen for opportunities to show update and welcome splash pages.
 	 */
 	public function hooks() {
+		if ( true === ( (bool) get_option( 'tribe_skip_welcome' ) ) ) {
```
bordoni: Lets save it as a `tribe_update_option`?

lucatume: Save it? You mean changing the option name?

bordoni: And pass it via `tribe_is_truthy`

lucatume: Oh, now I get it.
No, using `tribe_update_option` would mean dealing and have to update a serialized option which is non-trivial.
The idea here is to be able to run `wp option add tribe_skip_welcome 1` outside of PHP.

lucatume: `tribe_is_truthy` I can do, but the option has to be flat one to make sense; not a serialized one.

lucatume: Changed to use `tribe_is_truthy` for the check

PR #652, opened 2018-03-28
Set the timezone when the `DateTime` object is created

mitogh:
Ticket: https://central.tri.be/issues/98654

If the Timezone is provided and the `$date` is not a UNIX Timestamp format
the timezone object should be set on initial value instead as if not UTC is
used as default.

For UNIX timestamps it can't be provided on Object creation only once
object is already created.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,6 +4,8 @@
 
 = [4.7.11] TBD =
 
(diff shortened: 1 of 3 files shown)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for **, but this was pull requested against *release/M18.06*.

PR #658, opened 2018-03-30
Skip when a key is not set but not required

sc0ttkclark:
https://central.tri.be/issues/102545

### DIFF
--- a/src/Tribe/Formatter/Base.php
+++ b/src/Tribe/Formatter/Base.php
@@ -243,32 +243,38 @@ class Tribe__Formatter__Base implements Tribe__Formatter__Interface {
 			foreach ( array_keys( $format_map ) as $key ) {
 				$alias = ! empty( $format_map[ $key ]['alias'] ) ? $this->find_alias( $format_map[ $key ]['alias'], $value ) : $key;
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Formatter/Base.php
```diff
@@ -243,32 +243,38 @@ protected function format( $value, array $format_map, &$context = null ) {
 			foreach ( array_keys( $format_map ) as $key ) {
 				$alias = ! empty( $format_map[ $key ]['alias'] ) ? $this->find_alias( $format_map[ $key ]['alias'], $value ) : $key;
 
+				$key_or_alias_not_set = ! isset( $value[ $alias ] ) && ! isset( $value[ $key ] );
 				$key_or_alias_is_empty = empty( $value[ $alias ] ) && empty( $value[ $key ] );
 				$has_default_value = isset( $format_map[ $key ]['default'] );
 				$empty_is_allowed = ! empty( $format_map[ $key ]['allow_empty'] );
+				$is_required_key = ! empty( $format_map[ $key ]['required'] ) || $this->contains_required_keys( $format_map[ $key ] );
```
bordoni: should we use Array Get?

sc0ttkclark: Just following how it's being used now, not important to go refactoring that at the moment IMO

PR #660, opened 2018-04-03
Add base for async request handling and post thumbnail handler

lucatume:
Ticket: https://central.tri.be/issues/102323

Also: https://central.tri.be/issues/102195

This PR scaffolds the code base to use the [WP Background Processing package](https://github.com/A5hleyRich/wp-background-processing) to handle async requests and queueing.
In this PR I've set up the minimum amount of code needed to integrate the library in `common` and manage it via a Service Provider.
The first handler, the `Post_Thumbnail_Setter` one, handles requests to set, including download and upload, an image attachment as a post thumbnail as:

1. it's the single most consuming operation while processing CSV files (see ticket)
2. it's a perfect candidate for an async operation

How would more handlers evolve? Ideally all of MT's handlers should extend the `Tribe__Process__Handler` class and implement the `abstract` methods; I've chosen this solution as it seems the one requiring the less-effort to maintain and more re-usable; as the number of async process handlers grows having **all of them** `maybe_handle` would bog a site down so the Service Provider will perform some pre-flight check to only instance the required one, if required at all.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.18*, but this was pull requested against *release/M18.06*.

lucatume:
This is good for Code Review. Yes: I'm cheating the system to get around the ticket check; I'm a pirate.

src/Tribe/Service_Providers/Processes.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Handles the registration and creation of our async process handlers.
+ */
+class Tribe__Service_Providers__Processes extends tad_DI52_ServiceProvider {
+
+	/**
+	 * Binds and sets up implementations.
+	 */
+	public function register() {
+		if ( ! ( defined( 'DOING_AJAX' ) && DOING_AJAX && ! empty( $_REQUEST['action'] ) ) ) {
```
bordoni: Use doing Ajax from `Tribe__Main`
I would use also `tribe_get_request_var` to allow the check.

lucatume: the way the async request handling works is by sending `POST` requests to the `wp-admin-ajax` endpoint.
I feel now, reading it, the check should be **stricter**, on the `POST` array only, in place of **broader** (as `tribre_get_request_var` would allow) to take only the real supported case into account.

lucatume:
The last push restricts the check made by the `Processes` class to the `POST` superglobal only.

lucatume:
More iteration on this:
* I've deprected the `Tribe__Main::doing_ajax` method and moved it to the `Tribe__Context` class
* use `Tribe__Context::doing_ajax` in the `Processes` check
* I've replace the calls to `Tribe__Main::instance()->doing_ajax()` with calls to `tribe( 'context' )->doing_ajax()`

src/Tribe/Main.php
```diff
@@ -477,11 +474,9 @@ public static function post_id_helper( $post = null ) {
 	 * @return boolean
```
bordoni: Add a docblock for this.

lucatume:
Added missing doc block

PR #662, opened 2018-04-04
Add missing async request library files

lucatume:
Ticket: https://central.tri.be/issues/102323

Follows: https://github.com/moderntribe/tribe-common/pull/660

This PR adds the missing asyn library files... meh.


lucatume:
Please note that tests will fail here too due to https://github.com/moderntribe/tribe-common/pull/660#issuecomment-378552530

lucatume:
I've also fixed, in the last commit, an issue where with UTC offset conversion.

lucatume:
Added entries in the whitelist

package-whitelist.json
```diff
@@ -22,6 +22,11 @@
   "vendor/tribe-select2/*.js",
   "vendor/tribe-select2/*.gif",
   "vendor/tribe-select2/*.png",
+  "vendor/autoload.php",
+  "vendor/autoload_52.php",
+  "vendor/composer/*",
+  "vendor/lucatume/di52/src/**",
+  "vendor/a5hleyrich/wp-background-processing/**",
```
borkweb: We should probably only permit the files that are necessary for packaging `*.php` & `license.txt` for wp-background-processing. We can probably exclude .gitignore, composer.json, and the readme

lucatume: yep, definitely

lucatume: Ok added in https://github.com/moderntribe/tribe-common/pull/662/commits/a86107eeeab39fdccb536cee13275272286e7cb7

lucatume:
Changelog entry in https://github.com/moderntribe/tribe-common/pull/660/files#diff-eb6b6c90251ab33cee784713c451e6d8

PR #664, opened 2018-04-06
Feature/103577 - Select2 maximumSelectionSize

juanfra:
https://central.tri.be/issues/103577

Makes the select2 implementation compatible with maximumSelectionSize
http://select2.github.io/select2/#maximumSelectionSize

They can add the data attribute data-maximum-selection-size="3" in the template to have it working.

**Sorry! wrong branch for #663**


mitogh:
Code looks good to me. 

I think the only thing missing is a new entry on this file: 

- https://github.com/moderntribe/tribe-common/blob/release/M18.06/readme.txt

Here is some overall of how the format should be: 

```
typically fix,
 feature or                                     no need for
   tweak                                         spotfixes
    ↓                                                ↓
* Type - Short description (props if applicable) [#issue]
       ↑                                       ↑
    regular                                 no need
    hyphen,                                for a full
  not a fancy                            stop or period
    em-dash                                at the end
```

PR #667, opened 2018-04-10
Add filters to allow customization on countries names

mitogh:
Ticket: https://central.tri.be/issues/79880

First step in a series of making sure users can have multiple customization
on the names specially on the country names as they can be named different
or they might have multiple names know as "local names".


mitogh:
Filter names has been updated accordingly. 

Thanks @bordoni

PR #672, opened 2018-04-16
Replace Queue implementation with the new async one

lucatume:
Ticket: https://central.tri.be/issues/102195

This PR:
* removes the `Tribe__Queue` and accessory classes like the Worker; it was being used by RER code only so it should be ok
* puts in place the new `Tribe__Processs__Handler` and `Tribe__Process__Queue` base classes that we can extend in task-specific classes to handle async requests and queues (e.g. https://github.com/moderntribe/events-pro/pull/829/files#diff-d4a08b71fa37986f968d119d213dab93)
* introduces the `Tribe__Db` class to abstract Db information reading (and writing, possibly in the future)

Some features of the queueing system:
1. the background, cron and request independent, processing happens by sending, at the end of each batch process run, a non-blocking `POST` request to `admin-ajax.php`
2. the queue will process as many elements of the queue in a single request as fitting in the time and memory limits
3. I've extended the original implementation to add the possibility to stop a queue, get information about it, uniquely identify it
4. furthermore I've put in place "kinda" the same fragmentation logic used in the `Meta_Chunker` to avoid the last queue bottleneck: trying to send too much information in a single query thus running into the `max_allowed_packets` limit
5. since we are still working with the original code, I've extended, not replaced, the queue lends itself to chaining simply creating a different queue in the `complete` method. To make an example: in our most common case we have to match new data to existing data and update or create it, this could change using 2 queues: a first one will establish, for each piece of new data, if it's an update or a creation, once that is done for all the elements a second queue will process the updates and the changes.
6. the base `Tribe__Process__Queue` class covers most of the bases allowing extending classes, see https://github.com/moderntribe/events-pro/pull/829 for example, to only implement the particular `task` and `action` methods (optionally `complete` as per 5)

The biggest shift that should be done is in how we process queues: compute first, execute later. Tasks will be executed independently of each other, in a non-fixed order and should be independant of each other.


src/Tribe/Main.php
```diff
@@ -513,6 +507,8 @@ public function store_admin_notices( $page ) {
 	 * Runs tribe_plugins_loaded action, should be hooked to the end of plugins_loaded
 	 */
 	public function tribe_plugins_loaded() {
+		tribe()->register( 'Tribe__Service_Providers__Processes' );
```
bordoni: Why not `tribe_register_provider`?

src/Tribe/Process/Queue.php
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function update( $key, $data ) {
+		$meta = (array) get_transient( $key . '_meta' );
+		$done = $this->original_batch_count - count( $data );
+
+		set_transient( $key . '_meta', array_merge( $meta, array(
```
bordoni: Lets split this into multiple lines.

Maybe a get transient key method?

src/Tribe/Process/Queue.php
```diff
@@ -0,0 +1,316 @@
...
+		$per_frag        = max( (int) floor( count( $data ) / $frags_count ), 1 );
+
+		$split_data = array_chunk( $data, $per_frag );
+
+		if ( empty( $split_data ) ) {
+			return 0;
+		}
+
+		foreach ( $split_data as $i => $iValue ) {
+			$postfix = $i === 0 ? '' : "_{$i}";
```
bordoni: Yoda? (we need to make a decision if we will keep that little green standard)

src/Tribe/Process/Queue.php
```diff
@@ -0,0 +1,316 @@
...
+	 * Overrides the base `dispatch` method to allow for constants and/or environment vars to run
+	 * async requests in sync mode.
+	 *
+	 * @since TBD
+	 *
+	 * @return mixed
+	 */
+	public function dispatch() {
+		if (
+			( defined( 'TRIBE_NO_ASYNC' ) && true === TRIBE_NO_ASYNC )
```
bordoni: Should we use `tribe_is_truthy`? for both?

src/Tribe/Service_Providers/Processes.php
```diff
@@ -13,30 +13,48 @@ class Tribe__Service_Providers__Processes extends tad_DI52_ServiceProvider {
 	 * Binds and sets up implementations.
 	 */
 	public function register() {
-		if ( ! ( tribe( 'context' )->doing_ajax() && ! empty( $_POST['action'] ) ) ) {
+		if ( ! ( tribe( 'context' )->doing_ajax() && ! empty( $_REQUEST['action'] ) ) ) {
```
bordoni: Use `tribe_get_request_var` avoiding `$_COOKIES`

src/Tribe/Service_Providers/Processes.php
```diff
@@ -13,30 +13,48 @@ class Tribe__Service_Providers__Processes extends tad_DI52_ServiceProvider {
...
 			return;
 		}
 
-		$action = $_POST['action'];
+		$action = $_REQUEST['action'];
```
bordoni: Use `tribe_get_request_var` avoiding `$_COOKIES`

src/Tribe/Process/Queue.php
lucatume: In this specific case, I'd leave the check as is. Both are `developer` options that will be set to boolean or numeric values and I would skip the added overhead of a more complex function.

sc0ttkclark:
👍

PR #683, opened 2018-04-30
Feature/update datatables

Camwyn:
🎫 https://central.tri.be/issues/102465

Most of this is the structure change of datatables.

Things to note:
The js & css files are now combined, so we only have to load the two.
Had to remove `/vendor/datatables` from `.gitignore` to push the changes 🤦‍♂️


bordoni:
Do we have a why, for the merging?

PR #686, opened 2018-05-04
add the set_id method to the q

lucatume:
Ticket: n/a

This PR adds the `set_id` method to the queue class to allow clients to set a custom id, in place of the randomly generated one, on the q.
Mind that, in so doing, clients are taking charge of checking and managing said id uniqueness.


src/Tribe/Process/Queue.php
```diff
@@ -328,4 +337,22 @@ protected function get_post_args() {
...
 	}
+
+	/**
+	 * Sets the queue unique id.
+	 *
+	 * When using this method the client code takes charge of the queue id uniqueness;
+	 * the class will not check it.
+	 *
+	 * @param string $queue_id
+	 *
```
bordoni: `@since`

PR #691, opened 2018-05-04
Add `tribe_set_time_limit()` wrapper function to prevent errors/warnings

juanfra:
https://central.tri.be/issues/64183

Add `tribe_set_time_limit()` wrapper function to prevent errors from `set_time_limit()`.


src/functions/template-tags/general.php
```diff
@@ -704,3 +704,22 @@ function tribe_is_frontend() {
...
+if ( ! function_exists( 'tribe_set_time_limit' ) ) {
+	/**
+	 * Wrapper for set_time_limit to suppress errors
+	 *
+	 * @since 4.7.12
+	 *
+	 * @param int $limit Time limit.
+	 */
+	function tribe_set_time_limit( $limit = 0 ) {
+		if (
```
bordoni: Invert the conditional and `return false;`, and after the conditional use `return @set_time_limit...`

juanfra:
I still need @bordoni's approval :)

PR #703, opened 2018-05-17
Adjust `Tribe__Admin__Helpers::is_screen()` return

juanfra:
https://central.tri.be/issues/105217

I just got this one back from QA because the notice I was supposed to add (on all the "events" pages) in the admin pages, wasn't being displayed on the "Events > Tags" screen.

The problem was that `is_screen()` wasn't contemplating that case. I just reworked the conditional to make it work as it should.


bordoni:
Just make sure we highlight, this change in code can have huge impact, since this is used everywhere and doens't have proper testing.

juanfra:
The issue ended up passing to pending smoketest directly.

If you check it's just a change in the logic so, it can return true if the post_type of the current page is `tribe_events` or any returned by `Tribe__Main::get_post_types()`. There's a `return false;` right after that conditional.

For the specific case of "Events > Tags" (`wp-admin/edit-tags.php?taxonomy=post_tag&post_type=tribe_events`) is not considered by any preceding conditional, and with this change if would be considered, gien that it will actually `return true` if the $post_type is there.

You can check the changes:

```
if ( ! in_array( $current_screen->post_type, $defaults ) ) {
	return false;
}
```
for

```
if ( in_array( $current_screen->post_type, $defaults ) ) {
	return true;
 }
```

there's a return false right after that.

juanfra:
Let me know if I should merge directly, or if I have to raise up any special type of concern to QA. I don't know where this is used. But the change makes sense to me.

bordoni:
Totally, it's just that this specific method has implications on basically all assets and some loading of Aggregator and tickets, I would love to have a quick pass of these pages to check if we are good still, not a concern for the merge.

juanfra:
I just saw that the admin page titles defaults to "New Import" after this. Perhaps it's a better idea to make a ticket and work on it so it can be properly QAed?

bordoni:
Yeah lets do that I think. We might want to QA in branch this one.

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.08*, but this was pull requested against *release/M18.10*.

PR #718, opened 2018-06-13
Allow more empty validation handling

sc0ttkclark:
https://central.tri.be/issues/106026

### DIFF
--- a/src/Tribe/Validator/Base.php
+++ b/src/Tribe/Validator/Base.php
@@ -12,6 +12,24 @@ class Tribe__Validator__Base implements Tribe__Validator__Interface {
 	 *
 	 * @return bool
 	 */
+	public function is_not_null( $value ) {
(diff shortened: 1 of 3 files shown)


src/Tribe/Validator/Base.php
```diff
@@ -7,6 +7,24 @@
  */
 class Tribe__Validator__Base implements Tribe__Validator__Interface {
 
+	/**
+	 * @param mixed $value
+	 *
+	 * @return bool
+	 */
+	public function is_not_null( $value ) {
+		return null !== $value;
```
mitogh: How about: 

```php
return ! $this->is_null( $value );
```

sc0ttkclark: That would be cool but I feel it unnecessarily adds to the call stack.

PR #723, opened 2018-06-22
Deprecate filter as part of C#105116

cliffordp:
https://central.tri.be/issues/105116

As part of https://github.com/moderntribe/the-events-calendar/pull/2046

### DIFF
--- a/src/Tribe/Deprecation.php
+++ b/src/Tribe/Deprecation.php
@@ -30,6 +30,7 @@ class Tribe__Deprecation {
 	protected $deprecated_filters = array(
 		'tribe_cost_regex'            => array( '4.3', 'tribe_events_cost_regex' ),
 		'tribe_rewrite_prepared_slug' => array( '4.3', 'tribe_events_rewrite_prepared_slug' ),
+		''                            => array( 'TBD', 'tribe_events_linked_post_type_meta_key' ),
 	);
 
 	/**


src/Tribe/Deprecation.php
```diff
@@ -30,6 +30,7 @@ class Tribe__Deprecation {
 	protected $deprecated_filters = array(
 		'tribe_cost_regex'            => array( '4.3', 'tribe_events_cost_regex' ),
 		'tribe_rewrite_prepared_slug' => array( '4.3', 'tribe_events_rewrite_prepared_slug' ),
+		''                            => array( 'TBD', 'tribe_events_linked_post_type_meta_key' ),
```
bordoni: Why empty key? That doesn't seem right.

cliffordp: @bordoni this property's docblock states:
> Use the format `<new_filter_tag> => ...`

There is no new filter. What should I do instead?

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.11*, but this was pull requested against *release/M18.10*.

src/Tribe/Deprecation.php
barryhughes: Interesting one! I can think of two possibilities here. First, and requiring the least work, we could use a convention like `No replacement {tag}`:

- Give us a unique key
- Will result in safe (and useful-when-debugging) information being returned by _Tribe__Deprecation::get_filter_for_deprecated_tag()_

Alternatively, requiring a _little_ more work, if there is no replacement we could fall back on numerical indexing, ie:

```php
array(
 		'tribe_cost_regex'            => array( '4.3', 'tribe_events_cost_regex' ),
 		'tribe_rewrite_prepared_slug' => array( '4.3', 'tribe_events_rewrite_prepared_slug' ),
		array( 'TBD', 'tribe_events_linked_post_type_meta_key' ), # No replacement
);
```

If we did this, we'd also want to update our _Tribe__Deprecation::get_FILTER|ACTION_for_deprecated_tag()_ methods to detect that the key is numerical. They can then dynamically compose a return value like `'No alternative for $tag`.

PR #724, opened 2018-06-22
#106000 - move method to common to sanitize a multidimensional array

jesseeproductions:
_Ref:_ [C#106000](https://central.tri.be/issues/106000)

Used in Pro: https://github.com/moderntribe/events-pro/pull/923

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -3,6 +3,7 @@
 == Changelog ==
 
 = [4.7.16] TBD =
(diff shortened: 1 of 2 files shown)


Camwyn:
Any reason to not use `array_walk_recursive`?

jesseeproductions:
I think that could work too, I am more comfortable working and understanding it this way. I believe I would still have to rebuild the array using it too.

PR #725, opened 2018-06-25
Add filter for tribe plugins get list function in Tribe__Plugins

niconerd:
https://central.tri.be/issues/69581

### DIFF
--- a/src/Tribe/Plugins.php
+++ b/src/Tribe/Plugins.php
@@ -126,7 +126,7 @@ if ( ! class_exists( 'Tribe__Plugins' ) ) {
 			 *
 			 * @since TBD
 			 *
-			 * @return array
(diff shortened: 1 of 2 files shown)


src/Tribe/Plugins.php
```diff
@@ -121,7 +121,14 @@ public function get_plugin_by_class( $main_class ) {
 		 * @return array
 		 */
 		public function get_list() {
-			return $this->tribe_plugins;
+			/**
+			 * Gives an opportunity to filter the list of tribe plugins
+			 *
+			 * @since TBD
+			 *
+			 * @return array
```
barryhughes: This should be a `@param` tag (we're describing what's available to callbacks in this docblock, not what the method returns).

PR #736, opened 2018-07-12
Add stuck queue detection and garbage collection

lucatume:
Ticket: #110172

This PR adds methods on the queue base class to detect "stuck" queues (that did nothing for a period of time) and, eventually, kill them


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *1.0.23.2 (Hotfix)*, but this was pull requested against *release/M18.11*.

bordoni:
Add changelog.

src/Tribe/Process/Queue.php
```diff
@@ -127,12 +128,52 @@ public static function action() {
...
+		$limit = (int) apply_filters( 'tribe_process_queue_time_limit', 300 );
+
+		if ( ! empty( $queue_status['last_update'] ) && is_numeric( $queue_status['last_update'] ) ) {
+			$is_stuck = time() - (int) $queue_status['last_update'] > $limit;
+		} else {
+			$queue_status['last_update'] = time();
+			set_transient( $queue_id . '_meta', $queue_status->to_array(), DAY_IN_SECONDS );
+		}
+
+		return $is_stuck;
```
barryhughes: Not actually sure if this would be useful in practice, going to ask anyway: any merit in making the return value itself filterable (and not just $limit)?

bordoni:
Also tests are annoyed at this change.

src/Tribe/Process/Queue.php
lucatume: Filtering results is rarely a bad idea; I can think of a scenario where, to clean all queues (whether they are active or not) one could filter the `stuck` status to clean them all.

lucatume:
added filtering and tests

PR #740, opened 2018-07-23
#63144 - modify load_plugin_textdomain() to use relative path not fil…

jesseeproductions:
…e path

_Ref:_ [C#63144](https://central.tri.be/issues/63144)

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -366,7 +366,7 @@ class Tribe__Main {
 		$plugin_rel_path = WP_LANG_DIR . '/plugins/';
 
 		/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -363,20 +363,20 @@ public function load_text_domain( $domain, $dir = false ) {
 		}
 
 		$locale = get_locale();
-		$mofile = WP_LANG_DIR . '/plugins/' . $domain . '-' . $locale . '.mo';
+		$plugin_rel_path = WP_LANG_DIR . '/plugins/';
```
bordoni: Are we sure not passing the correct file works?Or is it something specific to Windows?

PR #741, opened 2018-07-25
Add and update Array and Validator methods

lucatume:
Ticket:  https://central.tri.be/issues/94326

This PR updates and adds some methods to the `Tribe__Utils__Array` and `Validator` classes to support the ticket functionalities.

### DIFF
--- a/src/Tribe/Validator/Base.php
+++ b/src/Tribe/Validator/Base.php
@@ -184,4 +184,24 @@ class Tribe__Validator__Base implements Tribe__Validator__Interface {
 
 		return in_array( $post_status, $post_stati );
 	}
+
+	/**
+	 * Whether a list or array of only contains positive integers or not.
+	 *
+	 * @since TBD
+	 *
+	 * @param  string|array $list
+	 * @param string $sep The separator used in the list to separate the elements; ignored if
+	 *                                  the input value is an array.
+	 *
+	 * @return bool
+	 */
+	public function is_positive_int_list( $list, $sep = ',' ) {
+		$sep      = is_string( $sep ) ? $sep : ',';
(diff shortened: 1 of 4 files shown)


src/Tribe/Validator/Base.php
```diff
@@ -184,4 +184,24 @@ public function is_post_status( $post_status ) {
...
+	 *
+	 * @return bool
+	 */
+	public function is_positive_int_list( $list, $sep = ',' ) {
+		$sep      = is_string( $sep ) ? $sep : ',';
+		$elements = Tribe__Utils__Array::list_to_array( $list, $sep );
+
+		$valid = array_filter( $elements, array( $this, 'is_positive_int' ) );
+
+		return ! empty( $elements ) && count( $valid ) === count( $elements );
```
mitogh: I think using an early return will help to make this last line more clean and avoid the call to `array_filter`

```php
if ( empty( $elements ) ) {
	return false;
}

$valid = array...

return count( $valid ) && count( $elements );
````

lucatume: `array_filter` will internally check the array to iterate on before iterating on it so, while explicit and more readable, the call to `empty` (which is really a number of checks), is not more performant. The real check to make here would be a `count($elements) > 0` but, again, it's already done internally.

mitogh: Right; however my comment was more in the sense that avoding the `!` operation will provide a more natural flow for reading the code as it read as: 

1. If empty `return false`
1. Filter elements
1. Return `if both have the same length`

Instead of.

1. Filter elements
1. Return if elements is not empty and `if both have the same length`.

In the end, is a matter of personal preference instead of a performance bottle neck. 

So I think we should be fine to move forward.

PR #745, opened 2018-07-31
Correct logic when enqueueing with multiple conditionals

Camwyn:
found while working on 🎫 https://central.tri.be/issues/111405

I believe the logic is wrong, they way it's set up it only requires one condition to pass?

Essentially, we're using an `OR` where we should use an `AND`

PR'd against this branch as it's the one used in Loxi

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -166,7 +166,7 @@ class Tribe__Assets {
 			foreach ( $asset->conditionals as $conditional ) {
 				$enqueue = call_user_func( $conditional );
 				// this may return false or null or? Only enqueue if they all pass by returning a truthy value
-				if ( empty( $enqueue ) ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Assets.php
```diff
@@ -165,7 +165,8 @@ public function enqueue( $forcibly_enqueue = null ) {
 			// If we have a set of conditionals we loop on then and get if they are true
 			foreach ( $asset->conditionals as $conditional ) {
 				$enqueue = call_user_func( $conditional );
-				if ( $enqueue ) {
+				// this may return false or null or? Only enqueue if they all pass by returning a truthy value
+				if ( empty( $enqueue ) ) {
```
bordoni: Possibly a case of `tribe_is_truthy`?

Camwyn: Man, always forget that one! YES.

Also - we might discuss adding a way to implement AND/OR in here?

PR #747, opened 2018-08-01
Add checksum functions and tests

lucatume:
Ticket: https://central.tri.be/issues/111525

This PR adds two supporting functions and tests:
* `tribe_post_checksum` will return the checksum for a single post (any post)
* `tribe_posts_checksum` (w/ an "s") will return the checksum of a group of posts

To avoid overhead the latter uses a direct db query.


src/functions/template-tags/general.php
```diff
@@ -704,3 +704,100 @@ function tribe_is_frontend() {
...
+		if ( empty( $post_ids ) ) {
+			return null;
+		}
+
+		/** @var wpdb $wpdb */
+		global $wpdb;
+
+		$prepared = array();
+		foreach ( $post_ids as $post_id ) {
+			if ( $post_id instanceof WP_Post ) {
```
sc0ttkclark: Why query at all if we already have access to post_modified?

src/functions/template-tags/general.php
```diff
@@ -704,3 +704,100 @@ function tribe_is_frontend() {
...
+	 *
+	 * @return null|string The checksum string or `null` if the post is not valid or no
+	 *                     fragments are passed.
+	 */
+	function tribe_post_checksum( $post_id, $use_frags = array( 'ID', 'post_modified' ) ) {
+		if ( empty( $post_id ) || empty( $use_frags ) ) {
+			return null;
+		}
+
+		$post = get_post( $post_id );
```
sc0ttkclark: If $post_id is already an instance of WP_Post, is this function call necessary?

lucatume: It's just convenient; if it's already an instance of `WP_Post` it will be returned: I would simply be duplicating the check made in the `get_post` function doing it here.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Space found before square bracket; expected &quot;$results[&quot; but found &quot;$results [&quot; | [general.php:790](https://github.com/moderntribe/tribe-common/blob/45e808ddbbf23b35ec4f89267835e45d0ff75815/src/functions/template-tags/general.php#L790)

src/functions/template-tags/general.php
lucatume: Right. I've updated the code to handle the case where the objects are all instances of `WP_Post`

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Space found before square bracket; expected &quot;$results[&quot; but found &quot;$results [&quot; | [general.php:790](https://github.com/moderntribe/tribe-common/blob/290e86366690ab2d348b64bfbbadac2421ac155b/src/functions/template-tags/general.php#L790)

lucatume:
I've updated the code and this is good for another look

PR #755, opened 2018-08-13
Fix issues emerged in QA

lucatume:
Tickets:
* https://central.tri.be/issues/108023
* https://central.tri.be/issues/109524
* https://central.tri.be/issues/108022
* https://central.tri.be/issues/109917

This PR fixes some issues emerged during the QA phase:
* make sure that the case where a date is empty is properly handled when filling in the date details of the response; previously it would return an array of `1970-...` dates: it will now return an array where all details are empty
* make sure the case where CSV lists are passed from REST requests are handled while casting them to arrays; before the URL encoded version of `USD,CHF` (`USD%2CHF`) would fail to processed into the `[ 'USD', 'CHF' ]` array; the update fixes it by decoding all input string lists of the `Tribe__Utills__Array::list_to_array` method


mitogh:
LGTM 👍

PR #757, opened 2018-08-15
Introduce Stack finding and better filters for Template class 

bordoni:
_Ref:_ [C#112478](http://central.tri.be/issues/112478)
---


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Sprint 1*, but this was pull requested against *release/M18.12*.

bordoni:
Fixed the Issue number

src/Tribe/Template.php
```diff
@@ -273,7 +274,164 @@ public function merge_context( $context = array(), $file = null, $name = null )
...
+	 *
+	 * @since  TBD
+	 *
+	 * @return array
+	 */
+	protected function get_template_path_stack() {
+		$stack = array();
+
+		// Only look into public folders if we tell to use stack
+		if ( $this->use_stack ) {
```
mitogh: Just few thoughts on this one in particular as the overall PR seems great but wanted to have your comments on the following items.

- Name of the structure, in this particular case is referred as stack, however the behavior of the structure is the one from a Queue as it's FIFO (First in first out) so I think the name of stack is great however I think the operations should be LIFO (Last in first out). 

- Data structure I think using an array to represent the stack will give a bit of flexibility instead like: 

```php
$stack[] = array(
	'type' => 'plugin',
	'path' => '',
);
```

- In terms of the theme / child theme I think a nested array structure have a bit more sense, however I'm not 100% confident as will introduce complexity during the iteration of the loop, this is what I mean.


```php
$stack[] = array(
	'type' => 'theme'
	'path' => '',
	'child' => array(
		'type' => 'theme',
		'path' => '',
	)
)
```

So during the loop of the stack if is a theme and has a child the child should take priority, I think this one gives a clear sense of nested structure and dependency on the themes.

bordoni: Yeah I imagined it would be a good idea to tackle and move into a more args like structure for the Stack.

Will address the naming and the how we store it. As of the Child, I will avoid it because it adds complexity we dont need.

bordoni:
Sorted the problem's mentioned by @mitogh

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.13*, but this was pull requested against *release/M18.12*.

PR #824, opened 2018-11-15
Removed closing parentheses

andrasguseo:
Removed closing parentheses, which is not needed in either case as it is being added already in the called method on line 416.


andrasguseo:
Central ticket:
https://central.tri.be/issues/118275

PR #832, opened 2018-11-20
fix blue shadow for creatable select

paulmskim:
For ticket: https://central.tri.be/issues/117466

### DIFF
--- a/src/modules/elements/creatable-select/style.pcss
+++ b/src/modules/elements/creatable-select/style.pcss
@@ -32,6 +32,11 @@
 		& > input {
 			margin: 0;
 			line-height: 1.5;
+
+			&,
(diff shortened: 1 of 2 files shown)


src/modules/elements/creatable-select/style.pcss
```diff
@@ -32,6 +32,11 @@
 		& > input {
 			margin: 0;
 			line-height: 1.5;
+
+			&,
```
nealfennimore: Is this needed?

paulmskim: there's a transparent box shadow on the input when not focused, so not necessarily needed, but added it to remove all box shadows.

PR #852, opened 2018-11-28
Mock has_blocks function when Gutenberg is not present

mitogh:


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task | [utils.php:493](https://github.com/moderntribe/tribe-common/blob/0df904965812595682a4bf002e1b669a05dfba72/src/functions/utils.php#L493)

src/functions/utils.php
```diff
@@ -486,3 +486,34 @@ function tribe_post_excerpt( $post ) {
...
 	}
 }
+
+/**
+ * Create a function to mock the real function if the extension or Beta is not present.
+ *
+ * @todo remove this function after WP 5.0 is released
+ *
+ */
+if ( ! function_exists( 'has_blocks' ) ) {
```
juanfra: Perhaps we can add the WP version compare to the conditional? When 5.0 is out we'll still have people using WP < 5.0 and they will still get a fatal if we remove it

mitogh: Oh that's right since even if 5.0 is there doesn't mean people is using.

However I think adding the conditional here for that version does not make too much difference as the main point is to make sure the function is there as even with the `gutenberg` extension this function can be present and adding an `||` conditional to the above does not provide more value IMHO, but let me know if you think otherwise.

mitogh: However we might need to update the wording around the `@todo`?

juanfra: Yeah, adding to the conditional won't do much. What I meant is basically, is that if we're mocking this function to prevent a fatal we cannot remove the function once 5.0 is out. Perhaps  just removing the `todo` is fine. Ideally people will update WP, but if they don't and they update our plugins they'll end up facing the fatal.

PR #854, opened 2018-11-28
Speed up and secure async process feature detection

lucatume:
Ticket: n/a

I found out, while working on another issue, that the feature detection of async process support could be sped up and made more robust to race conditions.

### DIFF
--- a/src/Tribe/Feature_Detection.php
+++ b/src/Tribe/Feature_Detection.php
@@ -12,12 +12,20 @@
  * @since 4.7.23
  */
 class Tribe__Feature_Detection {
+
 	/**
 	 * The name of the transient storing the support check results.
 	 *
 	 * @var string
 	 */
 	public static $transient = 'tribe_feature_detection';
+
+	/**
+	 * The name of the option that will be used to indicate a feature detection is running.
+	 *
+	 * @var string
+	 */
+	protected $lock_option_name;
 
 	/**
 	 * Checks whether async, AJAX-based, background processing is supported or not.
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | empty() and isset() must only contain variables (PHP 5.2 compatibility) | [Feature_Detection.php:133](https://github.com/moderntribe/tribe-common/blob/28218b5d9dc7866e1bfcbbdfcf6f8974f535ff73/src/Tribe/Feature_Detection.php#L133)

src/Tribe/Feature_Detection.php
```diff
@@ -52,33 +60,78 @@ public function supports_async_process( $force = false ) {
...
 			|| ( is_array( $cached ) && ! isset( $cached['supports_async_process'] ) )
 		) {
-			/*
-			 * Build and dispatch the tester: if it works a transient should be set.
-			 */
-			$tester = new Tribe__Process__Tester();
-			$tester->dispatch();
+			if ( ! $this->has_lock() ) {
+				// Let's avoid race conditions by running two or more checks at the same time.
+				$this->lock();
```
mitogh: Just wondering if the lock mechanism might be slower than the actual task as the lock is done via a write in the DB, so if a instance variable might be a good idea as well combined with the DB

lucatume: Writing to the DB has the precise purpose of avoiding other parallel running processes, running in another PHP thread all-together, from running.
When it comes to different PHP threads then the DB is the only source of truth (excluding the use of proc semaphores that might not be enabled at all).

mitogh: I see so this one might be running in a multi thread env as well? 🎉

lucatume: This code is not made to run in multi-thread but an AJAX call could fire shortly after the requst for a page, or cron, started. On average an admin page requests triggers about ~4 more requests.

PR #858, opened 2018-11-29
Fix/117446 tec date format

paulmskim:
For tickets:
- https://central.tri.be/issues/ 117446
- https://central.tri.be/issues/ 119017


src/modules/utils/globals.js
```diff
@@ -23,9 +23,10 @@ export const editor = () => tec().editor || {};
 export const settings = () => tec().settings || {};
 export const mapsAPI = () => tec().googleMap || {};
 export const priceSettings = () => tec().priceSettings || {};
+export const tecDateSettings = () => tec().dateSettings || {};
```
mitogh: Looks good however it might be simpler to call `tec()` where is used instead of creating a new variable for that setting.

paulmskim: every other setting is using this convention so i've followed that. also it's handy to have this as if tec isn't activated, it returns a blank object and the property is undefined (makes truthy/falsey checking easy)

tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Sprint 6*, but this was pull requested against *release/M18.18*.

PR #877, opened 2018-12-06
Make sure `func_get_args` is assigned to a variable

mitogh:


juanfra:
Let's add a readme entry

PR #880, opened 2018-12-06
Fix/117428 time picker formats

paulmskim:
For ticket: https://central.tri.be/issues/117428


src/modules/elements/time-picker/element.js
```diff
@@ -184,18 +187,18 @@ TimePicker.propTypes = {
 	 * using 24h clock in hh:mm format
 	 * e.g. 00:24, 03:57, 21:12
 	 */
-	current: TribePropTypes.timeFormat.isRequired,
-	min: TribePropTypes.timeFormat,
-	max: TribePropTypes.timeFormat,
-	start: TribePropTypes.timeFormat.isRequired,
+	allDay: PropTypes.bool,
+	current: TribePropTypes.string,
```
mitogh: Question here what's the difference with `PropTypes.string`

src/modules/utils/moment.js
```diff
@@ -13,6 +13,8 @@ import {
 	string,
 } from '@moderntribe/common/utils';
 
+export const TIME_FORMAT = 'h:mm a';
```
mitogh: Can we add a test for this one?

src/modules/elements/time-picker/element.js
paulmskim: We're changing the timepicker component from a time input to a text input. Because the format won't always be a time format, let alone a 24-hour time format, it needs to be a string.

src/modules/utils/moment.js
paulmskim: There is a test for this, see https://github.com/moderntribe/tribe-common/pull/880/files#diff-1af408429616daceb23d7d1da2045fcdR31

src/modules/elements/time-picker/element.js
nealfennimore: Do we have a `TribePropTypes.string` in common? I don't see it in `M18.19`

paulmskim: you're right, this should be `PropTypes.string`

PR #881, opened 2018-12-06
Fix the tooltip taking the full witdh

juanfra:
https://central.tri.be/issues/118883

Before: https://cloudup.com/cy8s1UEXjaJ
After: https://cloudup.com/ciKWaQX_VK6

### DIFF
--- a/src/modules/elements/tooltip/style.pcss
+++ b/src/modules/elements/tooltip/style.pcss
@@ -1,0 +1,14 @@
+.components-tooltip {
+
+	.components-popover__content {
+		white-space: normal;
+		min-width: 350px;
+		max-width: 600px;
+	}
+
+	&:not(.is-mobile) {
+		.components-popover__content {
+			min-width: 350px;
+		}
+	}
+}
--- a/src/modules/elements/tooltip/element.js
+++ b/src/modules/elements/tooltip/element.js
@@ -14,6 +14,7 @@ import { Tooltip as WpTooltip } from '@wordpress/components';
  * Internal dependencies
  */
 import { Button } from '@moderntribe/common/elements';
+import './style.pcss';
 
 class Tooltip extends PureComponent {
 	static defaultProps = {


src/modules/elements/tooltip/style.pcss
```diff
@@ -0,0 +1,14 @@
+.components-tooltip {
```
mitogh: Should this one be namespace to affect only `tribe` items?

juanfra: The div is generated outside the containers

juanfra: As far as I see it's our wrapper for `WpTooltip`

mitogh: Do you think it might be a good idea to attach a class to the wrapper otherwise seems like a good fix for the general tooltip if other components uses it  with long text.

paulmskim: https://github.com/WordPress/gutenberg/tree/master/packages/components/src/tooltip

The way the component is, there is no way to namespace or add a class to the tooltip.

src/modules/elements/tooltip/style.pcss
```diff
@@ -0,0 +1,14 @@
...
+
+	.components-popover__content {
+		white-space: normal;
+		min-width: 350px;
```
paulmskim: Curious why we have a `min-width` here and below. If a user has content that is really short, we'll still have a tooltip that's `350px` wide (and a lot of space).

paulmskim: So my point isn't really to use a different number, my question is rather why we have it.
- if we have short content, a min width will make the width to the min width value, which adds space. This seems suboptimal. 
- if we have long content, the min width doesn't matter. Only the max width does, which we have set.

juanfra: Oh, right. Yeah, I basically used that number because, if not, it's falling back to the number set here https://github.com/moderntribe/tribe-common/blob/f4fb8b981be2c0ed4740ead6a4b270450ac6cc45/src/modules/elements/time-picker/style.pcss#L56 and I'm not sure if that has a specific purpose.

paulmskim: I think that's got another class attached to it, `.tribe-editor__timepicker__dialog`, so I don't think that should affect this tooltip popover, right?

paulmskim: Gotcha. This is an issue with text-based width and absolutely-positioned items. In this case your max-width will never be used then (the width will always be 350px). Anyway, this should be fine.

PR #891, opened 2018-12-12
Add function used to remove innser blocks

mitogh:
- Add utility to remove an inner block
- Refactor and create function to replace content text in a post.
- Update function `is_classic_editor` to support latest version

Ticket: https://central.tri.be/issues/119426

### DIFF
--- a/src/Tribe/Editor.php
+++ b/src/Tribe/Editor.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Initialize Gutenberg editor blocks
  *
@@ -27,10 +28,10 @@ class Tribe__Editor {
 	 */
 	public function should_load_blocks() {
 		return (
-			$this->is_gutenberg_active()
-			|| $this->is_wp_version()
-		)
-		&& $this->is_blocks_editor_active();
+			       $this->is_gutenberg_active()
+			       || $this->is_wp_version()
+		       )
+		       && $this->is_blocks_editor_active();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Editor/Utils.php
```diff
@@ -41,25 +41,53 @@ public function to_block_name( $name = '' ) {
...
 			'ID'           => $post_id,
-			'post_content' => preg_replace(
-				'/^\s*<!-- ' . $block_name . ' .* \/-->\s*$/gm',
-				'',
-				$content
-			),
-		);
-
-		return wp_update_post( $args );
+			'post_content' => preg_replace( $pattern, $replacement, $content ),
```
ggwicz: Maybe check if preg_replace returns a new version of the `post_content` string **first**, and then `wp_update_post()` if so? So that it only modifies the post's content if the replacement actually runs...

src/Tribe/Editor.php
```diff
@@ -26,11 +27,7 @@ class Tribe__Editor {
 	 * @return bool
 	 */
 	public function should_load_blocks() {
-		return (
-			$this->is_gutenberg_active()
-			|| $this->is_wp_version()
-		)
-		&& $this->is_blocks_editor_active();
+		return ( $this->is_gutenberg_active() || $this->is_wp_version() ) && $this->is_blocks_editor_active();
```
bordoni: Lets move back to multiple lines.

src/Tribe/Editor/Utils.php
bordoni: Agreed on that, so we can check if it finds something matching the patterns first.

src/Tribe/Editor/Utils.php
```diff
@@ -41,25 +41,53 @@ public function to_block_name( $name = '' ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param        $post_id
+	 * @param        $pattern
+	 * @param string $replacement The string used to replace the value of the searched block
+	 *
+	 * @return bool
+	 */
+	public function update_post_content( $post_id, $pattern, $replacement = '' ) {
```
bordoni: Also make this Private method, since it should only be accessed by the other method.

src/Tribe/Editor/Utils.php
```diff
@@ -41,25 +41,53 @@ public function to_block_name( $name = '' ) {
 	 *
 	 * @param        $post_id
 	 * @param string $block_name
+	 * @param string $replacement
 	 *
 	 * @return bool
 	 */
-	public function remove_block( $post_id, $block_name = '' ) {
+	public function remove_block( $post_id, $block_name = '', $replacement = '' ) {
+		$patttern = '/^\s*<!-- ' . $block_name . '.*\/-->\s*$/gm';
```
bordoni: Is it case-insensitive?

src/Tribe/Editor/Utils.php
```diff
@@ -41,25 +41,53 @@ public function to_block_name( $name = '' ) {
...
+	 * @since TBD
+	 *
+	 * @param        $post_id
+	 * @param        $block_name The name of the block
+	 * @param string $replacement The string used to replace the value of the searched block
+	 *
+	 * @return bool
+	 */
+	public function remove_inner_blocks( $post_id, $block_name, $replacement = '' ) {
+		$pattern = '/^\s*<!-- ' . $block_name . '.*-->\s.*<!-- \/' . $block_name . ' -->/ms';
```
bordoni: Is it case-insensitive?

src/Tribe/Editor/Utils.php
mitogh: Thanks love that @ggwicz

src/Tribe/Editor/Utils.php
mitogh: I think it might be helpful for an outsider that wants to execute an update on the `post_content` as the only params are: 

- `$post_id`
- `$pattern`
- `$replacement` (optional)

I think `public` makes sense as will provide externals just to call `tribe( 'editor' )->update_post_content( $post_id, $pattern )` to trigger an update on post content. 

Also as does not modifies any of the internals of the class I don't think making `public` is a bad thing as does not exposes any internal variables as it uses `wp_update_post` under the hood to do it 

Let me know what do you think about this.

PR #907, opened 2019-01-05
add registration prompt to localized data

paulmskim:
For ticket: https://central.tri.be/issues/119760

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -275,6 +275,7 @@ class Tribe__Main {
 				'today'           => esc_html__( 'Today', 'the-events-calendar' ),
 				'clear'           => esc_html__( 'Clear', 'the-events-calendar' ),
 			),
+			'registration_prompt' => __( 'There is unsaved attendee information. Are you sure you want to continue?', 'event-tickets' ),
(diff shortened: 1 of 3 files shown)


src/Tribe/Main.php
```diff
@@ -275,6 +275,7 @@ public function load_localize_data() {
 				'today'           => esc_html__( 'Today', 'the-events-calendar' ),
 				'clear'           => esc_html__( 'Clear', 'the-events-calendar' ),
 			),
+			'registration_prompt' => __( 'There is unsaved attendee information. Are you sure you want to continue?', 'event-tickets' ),
```
sc0ttkclark: Should be tribe-common and not event-tickets here

sc0ttkclark: Also, I feel like adding this into common needs to be revisited, it might be that we need a filter in common and ET will add `registration_prompt` to the list from it's own side.

paulmskim: There seem to be other text that is translated that use different text domains, not `tribe-common`, so I kept with that. But I totally agree we need to revisit this and see if we can separate these.

PR #918, opened 2019-01-16
#116843 - fix possible fatal, remove escaping on link

jesseeproductions:
_Ref:_ [C#116843](https://central.tri.be/issues/116843)

Fixes an issue if ET was the latest version and TEC an older, TEC would load, but ET does not and ET+ would load and generate a fatal. 

Also, it was escaping the links in the admin message.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$this' | [Plugin_Download.php:99](https://github.com/moderntribe/tribe-common/blob/e9a3571d531e038197d5811dca4321ac58bc842e/src/Tribe/Admin/Notice/Plugin_Download.php#L99)

PR #933, opened 2019-02-01
Prevent warnings when testing Regular Expressions

bordoni:
_Ref:_ [C#121655](http://central.tri.be/issues/121655)
---


src/functions/utils.php
```diff
@@ -518,7 +518,12 @@ function tribe_is_regex( $candidate ) {
 			return false;
 		}
 
-		return ! ( @preg_match( $candidate, null ) === false );
+		// We need to have the Try/Catch for Warnings too
+		try {
+			return ! ( @preg_match( $candidate, null ) === false );
```
juanfra: Maybe yoda conditional. Otherwise looking good.

juanfra:
I guess we also need to add a readme entry for this.

PR #974, opened 2019-04-04
Add utility back in order to enable filtering on endpoints

mitogh:
Function ported from `bucket/recurrence`, this one is used to filter the arguments used when an endpoint is registered.


sc0ttkclark:
@mitogh @bordoni Be sure to follow up on this and get it back for B19.05

PR #976, opened 2019-04-04
Bail if the page has no `$download_link` to avoid console errors

juanfra:
🎫 https://central.tri.be/issues/125042

Seems like with the new freemius implementation there are pages where we don't have the download link so we were trying to manipulate data that wasn't there, hence we were getting a console error.

### DIFF
--- a/src/resources/js/admin-log-controls.js
+++ b/src/resources/js/admin-log-controls.js
@@ -90,13 +90,12 @@ var tribe_logger_data  = tribe_logger_data || {};
 	 * currently selected log file.
 	 */
 	function update_download_link() {
-		var url = $download_link.attr( 'href' );
-
 		// bail if not in DOM
-		if ( typeof url === 'undefined' ) {
+		if ( 1 > $download_link.length ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/resources/js/admin-log-controls.js
```diff
@@ -91,6 +91,12 @@ var tribe_logger_data  = tribe_logger_data || {};
 	 */
 	function update_download_link() {
 		var url = $download_link.attr( 'href' );
+
+		// bail if not in DOM
```
sc0ttkclark: Should there be a `$download_link` check too, in case it's not there?

juanfra: Yeah, it can be done both ways. The href is more specific, but if there's no download link it can bail too.

juanfra:
@sc0ttkclark Thanks, ready for another 👀

src/resources/js/admin-log-controls.js
sc0ttkclark: My worry was mainly that .attr() would trigger a console error or something if $download_link wasn't found. Thanks!

PR #986, opened 2019-04-16
B19.04 Readme entries review

juanfra:
This PR contains the changelog entries review.


bordoni:
Those were coming from the Bucket work, I created those entries after the fact, so relating to a particular ticket was not possible.

PR #995, opened 2019-04-25
Add magic methods for more decorated property/method compatibility

sc0ttkclark:
FYI this issue was introduced by the TEC hotfix itself for B19.04.1 yesterday:

https://github.com/moderntribe/the-events-calendar/commit/92f82ef7f5c850d57fae9a5b16637ab1631b8644#diff-11b3f3a370885fc38177fe5c2798ce9fR412

### DIFF
--- a/src/Tribe/Repository/Decorator.php
+++ b/src/Tribe/Repository/Decorator.php
@@ -621,9 +621,9 @@ abstract class Tribe__Repository__Decorator implements Tribe__Repository__Interf
 	/**
 	 * Check if additional property on decorated object exists.
 	 *
-	 * @param string $name Property name.
-	 *
 	 * @since TBD
+	 *
+	 * @param string $name Property name.
 	 *
 	 * @return bool
 	 */
@@ -634,10 +634,10 @@ abstract class Tribe__Repository__Decorator implements Tribe__Repository__Interf
 	/**
 	 * Call methods on decorated object.
 	 *
+	 * @since TBD
+	 *
 	 * @param string $name      Method name.
 	 * @param array  $arguments Method arguments.
-	 *
-	 * @since TBD
 	 *
 	 * @return mixed
 	 */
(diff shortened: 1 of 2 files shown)


src/Tribe/Repository/Decorator.php
```diff
@@ -592,4 +592,56 @@ public function where_multi( array $fields, $compare, $value, $where_relation =
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $name      Method name.
+	 * @param array  $arguments Method arguments.
+	 *
+	 * @return mixed
+	 */
+	public function __call( $name, $arguments ) {
+		return call_user_func_array( [ $this->decorated, $name ], $arguments );
```
mitogh: I think an exception might be good if `this->decorated` is not `callable` before `call_user_func_array`

sc0ttkclark: I'm not sure I want to make that sort of decision at this point. It's not the duty of this class to add exceptions where none are expected, it will cause a fatal either way.

sc0ttkclark: I grabbed some of this code from ical-tec Events repository, it had implemented it on it's own apart from the Decorator in common. It didn't check if a property exists before trying to call it. I think it's important to remain neutral on whether the method/property exists or is public.

mitogh: I see, however if `$this->decorated` is not callable an error still will be trigger by `call_user_func_array`. I suggest an exception as a way to communicate more clear but if this is the current design on `common` probably just a check before the call will be enough, however when `decorated` is not callable will fail quietly

bordoni: Lets go with how @sc0ttkclark  did and in the future adjust.

mitogh: Yeah not a major stuff can be moved forward for now…

PR #1001, opened 2019-05-02
Feature/kitchen sink

paulmskim:
For ticket: https://central.tri.be/issues/126095

**Notes on using this branch before merge:**

- To get started, check out this branch
- Nuke node modules (`rm -rf node_modules`)
- In `package.json`, change `"product-taskmaster": "github:moderntribe/product-taskmaster"` to `"product-taskmaster": "github:moderntribe/product-taskmaster#bucket/views-redesign"` **(make sure not to commit this change)**
- **Optional:** In `package.json`, change `"dev": "cross-env NODE_ENV=development webpack -d --watch"` to `"dev": "cross-env NODE_ENV=development webpack -d --watch & gulp watch"` **(make sure not to commit this change)**
- Run `npm install`, then `npm run build`
- Visit kitchen sink page
- Base and component classes are defined in `src/resources/postcss/base` and `src/resources/postcss/components`, visit `https;//your.url/tribe/events/kitchen-sink`, compare with `src/views/kitchen-sink/page.php` and see how the classes are used with the markup.


paulmskim:
@ryanurban @faction23 curious on your take here, it looks like we don't have components like agency does (pass props to a component controller in php and render the template). As such, we're only looking at styles here. So semantically, from the `pcss` side, would it make sense to have a `components` folder? Should the buttons and search be moved into `base`, or somewhere else? What are your thoughts?

src/resources/css/reset.css
```diff
@@ -0,0 +1,370 @@
...
+	/* -----------------------------------------------------------------------------
+	*
+	* Normalize "Light"
+	*
+	* See normalize source code for clarity on rules
+	* https://github.com/necolas/normalize.css
+	*
+	* ----------------------------------------------------------------------------- */
+
+	.tribe-events html {
```
faction23: How does this reset and the body one work. For the body one to work tribe-events class would have to be on the html el, but for this one, i dont see how it would work?

src/resources/postcss/reset.pcss
```diff
@@ -0,0 +1,357 @@
...
+	/* -----------------------------------------------------------------------------
+	*
+	* Normalize "Light"
+	*
+	* See normalize source code for clarity on rules
+	* https://github.com/necolas/normalize.css
+	*
+	* ----------------------------------------------------------------------------- */
+
+	html {
```
faction23: Ya was looking at compiled but these two can go i imagine.

src/resources/postcss/reset.pcss
```diff
@@ -0,0 +1,357 @@
...
+		padding: 0;
+	}
+
+	/* -----------------------------------------------------------------------------
+	*
+	* Reset "Light"
+	*
+	* ----------------------------------------------------------------------------- */
+
+	html,
```
faction23: same here

faction23:
On the components question, before we had those on agency ya we would just put them on base. At the same time, it could get messy/long. Maybe bringing components dir here is ok for those types of well used elements. Thoughts @ryanurban ? Also, please review the forms css here, thanks!

src/resources/postcss/reset.pcss
paulmskim: hah yeah good catch, i just copied over the reset from sq1. but definitely wouldn't work the way the compiled selectors are set up.

faction23:
This is awesome and clean Paul, really liking it. I'll let the css heavyweight Ry chime in on some of the forms stuff and other aspects, but for me, bar that patch to the html/body this gtg!

src/resources/postcss/base/forms/_selects.pcss
```diff
@@ -0,0 +1,70 @@
...
+	 *     <li id="optionOne" role="option">Option One</li>
+	 *     <li id="optionTwo" role="option">Option Two</li>
+	 *   </ul>
+	 * </div>
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	 /**
+	  * @todo: implement listbox JS
+		* https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
```
ryanurban: Are you planning to use this lib and the corresponding markup above to power the filter bar view selector, @paulmskim?

src/resources/postcss/base/forms/_tabs.pcss
```diff
@@ -0,0 +1,117 @@
...
+	 *       <label id="tabTwoLabel" for="tabTwo" role="option">Tab Two</label>
+	 *     </li>
+	 *   </ul>
+	 * </div>
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	 /**
+	  * @todo: implement listbox JS
+		* https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
```
ryanurban: Assuming this isn't for powering tabs? Do you have an a11y tab solution in mind?

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -0,0 +1,33 @@
...
+	 *
+	 * Example:
+	 * <div class="tribe-common-form-control-text">
+	 *   <label for="textInput">Text Input</label>
+	 *   <input id="textInput" name="textInput" type="text" placeholder="Text Input" />
+	 * </div>
+	 *
+	 * ----------------------------------------------------------------------------- */
+
+	.tribe-common-form-control-text {
```
ryanurban: Curious, is this a special form input styling for text inputs only or should this be a bit more global in nature for use on fields that are more generic and will have this style?

src/resources/postcss/base/typography/_headings.pcss
```diff
@@ -0,0 +1,97 @@
+.tribe-events {
+	/* -------------------------------------------------------------------------
+	 * Heading: h1
+	 * ------------------------------------------------------------------------- */
+
+	.tribe-common-h1 {
+		color: var(--color-text-primary);
```
ryanurban: Wonder if it would be good to set these up as mixins so they could be easily reused as needed, thoughts?

src/resources/postcss/base/forms/_selects.pcss
paulmskim: Yes, I'm thinking a trimmed-down version. The version on w3 has a bunch of customizations (multi-select capability, etc) and I don't think we need that extra JS.

src/resources/postcss/base/forms/_tabs.pcss
paulmskim: The tabs are radio buttons with radios hidden, and labels as the tabs. When the tabs go to mobile, they will turn into a dropdown.

One thought I had here was to have 2 different markups, one for the tabs (using radios) on desktop, one for dropdown using the list as you saw in the dropdown example, and hiding the other in each media query case. If the radios + labels can also serve as dropdown list items, then we won't need the second markup, but if it doesn't work as well (especially from a11y perspective) then we may need to use a second markup (or another approach). Thoughts @ryanurban?

src/resources/postcss/base/typography/_headings.pcss
paulmskim: Will look, as I don't remember exactly how I set it up. I do remember trying that, but since it was only used once I decided against it. I thought that if we needed the `h1` styles we would just apply the class. But it couldn't hurt to have it as a mixin.

src/resources/postcss/base/forms/_text.pcss
paulmskim: Not used for radios, checkboxes, sliders, or toggles. More for inputs (text, number, email, etc). In the style guide we didn't have any other input types, so I went with text. I can change it to `-input` instead.

paulmskim:
Thanks Ryan. Responses to your notes:

- Will do on components, thanks for the feedback.
- I'm not sure I'm following the concern here. The way I was seeing it was any TEC block would be wrapped in `tribe-events`, and any TEC page would also be wrapped in `tribe-events`. Anything else (user-defined page templates) would not have the `tribe-events` class, and would not get the reset. Any other blocks on a blocks page would also not have the `tribe-events` class, so they would also not get the reset. Currently the reset would be part of the no style option and aid in the layout for the skeleton styles.
- I tried wrapping all the `@import` statements in `.tribe-events` but postcss-import did not like that and wouldn't import anything. postcss-partial-import works for `square-one` (to wrap an import statement in a class), but that's because it's on version 2. Version 4 does not allow that.

src/resources/postcss/base/forms/_tabs.pcss
ryanurban: Needing 2 different sets of markup to hide or show based on mobile or desktop feels...icky : ) So the UI changes based on mobile or desktop for this UI?

src/resources/postcss/base/forms/_text.pcss
ryanurban: Cool that makes sense, but yeah a more generic class name would be helpful 👍

src/resources/postcss/base/typography/_headings.pcss
ryanurban: Totally up to you and you can probably feel it out as you go 👍

ryanurban:
* For sure
* We should chat about this offline
* That's kind've a bummer

src/resources/postcss/base/forms/_tabs.pcss
paulmskim: Definitely icky. Gotta think a bit about how best to implement this combined tab/dropdown component.

PR #1005, opened 2019-05-07
Remove version number from dependency message for self.

lucatume:
Ticket: https://central.tri.be/issues/126615

This fix removes the version number from the message/link that will alllow users to download a plugin latest version when the plugin version is not above the minimum required one by the parent plugin.
I.e. TEC required PRO `>=4.7` but PRO is on version `4.6`.
The message, before, would show the **minimum** required version, not the latest one while the text suggests downloading the latest version.
This PR removes the version from the message entirely to simply redirect users to the latest version download.

Before: 
Screen Shot 2019-05-07 at 17 05 24

After:
Screen Shot 2019-05-07 at 17 12 32

### DIFF
--- a/src/Tribe/Dependency.php
+++ b/src/Tribe/Dependency.php
@@ -356,6 +356,15 @@ if ( ! class_exists( 'Tribe__Dependency' ) ) {
 					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


jesseeproductions:
Looks good to me.

PR #1011, opened 2019-05-09
VR > Add navigation elements to the kitchen sink

juanfra:
🎫 https://central.tri.be/issues/127499

Common classes to handle the global navigation elements. The markup will be the following:

```
<nav class="tribe-common-c-nav">
	<a href="#" rel="prev" class="tribe-common-c-nav--prev">Previous Events</a>
	<a href="#" rel="next" class="tribe-common-c-nav--next">Next Events</a>
</nav>
```


src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,93 @@
+/* = Navigation
+============================ */
+.tribe-events {
```
sc0ttkclark: Does this make it hard to reuse the classes in other plugins?

juanfra: We're actually using it like that already on the other components. We'll be probably moving out these into their own repo. I think events is the only one with views and maybe that's the reason why we ended up going that way https://github.com/moderntribe/tribe-common/blob/release/B19.06/src/resources/postcss/components/_search.pcss#L1

Maybe @bordoni or @paulmskim have another answer for this question/

sc0ttkclark: If it's prefixed with .tribe-events, it ought to be put into TEC styles?

juanfra: Everything in the kitchen sink is  prefixed with .tribe-events

For example:

- https://github.com/moderntribe/tribe-common/blob/release/B19.06/src/resources/postcss/base/typography/_body.pcss
- https://github.com/moderntribe/tribe-common/blob/release/B19.06/src/resources/postcss/base/typography/_ctas.pcss
- https://github.com/moderntribe/tribe-common/blob/release/B19.06/src/resources/postcss/base/typography/_headings.pcss

We're reseting many html things, and ideally we'll move these pcss files to their own repo that will be included on the other plugins -> https://github.com/moderntribe/tribe-common-styles

sc0ttkclark: I just think it's odd anything plugin-specific should end up in common/common-styles. If everyone is OK with this then that's fine for me, just want to bring it up :)

juanfra: TEC is the only one that has views. We'll eventually have a separate repo for this (the one I linked above) to store some variables, mixins, etc. This is a WIP, so until we have the separate repo working as a module of the plugins, we're having this in common. 

Let's wait for @bordoni or @paulmskim to chime in.

paulmskim: All the styles (resets, components, base, etc.) will have a wrapper of `.tribe-events`. Though we define our own classes (`.tribe-common-h1`, for example), there is the small chance we clash and override a customer's styles. We don't want that. Also in the reset, we don't want to reset customer elements. So we wrap all styles in the wrapper class. Each component, block, page, etc. that is defined by TEC will have this setup. If we extend this views work to tickets, for example, we will also add `.tribe-tickets` wrapper class.

paulmskim:
@juanfra let's make the markup for the nav a `<ul>`:

```
<nav>
    <ul>
        <li>
            <a href="#">Press me</a>
        </li>
        ...
    </ul>
</nav>
```

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,93 @@
...
+
+	.tribe-common-c-nav {
+		display: flex;
+		justify-content: space-between;
+		border-top: 1px solid var(--color-border-default);
+		padding-top: var(--spacer-6);
+	}
+
+	.tribe-common-c-nav--prev,
```
paulmskim: These should be `.tribe-common-c-nav__prev` and `.tribe-common-c-nav__next`

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,93 @@
...
+			&:before {
+				background-image: svg-inline(caret-left-disabled);
+			}
+		}
+
+	}
+
+	.tribe-common-c-nav--next {
+
+		&:after {
```
paulmskim: should be `&:before`

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,93 @@
...
+		color: var(--color-text-secondary);
+		font-size: var(--font-size-1);
+		font-weight: var(--font-weight-bold);
+
+		@media (--viewport-medium) {
+			font-size: var(--font-size-2)
+		}
+
+		&:hover {
+			color: rgba(var(--color-text-primary), 0.8);
```
paulmskim: Let's add this to the `color` variables file

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,93 @@
...
+	.tribe-common-c-nav--prev {
+
+		&:before {
+			background-image: svg-inline(caret-left);
+			background-repeat: no-repeat;
+			background-size: contain;
+			content: '';
+			height: 14px;
+			margin-right: 10px;
+			width: 12px;
```
paulmskim: let's move `background-repeat`, `background-size`, `content`, `height`, and `width` up to the block with both `prev` and `next` (keep things DRY)

src/resources/postcss/components/_navigation.pcss
paulmskim: @juanfra let's change this to `.tribe-common` as per our earlier discussion

juanfra:
Ready for a new pass

src/resources/postcss/components/_navigation.pcss
sc0ttkclark:
```suggestion
		&:before {
```

sc0ttkclark: @paulmskim Is this resolved by recent commits?

paulmskim: @juanfra this still needs to be fixed here.

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,96 @@
+/* = Navigation
+============================ */
+.tribe-common {
+
+	.tribe-common-c-nav {
+		border-top: 1px solid var(--color-border-default);
+		padding-top: var(--spacer-6);
+
+		ul {
```
paulmskim: could we add a class to the nav `ul` and `li` elements, such as `.tribe-common-c-nav__list` and `.tribe-common-c-nav__list-item` and use them as selectors rather than nesting `ul` and `li` under classes? this way we can keep the specificity of each element at a minimum.

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,96 @@
...
+			background-repeat: no-repeat;
+			background-size: contain;
+			content: '';
+			height: 14px;
+			width: 12px;
+		}
+	}
+
+	.tribe-common-c-nav__prev {
+		float: left;
```
paulmskim: Let's not use floats here, otherwise the parent elements of `.tribe-common-c-nav__prev` and `next` now have no children, and their sizes collapse, resulting in a navigation that has 0 height and 0 width. This makes it difficult to build markup and style around.

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,96 @@
...
+
+			&:before {
+				background-image: svg-inline(caret-left-disabled);
+			}
+		}
+
+	}
+
+	.tribe-common-c-nav__next {
+		float: right;
```
paulmskim: Let's not use floats here, otherwise the parent elements of `.tribe-common-c-nav__prev` and `next` now have no children, and their sizes collapse, resulting in a navigation that has 0 height and 0 width. This makes it difficult to build markup and style around.

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,96 @@
...
+		ul {
+			display: block;
+			list-style: none;
+			width: 100%;
+		}
+	}
+
+	.tribe-common-c-nav__prev,
+	.tribe-common-c-nav__next {
+		display: inline-flex;
```
paulmskim: Let's also try to avoid `inline-flex` unless these elements absolutely need to be `inline`. Inline elements add a space below the element for text descenders that cannot be removed, and it makes spacing impossible to get perfect.

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,96 @@
...
+		align-items: center;
+		color: var(--color-text-secondary);
+		font-size: var(--font-size-1);
```
paulmskim: We're going to need `line-height` style for almost all text styling properties so that we properly set the size of the text.

src/resources/postcss/components/_navigation.pcss
juanfra: I'm using :before for the prev and :after for the next arrow. Is there any reason for them to live both in the same pseudo selector?

paulmskim: Oh I understand, no that's fine. I know I said to move the common style to https://github.com/moderntribe/tribe-common/pull/1011/files#diff-e1567768e26c49c75604027039283f1fR38, but let's move them back to https://github.com/moderntribe/tribe-common/pull/1011/files#diff-e1567768e26c49c75604027039283f1fR50 and https://github.com/moderntribe/tribe-common/pull/1011/files#diff-e1567768e26c49c75604027039283f1fR76, the reason being there are unused styles in `.tribe-common-c-nav__prev:after` and `.tribe-common-c-nav__next:before` in the common styles.

src/resources/postcss/components/_navigation.pcss
paulmskim: I think you're still missing this @juanfra , looks like the last one.

paulmskim: I think you're still missing this last one @juanfra

juanfra:
@paulmskim would you check this again please?

src/resources/postcss/components/_navigation.pcss
```diff
@@ -0,0 +1,106 @@
...
+
+	.tribe-common-c-nav__prev,
+	.tribe-common-c-nav__next {
+		display: flex;
+		justify-content: center;
+		flex-wrap: wrap;
+		align-items: center;
+		color: var(--color-text-secondary);
+		font-family: var(--font-family-base);
+		font-size: var(--font-size-1);
```
paulmskim: I came across this in a pass I did, but any place where we're manually defining `font-size` and font-related styles, we should actually not be doing this. On the `tribe-common-c-nav__prev` and `tribe-common-c-nav__next` classes we should be applying, say, `.tribe-common-b1` classes to them. Let's take care of this in another pass.

PR #1019, opened 2019-05-16
VR >  Add theme compatibility for the views elements

juanfra:
🎫 https://central.tri.be/127066


src/resources/postcss/base/forms/_tabs.pcss
```diff
@@ -68,18 +69,36 @@
 			&:first-child label {
 				@mixin form-control-select-list-item-first;
 
+				body#top & { /* Enfold compatibility */
```
paulmskim: let's move all these overrides to the bottom in a new wrapper, like so:

```css
/* Enfold compatibility */
body#top & .tribe-common-form-control-tabs {
    li {
        &:first-child label {
            /* style overrides */
        }
        ...
    }
    ...
}
```

This way we can keep all the override styles in one place rather than having them spread out in the file. In the final CSS file they'll have their own selectors anyway so it won't create a larger CSS file.

src/resources/postcss/base/forms/_tabs.pcss
```diff
@@ -68,18 +69,36 @@
...
+					@mixin form-control-select-list-item-first;
```
paulmskim: Are all the mixin parameters overridden by enfold, or some of them? If we can help it, we should only target those affected.

src/resources/postcss/base/forms/_tabs.pcss
```diff
@@ -68,18 +69,36 @@
...
+				}
+
 				@media (--viewport-medium) {
 					margin-top: -6px;
 					padding-top: 17px;
+
+					body#top & { /* Enfold compatibility */
```
paulmskim: Let's keep media query overrides within the same selector, so for example:

```css
body#top & {
    @mixin form-control-select-list-item-first;

    @media (--viewport-medium) {
        margin-top: -6px;
        padding-top: 17px;
    }
}
```

This keeps all styles packaged together rather than spread apart in media queries.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,9 +24,22 @@
 			padding: 12px 18px;
 			width: 100%;
 
+			body#top & { /* Enfold compatibility */
+				@mixin desktop-body-2-bold;
```
paulmskim: same question here, are all styles affected or can we target only those affected?

src/resources/postcss/resets/_enfold.pcss
```diff
@@ -0,0 +1,17 @@
+.tribe-theme-enfold {
+
+	/* Enfold is using the #top id in the <body> */
+	#top& {
+		.tribe-common {
+			input,
+			button,
+			select,
+			textarea {
```
paulmskim: just checking that it's all 4 of these selectors that are affected, or if we can narrow it down to only the ones that are affected.

src/resources/postcss/base/forms/_tabs.pcss
juanfra: I did it that way because that's what it said in the doc.

paulmskim: You mean the CSS architecture document, right? Yeah, I know, but template overrides are a bit different. Thinking of maintenance, it's going to be quite difficult to maintain style overrides if they're interspersed all over a specific component/element's styles rather than all in one place. And this won't just be for one theme, this is for all themes, for all components.

I'll update the CSS architecture document to reflect this.

src/resources/postcss/base/forms/_tabs.pcss
juanfra: Yes, those mixins are only paddings. I can set the properties there directly in case the mixins grow in the future.

src/resources/postcss/base/forms/_text.pcss
juanfra: Specifically for this one, yes. They're setting styles for the input font, font weight and font color, padding and borders.

paulmskim: great, sounds good

src/resources/postcss/base/forms/_toggles.pcss
```diff
@@ -66,4 +67,41 @@
 			margin: 0 0 6px;
 		}
 	}
+
+	body#top & { /* Enfold Compatibility */
+
+		.tribe-common-form-control-toggle {
+
+			input[type="checkbox"] {
+				@mixin form-control-slider-toggle-track;
```
paulmskim: same question on whether we need all these mixins / styles or if it's just a certain number of them.

juanfra: only one! Ready

paulmskim: Also the `&:after` and the `label` below this.

juanfra: Done @paulmskim

src/resources/postcss/base/forms/_tabs.pcss
paulmskim: ah okay, if they are all affected then no worries, just wanted to check.

PR #1021, opened 2019-05-20
Fix/128031 kitchen sink refinements 2

paulmskim:
For ticket: https://central.tri.be/issues/128031


src/resources/postcss/base/forms/_sliders.pcss
```diff
@@ -5,90 +5,103 @@
...
+		width: 120px;
+		vertical-align: middle;
 
-			/* -----------------------------------------------------------------------------
-			 * Track styles
-			 * ----------------------------------------------------------------------------- */
-			&::-webkit-slider-runnable-track {
-				@mixin form-control-slider-track;
-			}
+		/* -----------------------------------------------------------------------------
```
ryanurban: All of these nested comment blocks look off on their tabbing for the 2nd & 3rd lines?

src/resources/postcss/base/forms/_tabs.pcss
```diff
@@ -31,87 +65,87 @@
...
 		}
+	}
 
-		li {
+	.tribe-common-form-control-tabs__list-item {
 
-			&:first-child label {
-				@mixin form-control-select-list-item-first;
+		&:first-child label {
+			@mixin form-control-select-list-item-first;
```
ryanurban: Noticing in general there are a lot of mixins for what potentially feels like one-off, non-reused styles around specific selectors. What and how are you deciding what to optimize for in how you're going about your use of mixins?

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -5,29 +5,34 @@
...
-		label {
-			@mixin visually-hide;
-		}
+	.tribe-common-form-control-text__label {
+		@mixin visually-hide;
+	}
 
-		input[type="text"] {
-			@mixin desktop-body-2-bold;
+	.tribe-common-form-control-text__input {
```
ryanurban: Are we handling focus styles for form inputs?

paulmskim: looks like not, will check with matt

src/resources/postcss/base/forms/_tabs.pcss
paulmskim: yeah, not sure what i was thinking with this (the mixin is just one style), will remove.

src/resources/postcss/base/forms/_sliders.pcss
paulmskim: yup, fixed!

src/resources/postcss/base/forms/_text.pcss
paulmskim: still in flux he says, will update once it comes out ;)

src/resources/postcss/base/forms/_tabs.pcss
ryanurban: Fwiw, this question isn't specific to just this instance of a mixin, curious about quite a few of the ones I see.

paulmskim: Yup, I'll have another pass at them.

paulmskim: So had a quick pass, there are a couple mixins that are only used once (a couple of the `-desktop` variations), which I'm sure could just be put into the class itself. But I figured when I did this that keeping all the styles in one place would be good, even if it's used only once.

PR #1031, opened 2019-05-27
Views v2 - navigation support code

lucatume:
This PR puts in place the code to support the production of current, next and prev links for the new View system.

Tickets:
* https://central.tri.be/issues/127983
* https://central.tri.be/issues/127981
* https://central.tri.be/issues/127980


src/Tribe/Context.php
```diff
@@ -1215,4 +1215,72 @@ public function filter( array $filters, $default ) {
...
+				foreach ( (array) $name as $destination ) {
+					$mapped[ $destination ] = $input[ $key ];
+				}
+			}
+		}
+
+		if ( $passthru ) {
+			$mapped = array_merge(
+				$mapped,
+				array_diff_key( $input, array_keys( $locations ), array_combine( $processed, $processed ) )
```
bordoni: Double `$processed` here on `array_combine( $processed, $processed )`? Not sure if correct.

src/Tribe/Template.php
```diff
@@ -613,4 +613,55 @@ public function template( $name, $context = array(), $echo = true ) {
...
+
+	/**
+	 * Returns the Template local context.
+	 *
+	 * @since TBD
+	 *
+	 * @return array An associative key/value array of the Template local context.
+	 */
+	public function get_local_values() {
+		return $this->context;
```
bordoni: Filter local values, and maybe `get_template_local_values()`, not sure if we need it.

src/Tribe/Template.php
```diff
@@ -613,4 +613,55 @@ public function template( $name, $context = array(), $echo = true ) {
...
+
+	/**
+	 * Returns the Template global context.
+	 *
+	 * @since TBD
+	 *
+	 * @return array An associative key/value array of the Template global context.
+	 */
+	public function get_global_values() {
+		return $this->global;
```
bordoni: Filter global values, and maybe `get_template_global_values()`, not sure if we need it.

src/Tribe/Template.php
```diff
@@ -613,4 +613,55 @@ public function template( $name, $context = array(), $echo = true ) {
...
+	/**
+	 * Returns the Template global and local context values.
+	 *
+	 * Local values will override the template global context values.
+	 *
+	 * @since TBD
+	 *
+	 * @return array An associative key/value array of the Template global and local context.
+	 */
+	public function get_values() {
```
bordoni: Filter values, and maybe `get_template_values()`, not sure if we need it.

bordoni: We need to make sure we are using this method on the `$this->get()`

src/Tribe/Rewrite.php
```diff
@@ -393,19 +426,22 @@ public function get_canonical_url( $url, $force = false ) {
...
+		}
+
 		if ( $canonical_url !== $resolved ) {
 			$resolved = trailingslashit( $resolved );
 		}
 
 		if ( count( $unmatched_vars ) ) {
-			$resolved = add_query_arg( $unmatched_vars, $canonical_url );
+			$resolved = add_query_arg( $unmatched_vars, $resolved );
 		}
```
bordoni: We should add a filter for the canonical URL.

src/Tribe/Template.php
```diff
@@ -613,4 +613,55 @@ public function template( $name, $context = array(), $echo = true ) {
 
 		return $html;
 	}
+
+	/**
+	 * Sets a number of values at the same time.
```
bordoni: Incorrect indentation. Around this docblock.

src/Tribe/Context.php
```diff
@@ -1215,4 +1215,72 @@ public function filter( array $filters, $default ) {
 
 				return $default;
 		}
+
+	/**
+	 * Maps an input array to the corresponding read locations.
+	 *
+	 * The resulting array can be used as input for the `alter_values` method.
```
bordoni: Weird indentation around this docblock.

src/Tribe/Context.php
lucatume: It is: since we have to diff the array by keys I need to have an array that has `$processed` as keys; I could either use `array_combine($processed, array_fill(0, count($processed), true))` or just do what I did and avoid wasted computation.

src/Tribe/Template.php
lucatume: Since this is a method on the template object calling it `get_template_local_values()` seem redundant:
```php
$template->get_template_local_values(); // redundant
$template->get_local_values(); // cleaner
```

Is there value in filtering the values? Here we're getting the values that have been set on the template and then we read them. What value filtering the values would provide? The Template is not going to use the filtered version, but the unfiltetered one.

src/Tribe/Template.php
lucatume: Same as above.

lucatume: I've modified this method to fetch local values using `get_local_values` and to fetch global values using `get_global_values`.
Since we're applying the `tribe_template_context_get` filter right after that I feel we've got enough filtering.

src/Tribe/Rewrite.php
lucatume: Right, thanks.

lucatume:
I've applied the changes requested by @bordoni  and improved this code (but tested in [TEC PR](https://github.com/moderntribe/the-events-calendar/pull/2527) due to its own nature) to handle localized slugs.

src/Tribe/Template.php
lucatume: Same as above.

PR #1032, opened 2019-05-29
Fix/update icons

paulmskim:


src/resources/postcss/base/_svg-icons.pcss
```diff
@@ -10,13 +10,13 @@
 
 	.tribe-common-svgicon--featured {
 		background-image: svg-inline(featured);
-		height: 12px;
-		width: 12px;
+		height: 10px;
```
juanfra: Why are we changing the sizes? For the multiday for example I just implemented the icon and the inspector said it was 12x12.

paulmskim: So the iconography section in the list view elements shows the featured icon as 10x12 and the recurring as 10x10, but every single implementation of it is 8x10 and 10x10, respectively, even in the multiday component.

PR #1040, opened 2019-06-05
update navigation css, remove padding from button

paulmskim:


juanfra:
Awesome stuff. Please merge 19.07 in it and make sure there are not problems with the tests (not that we're introducing code that's failing, because it's pcss, but to be sure we don't have problems with tests)

PR #1042, opened 2019-06-06
fix(Context): remove import of Guzzle `parse_query` function from code

lucatume:


juanfra:
Not sure if we need to switch the base branch as B19.08 cause 19.07 is going out now

PR #1046, opened 2019-06-11
Set the default link color, to prevent theme overrides.

juanfra:
As we talked with @paulmskim on the Month view PR, the idea would be to set the default link color on the common styles directly.

### DIFF
--- a/src/resources/postcss/base/typography/_anchors.pcss
+++ b/src/resources/postcss/base/typography/_anchors.pcss
@@ -1,7 +1,7 @@
 .tribe-common {
 
 	a {
-		color: color: var(--color-text-primary);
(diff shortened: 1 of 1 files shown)


src/resources/postcss/base/typography/_anchors.pcss
```diff
@@ -1,6 +1,7 @@
 .tribe-common {
 
 	a {
+		color: color: var(--color-text-primary);
```
sc0ttkclark:
```suggestion
		color: var(--color-text-primary);
```

juanfra:
Dang! copy and paste. Thanks Scott. No changelog required

PR #1048, opened 2019-06-12
feat(Context.php): add the `duplicate` method

lucatume:


src/Tribe/Context.php
```diff
@@ -1416,4 +1416,22 @@ public function map_to_read( array $input, $types = null, $passthru = false ) {
 
 		return $mapped;
 	}
+
+	public function duplicate( array $values, $type, $direction = 'read' ) {
```
bordoni: Docblock.

PR #1049, opened 2019-06-12
[VR] Add link color for different states

juanfra:
🎫 https://central.tri.be/issues/129307


juanfra:
No changelog entry required.

PR #1053, opened 2019-06-17
VR > Add new icons

juanfra:
🎫 https://central.tri.be/issues/129296

Add new icons that will be used on the new events bar.


bordoni:
Why are these icons here instead of on TEC?

juanfra:
@bordoni so, they live in both places. The svgs are normally included in the mixins that live in the `tribe-common-styles`. As that's part of both repos, then the icons should be on both. Please check: https://github.com/moderntribe/tribe-common-styles/pull/16

bordoni:
I really feel like these icons should live inside either The Events Calendar or `tribe-common-styles`. there isn't much sense on been on Common, since their dependency is on the other repo.

juanfra:
@paulmskim thoughts? ☝️

paulmskim:
The way we currently have our repository set up, our icons live in `src/resources/images/icons` in both common and TEC. Both common and TEC will use these icons used in `src/resources/postcss/utilities/variables/_svg.pcss`. I have no problems with this living in `tribe-common-styles`, but we will need to create a folder for these.

juanfra:
1. Cool, didn't know we were minifying those with that tool. 
2. Yes, we should definitely try to define that sooner rather than later.

bordoni:
We ended up with a decision to move these to `tribe-common-styles`.

PR #1057, opened 2019-06-18
VR > List view refinements

juanfra:
🎫 https://central.tri.be/issues/129307

- Adding modifier for the list (and day) view titles.
- Adding modifier for the month separator and day marker.


src/resources/postcss/base/typography/_headings.pcss
```diff
@@ -77,6 +77,12 @@
 		@mixin heading-5;
 	}
 
+	.tribe-common-h5--min-medium {
```
paulmskim: Let's move this to the bottom, below `h8`. If we end up defining other `--min-medium` classes and we apply it to a heading size that already has a `@media`, they won't be applied because of the order of this document.

paulmskim: let's add `h6--min-medium` too for https://github.com/moderntribe/the-events-calendar/pull/2560#discussion_r295105572

PR #1058, opened 2019-06-19
add alt for h8

paulmskim:


juanfra:
The css changes look good! I'm wandering why are you submitting also the minified version of the `tribe-common.js` - Let's try to remove that, because we're not involving QA on these changes and a change on the JS can bring new possible problems.

paulmskim:
good catch, must've been a bad `git add`. will create a new PR.

PR #1061, opened 2019-06-20
VR > Events Bar > Form Inputs

juanfra:
🎫 https://central.tri.be/issues/129299

Re-style form elements according to new design.


src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
 
 	.tribe-common-form-control-text__input {
-		@mixin desktop-body-2-bold;
-
-		border: 1px solid var(--color-border-default);
-		padding: 12px 18px;
+		@mixin desktop-body-1;
+		border: none;
+		border-bottom: 1px solid var(--color-border-default);
+		padding: 12px 20px 12px 0;
```
paulmskim: let's use `--spacer` variables for these?

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
 		width: 100%;
 
+		@media (--viewport-medium) {
+			@mixin desktop-body-2;
+
+			border: none;
+			padding: 12px 18px;
```
paulmskim: let's use `--spacer` for these. also, i think these are different now, confirm with the desktop designs.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
+		}
+
 		&::placeholder {
 			color: var(--color-text-secondary);
 			font-weight: var(--font-weight-regular);
```
paulmskim: let's remove this, since we aren't using a bold font anymore.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
 	body#top & { /* Enfold Compatibility */
 
 		.tribe-common-form-control-text__input {
-			@mixin desktop-body-2-bold;
+			@mixin desktop-body-1;
 
-			border: 1px solid var(--color-border-default);
-			padding: 12px 18px;
+			border-bottom: 1px solid var(--color-border-default);
+			padding: 12px 20px 12px 0;
```
paulmskim: let's use `--spacer` for these

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
 			width: 100%;
 
+			@media (--viewport-medium) {
+				@mixin desktop-body-2;
+
+				padding: 12px 18px;
```
paulmskim: let's use `--spacer` for these. also, i think these are different now, confirm with the desktop designs.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,27 +24,44 @@
...
+			}
+
 			&::placeholder {
 				color: var(--color-text-secondary);
 				font-weight: var(--font-weight-regular);
```
paulmskim: no need for this anymore

src/resources/postcss/components/_search.pcss
```diff
@@ -48,18 +48,19 @@
 	.tribe-common-c-search {
 		@media (--viewport-medium) {
 			display: flex;
+			flex: 1 1 100%;
```
paulmskim: has the markup changed such that this is a flex child? curious why we're adding this here.

src/resources/postcss/components/_search.pcss
```diff
@@ -48,18 +48,19 @@
...
 		@media (--viewport-medium) {
 			background-color: var(--color-background);
-			border: 1px solid var(--color-border-default);
 			border-right: none;
 			display: flex;
 			flex: auto;
-			max-width: 645px;
 			padding: 8px 0;
+			align-items: center;
+			justify-content: center;
```
paulmskim: curious why you decided to center justify the input group contents

src/resources/postcss/components/_search.pcss
```diff
@@ -68,26 +69,82 @@
...
 			height: auto;
 			padding: 7px 18px 8px;
 		}
+
+		&:focus {
+			outline: none;
+		}
+	}
+
+	.tribe-common-c-search__input--keyword {
```
paulmskim: we should probably put the `--keyword` modifier around the `<div>` wrapper of the label and input.

src/resources/postcss/components/_search.pcss
```diff
@@ -68,26 +69,82 @@
...
 	.tribe-common-c-search__input {
+
 		@media (--viewport-medium) {
-			border: none;
 			border-right: 1px solid var(--color-border-default);
 			height: auto;
 			padding: 7px 18px 8px;
 		}
+
+		&:focus {
```
paulmskim: this probably belongs to a more global `input` style rather than specific to search. let's move this to `src/resources/postcss/base/forms/_text.pcss`

src/resources/postcss/components/_search.pcss
```diff
@@ -68,26 +69,82 @@
...
+	}
+
+	.tribe-common-c-search__input--keyword {
+		background-image: svg-inline(search);
+		background-repeat: no-repeat;
+		background-size: 16px;
+		background-position: right center;
+
+		@media (--viewport-medium) {
+			border: none;
```
paulmskim: shouldn't need this.

src/resources/postcss/components/_search.pcss
```diff
@@ -68,26 +69,82 @@
...
+			border-right: 1px solid var(--color-border-default);
+			height: auto;
+			padding: 7px 18px 7px 40px;
```
paulmskim: We should not be overriding padding styles from form control styles, otherwise there's no reason for us to have them. I see that this is to get the `border-right` style correct here. I'd do a few things here:

1. set that `40px` padding in the form control styles so default styles are applied with that padding
2. move the `border-right` style to the wrapper div (or even better, move it to `.tribe-common-c-search__input-group > *`)
3. to get the correct border height, apply a negative margin on top and bottom on `.tribe-common-c-search__input`.

This way, we get the full height for the input, but the border is still the smaller height as per the design.

src/resources/postcss/components/_search.pcss
```diff
@@ -68,26 +69,82 @@
...
+		}
+	}
+
+	.tribe-common-c-search__input--keyword-focus {
+		background-image: svg-inline(search-focus);
+	}
+
+	.tribe-common-c-search__input--location {
+		background-image: svg-inline(location);
+		background-repeat: no-repeat;
```
paulmskim: lets consolidate some of these common repeated styles for `.tribe-common-c-search__input`

src/resources/postcss/components/_search.pcss
```diff
@@ -112,7 +169,14 @@
...
 				border-right: 1px solid var(--color-border-default);
-				padding: 7px 18px 8px;
+			}
+		}
+	}
+
+	body.tribe-theme-twentynineteen & {
+		.tribe-common-c-search__input {
+			&:focus {
+				border-right-color: var(--color-border-default);
```
paulmskim: this border color shouldn't be an issue if we move the border to the div wrapper

src/resources/postcss/components/buttons/_solid.pcss
```diff
@@ -13,11 +13,12 @@
 		@mixin button-solid;
 
 		background-color: var(--color-accent-primary);
-		padding: 16px 22px 17px;
+		padding: 11px 22px 11px;
 		width: 100%;
 
 		@media (--viewport-medium) {
 			width: auto;
+			padding: 16px 22px 17px;
```
paulmskim: check this padding again, i'm not sure this matches the design

src/resources/postcss/base/forms/_text.pcss
paulmskim: we also have new spacer variables, so be sure to pull the latest `tribe-common-styles`.

src/resources/postcss/components/_search.pcss
juanfra: Snap, was testing something and it ended up there.

src/resources/postcss/components/_search.pcss
juanfra: Actually, the padding of these overrides are to include the icons on the different inputs.

src/resources/postcss/components/_search.pcss
juanfra: These are only for the inputs that have icons. The idea is that `.tribe-common-c-search__input` is the base class for inputs, and this modifiers are specifically for the inputs that came from design (search and location). If somebody (or us as an extension) wants to add a new input, they will use `.tribe-common-c-search__input` and will have the correct input styles there, without icons.

src/resources/postcss/components/buttons/_solid.pcss
juanfra: From what I checked it's still matching.

src/resources/postcss/components/_search.pcss
juanfra: The idea of this modifier is to actually make the input styles different, and not the label or anything else. Why do you think we should move it to the wrapper?

src/resources/postcss/base/forms/_text.pcss
paulmskim: @juanfra were you able to confirm with the designs? I think matt is using `--spacer-4` for top and bottom.

src/resources/postcss/components/_search.pcss
paulmskim: The same could be said about the list items and the links in the select/tabs for the view selector. If we want to allow this to be extendable, then we should stay consistent in our approach.

src/resources/postcss/components/_search.pcss
paulmskim: Right, so the padding values in `.tribe-common-form-control-text__input`, I don't see them in the design anywhere. I believe if we only have the event bar styles given to us for input styles, we should use that as the default. We can always come back and shuffle around what's common styles and what's specific to events bar when a new input style comes, but until then, having multiple different padding values is unnecessary CSS and can be confusing when we do try to make an update.

src/resources/postcss/components/_search.pcss
paulmskim: same note as https://github.com/moderntribe/tribe-common/pull/1061#discussion_r296099821. I'm also OK with a `.tribe-common-c-search__input--icon` modifier with common styles, but same with the note above, we should keep as much common as possible until a new design forces us to split off specifics.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,30 +24,46 @@
...
 
 	.tribe-common-form-control-text__input {
-		@mixin desktop-body-2-bold;
-
-		border: 1px solid var(--color-border-default);
-		padding: 12px 18px;
+		@mixin desktop-body-1;
+		border: none;
+		border-bottom: 1px solid var(--color-border-default);
+		padding: var(--spacer-2) 20px var(--spacer-2) 0;
```
paulmskim: we have a 20px spacer, let's use that.

src/resources/postcss/base/forms/_text.pcss
```diff
@@ -24,30 +24,46 @@
...
 	body#top & { /* Enfold Compatibility */
 
 		.tribe-common-form-control-text__input {
-			@mixin desktop-body-2-bold;
+			@mixin desktop-body-1;
 
-			border: 1px solid var(--color-border-default);
-			padding: 12px 18px;
+			border-bottom: 1px solid var(--color-border-default);
+			padding: var(--spacer-2) 20px var(--spacer-2) 0;
```
paulmskim: we have a 20px spacer, let's use that

src/resources/postcss/components/_search.pcss
```diff
@@ -48,18 +48,18 @@
 	.tribe-common-c-search {
 		@media (--viewport-medium) {
 			display: flex;
+			width: 100%;
```
paulmskim: My main question around this is that `.tribe-common-c-search` should be a standalone component, and all styles should be contained within the `.tribe-common-c-search` component. How this component is used and fit into the flow of the document shouldn't be defined in the component. Rather, we can define this in `.tribe-events-c-events-bar__search` (or another class, but the `<form>` element) to define how the search component fits into the flow of the events bar. Not likely, but if we end up using this component somewhere else, the width and how this component flows within that markup may be different.

src/resources/postcss/components/buttons/_solid.pcss
paulmskim: So body-2 text line height is 23px, padding top is 16px, and padding bottom is 17px, total height I'm getting is 56px. In the designs I'm seeing a height of 45px.

src/resources/postcss/components/_search.pcss
juanfra: Right, the padding is not anymore in the design anymore because the two input fields we have, are now using icons. I thought keeping the default values for the default input, and adding modifiers, would allow us to add a new field without any trouble, and that field would just look ok and without any strange padding.

It's like -> Default field has a default padding which is OK for text. Modifiers have their own padding because they have an icon and they need space. If not the text will be on top of the icon. If you want I can make a modifier for icon. I feel like that would be OK, but at the same time it'd be better to keep that on the input rather than the wrapper.

src/resources/postcss/base/forms/_text.pcss
juanfra: He's using `--spacer-4` but in this way: https://cloudup.com/cwJOV8799UJ - That's not replicable on HTML, and we'll have to find a way to make it look like that without having 20px padding how he demonstrated on the sketch.

src/resources/postcss/components/_search.pcss
juanfra: I feel this case is different, because these are inputs, and the styles here are purely related to the input. The wrapper is really generic on this one. The list items of navigations are purely related to the action they have inside.

juanfra: But I get you, with the shoutout for consistency, I agree we should make the effort to be as consistent as we can.

src/resources/postcss/components/_search.pcss
juanfra: Ok, so you would set the width directly on the first child of `.tribe-events-c-events-bar__form` and not this class directly.

paulmskim: Yes, whatever the `<form>` element class is in `.tribe-events-c-events-bar`. I don't know what the class is off hand, but if there isn't one, define one next to `.tribe-common-c-search`.

src/resources/postcss/base/forms/_text.pcss
paulmskim: Let's chat about this, we can replicate this on HTML.

src/resources/postcss/components/_search.pcss
paulmskim: @bordoni want to weigh in here? i know you had a preference about the `<li>` case last time.

src/resources/postcss/components/_search.pcss
paulmskim: I hear you, let's chat about this and figure out the way forward.

src/resources/postcss/components/_search.pcss
bordoni: I do like it to be consistent with what we did on the other PRs where we use the class on the wrapper element.

juanfra: So, this would end up being like this: 

➡️  having a the wrapper div with the additional class, for example for the keyword: `tribe-common-form-control-text__keyword` 

- Applying the focus modifier class on the wrapper and leaving the kind of BEM format we've been following, because we'll have to target the input inside the wrapper. E.g.:

`
.tribe-common-form-control-text__keyword .tribe-common-c-search__input { /* whatever */ }
.tribe-common-form-control-text__keyword--focus .tribe-common-c-search__input { /* whatever focus state */ }
`

At some point I feel it's not consistent with the rest, independently from what we have done on the navigation, where the `<li>` was the "owner" of the action inside (a link).

paulmskim: The `tribe-common-form-control-text` classes are common classes, let's not add modifiers to these as these should not know about whether there's a keyword modifier or location modifier. Instead, let's add a `tribe-common-c-search__input-control` or `__input-wrapper` class to the div that has the `tribe-common-form-control-text` class. That way, we're working within the search component context and can add `tribe-common-c-search__input-control--keyword` class.

PR #1065, opened 2019-06-21
Fix/assets improve code and docblocks

cliffordp:
https://central.tri.be/issues/129061

Dug into assets stuff while working this ticket, although this PR doesn't actually address the ticket's issue(s). Just thought it best to separate this stuff out for easier review since there are a lot of little changes.

### DIFF
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -295,45 +295,50 @@ if ( ! class_exists( 'Tribe__Settings' ) ) {
 		 * @return void
 		 */
 		public function initTabs() {
-			if ( isset( $_GET['page'] ) && $_GET['page'] == $this->adminSlug ) {
-				// Load settings tab-specific helpers and enhancements
-				Tribe__Admin__Live_Date_Preview::instance();
-
-				do_action( 'tribe_settings_do_tabs' ); // this is the hook to use to add new tabs
-				$this->tabs       = (array) apply_filters( 'tribe_settings_tabs', array() );
-				$this->allTabs    = (array) apply_filters( 'tribe_settings_all_tabs', array() );
-				$this->noSaveTabs = (array) apply_filters( 'tribe_settings_no_save_tabs', array() );
-				if ( is_network_admin() ) {
-					$this->defaultTab = apply_filters( 'tribe_settings_default_tab_network', 'network' );
-					$this->currentTab = apply_filters( 'tribe_settings_current_tab', ( isset( $_GET['tab'] ) && $_GET['tab'] ) ? esc_attr( $_GET['tab'] ) : $this->defaultTab );
-					$this->url        = apply_filters(
-						'tribe_settings_url', add_query_arg(
-							array(
-								'page' => $this->adminSlug,
-								'tab'  => $this->currentTab,
-							), network_admin_url( 'settings.php' )
-						)
-					);
-				}
-				if ( ! is_network_admin() ) {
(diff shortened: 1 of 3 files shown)


src/functions/template-tags/general.php
```diff
@@ -588,43 +588,49 @@ function tribe_register_error( $indexes, $message ) {
...
  *
- * @return string
+ * @param string|array  $group  Which group(s) should be enqueued.
  */
 function tribe_asset_enqueue_group( $group ) {
-	return tribe( 'assets' )->enqueue_group( $group );
+	/** @var Tribe__Assets $assets */
+	$assets = tribe( 'assets' );
+
+	$assets->enqueue_group( $group );
```
mitogh: Missing a return here.

src/functions/template-tags/general.php
```diff
@@ -588,43 +588,49 @@ function tribe_register_error( $indexes, $message ) {
...
  *
- * @return string
+ * @param string|array $slug Slug to enqueue
  */
 function tribe_asset_enqueue( $slug ) {
-	return tribe( 'assets' )->enqueue( $slug );
+	/** @var Tribe__Assets $assets */
+	$assets = tribe( 'assets' );
+
+	$assets->enqueue( $slug );
```
mitogh: Missing return here as well.

src/Tribe/Settings.php
```diff
@@ -295,45 +295,51 @@ public function addNetworkPage() {
...
-							admin_url( self::$parent_page )
-						)
-					);
-				}
-				$this->fields_for_save = (array) apply_filters( 'tribe_settings_fields', array() );
-				do_action( 'tribe_settings_after_do_tabs' );
-				$this->fields = (array) apply_filters( 'tribe_settings_fields', array() );
-				$this->validate();
+			if (
+				! isset( $_GET['page'] )
```
mitogh: Personal preference here but I will change `! isset` with a `empty` as removes the need to use a `!` operator

src/functions/template-tags/general.php
cliffordp: enqueue_group() has no `@return` (i.e. returns void) so this shouldn't either, correct?

src/functions/template-tags/general.php
cliffordp: enqueue() has no `@return` / returns void

src/Tribe/Settings.php
cliffordp: I like empty(), too. Updated.

src/functions/template-tags/general.php
mitogh: In that case the docblock should be updated accordingly then, as right now it says it returns a string or an array.

src/functions/template-tags/general.php
mitogh: In that case the docblock should be updated accordingly then, as right now it says it returns a string or an array.

cliffordp: That's what I did. It used to have @return; now it doesn't.

src/functions/template-tags/general.php
cliffordp: Same: That's what I did. It used to have `@return`; now it doesn't.

src/functions/template-tags/general.php
mitogh: My bad I was overlooking at `@param`

cliffordp: @mitogh - okay to merge?

PR #1066, opened 2019-06-21
Fix/129061 fix ctix radio checkbox labels alignment

cliffordp:
https://central.tri.be/issues/129061#note-10

### DIFF
--- a/src/resources/postcss/tribe-common-admin.pcss
+++ b/src/resources/postcss/tribe-common-admin.pcss
@@ -134,6 +134,9 @@ table.plugins .tribe-plugin-update-message a {
 	display: block;
 	margin: 5px 0 5px 20px;
 	text-indent: -20px;
+	> p {
+		text-indent: 0;
+	}
 }
 .tribe-settings-form .tribe-field-radio label input,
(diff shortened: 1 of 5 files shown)


src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -129,11 +129,17 @@ table.plugins .tribe-plugin-update-message a {
 		margin-top: 0;
 	}
 }
+.tribe-settings-form .tribe-field-checkbox_bool p.tooltip {
```
bordoni: Around this element there a couple of uses of `.tribe-settings-form` to then include a child rule. It would be amazing to move the child rules into a:

```
.tribe-settings-form {
    // .child-rules....
}
```

cliffordp: Nested quite a few I found. Good suggestion.

cliffordp:
@sc0ttkclark please re-review after my multiple commits, incorporating (but slightly changing your edit - mine are more nested to match exactly what they were before, i.e. just conversion to nesting)

PR #1070, opened 2019-06-25
Prevent cache warmup from throwing fatal when cache instance is invalid 

bordoni:
_Ref:_ [C#129860](http://central.tri.be/issues/129860)
---


bordoni:
Closed this in favor of #1072 which solves the problem the correct way.

PR #1076, opened 2019-06-28
Add template-tag for handling element attribute class.

bordoni:
Added a simple way to handle class names, look at the tests for the behavior implemented here.

https://github.com/moderntribe/the-events-calendar/pull/2597

### DIFF
--- a/src/Tribe/Utils/Element_Classes.php
+++ b/src/Tribe/Utils/Element_Classes.php
@@ -149,7 +149,9 @@ class Element_Classes {
 			return;
 		}
 
-		if ( is_string( $arguments ) ) {
+		if ( is_numeric( $arguments ) ) { // phpcs:ignore
+			// Bail on any numeric values
+		} elseif ( is_string( $arguments ) ) {
 			// 'foo bar'
 			$this->parse_string( $arguments );
 		} elseif ( $arguments instanceof \Closure || is_callable( $arguments ) ) {
@@ -161,11 +163,6 @@ class Element_Classes {
 		} elseif ( is_object( $arguments ) ) {
 			// stdClass
 			$this->parse_object( $arguments );
-		} elseif ( is_int( $arguments ) ) {
-			// '130'
-			$this->results[ $arguments ] = $default_value;
-		} elseif ( is_float( $arguments ) ) {
-			$this->results[ (string) $arguments ] = $default_value;
 		}
 	}
 
--- a/src/Tribe/Utils/Element_Classes.php
+++ b/src/Tribe/Utils/Element_Classes.php
@@ -1,0 +1,238 @@
+<?php
+namespace Tribe\Utils;
+
+/**
+ * Class Element_Classes to handle HTML class attribute for elements.
+ *
+ * @since  TBD
+ *
+ * @package Tribe\Utils
+ */
+class Element_Classes {
+	/**
+	 * Store the results of parsing the classes.
+	 *
+	 * @since  TBD
+	 *
+	 * @var array
+	 */
+	protected $result = [];
+
+	/**
+	 * Stores the arguments passed.
+	 *
+	 * @since  TBD
+	 *
+	 * @var array
+	 */
+	protected $arguments = [];
+
+	/**
+	 * Setups an instance of Element Classes.
+	 *
+	 * @since  TBD
+	 *
+	 * @return void
+	 */
+	public function __construct() {
+		$this->arguments = func_get_args();
+	}
+
+	/**
+	 * When invoked this class will return the full HTML class attribute.
+	 *
+	 * @since  TBD
+	 *
+	 * @return string In the format ` class="class1 class2" `
+	 */
+	public function __invoke() {
+		$this->arguments = func_get_args();
+		return $this->get_attribute();
+	}
+
+
+	/**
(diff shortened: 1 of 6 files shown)


src/Tribe/Utils/Element_Classes.php
```diff
@@ -0,0 +1,238 @@
...
+		foreach ( $this->results as $key => $val ) {
+			if ( ! $val ) {
+				continue;
+			}
+
+			$classes[] = $key;
+		}
+
+		$classes = array_map( 'sanitize_html_class', $classes );
+		$classes = array_unique( $classes );
```
sc0ttkclark:
```suggestion
		$classes = array_filter( array_unique( $classes ) );
```

Might be a nice if the result of sanitizing means an empty string

src/functions/template-tags/html.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @since  TBD
+ *
+ * @param  mixed $classes,... unlimited Any amount of params to be rendered as classes.
+ *
+ * @return string
+ */
+function tribe_classes() {
+	$element_classes = new Element_Classes( func_get_args() );
+	echo $element_classes->get_attribute();
```
mitogh: I think adding `esc_attr` to escape the attribute value would be great here as well.

src/Tribe/Utils/Element_Classes.php
```diff
@@ -0,0 +1,238 @@
...
+	 * Gets the full HTML class attribute for this instance of Element Classes.
+	 * It will contain a space on each end of the attribute.
+	 *
+	 * @since  TBD
+	 *
+	 * @return string In the format ` class="class1 class2" `
+	 */
+	public function get_attribute() {
+		$classes = $this->get_classes_as_string();
+		return " class=\"{$classes}\" ";
```
mitogh: I think it should return empty string if `$classes` is empty.

mitogh: Otherwise will output ` class="" `

src/Tribe/Utils/Element_Classes.php
```diff
@@ -0,0 +1,238 @@
...
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string  $arguments     Space separated string of classes to be parsed
+	 * @param  boolean $default_value What is the default value for a given class.
+	 *
+	 * @return void
+	 */
+	protected function parse_string( $arguments, $default_value = true ) {
+		$values = preg_split( '/\s+/', $arguments, -1, PREG_SPLIT_NO_EMPTY );
```
mitogh: `preg_split` might return `false` on failure so `$values` will trigger an error in that case as `false` is not an array

src/functions/template-tags/html.php
bordoni: Used `esc_html()` since `esc_attr` would remove the correct `"'` usage around the `class=""`

mitogh: Oh I see I think instead of writing the ` class="" ` might be better to return only the result that goes inside passing the controller to the caller to use something like: `class="<?php tribe_classes(); ?>"`.

Kind of the same way https://github.com/JedWatson/classnames works.

src/Tribe/Utils/Element_Classes.php
```diff
@@ -0,0 +1,249 @@
...
+		} elseif ( is_array( $arguments ) ) {
+			// ['foo', 'bar', ...] || ['foo' => TRUE, 'bar' => FALSE, ...]
+			$this->parse_array( $arguments );
+		} elseif ( is_object( $arguments ) ) {
+			// stdClass
+			$this->parse_object( $arguments );
+		} elseif ( is_int( $arguments ) ) {
+			// '130'
+			$this->results[ $arguments ] = $default_value;
+		} elseif ( is_float( $arguments ) ) {
```
Camwyn: Classes cannot start with a digit so I'm not sure this is valid?
ie: Converting, say, `2.5` to a string does not make it a valid CSS class.

I'd suggest that the logic throw out any classes that start with `a digit, two hyphens, or a hyphen followed by a digit` as those are all invalid? Or maybe a separate`validate_class` function.

Also worth noting that WP `sanitize_html_class` (above) does not validate classes - it only strips out illegal characters (and it needs updating)

bordoni: Since we are relying on the WP method, I would not care about that particular case. Unless we ported their code internally and applied rules. What do you think? Should we implement our own?

Camwyn: I'd still remove this case - unless there is a compelling reason to allow floats as classes? (I expect not since they are invalid) - or tell it to bail on a float, etc.

I'd start the `if-else` chain with checks for:
```
if ( substr( $string, 0, 2 ) === '--') { 
// starts with two hyphens - bail
} elseif( is_numeric( substr( $string, 0, 1  )) ) {
// starts with a number - bail
} elseif( substr( $string, 0, 1 ) === '-` && is_numeric( substr( $string, 1, 1 ) ) ) {
// starts with a hyphen + a number  - bail
}
```
Since once it's out there we have no control over what gets passed to it via hooks, etc, it seems prudent to put those sanity checks in?

PR #1081, opened 2019-07-05
Feature/130235 theme overrides twenty nineteen

paulmskim:
https://central.tri.be/issues/130235


src/resources/postcss/base/typography/_anchors.pcss
```diff
@@ -3,12 +3,20 @@
...
 		&:focus,
 		&:active,
 		&:visited {
 			color: var(--color-text-primary);
+			outline: none;
+			text-decoration: none;
+		}
+
+		/* Twenty Nineteen Theme Override */
+		.tribe-theme-twentynineteen .entry & {
```
juanfra: Maybe I'm wrong, but I remember we decided to add all the theme overrides at the bottom of the file, in order to make it more readable, traceable and easy to understand. And because once we start working on the different theme overrides we could end up with overrides here and there.

paulmskim: Yup, you're right. But what comes after this are `tribe-common-anchor` class styles, does it make sense to put it after that? I think I'll also put those large comment bars that we've used in other files to make the positioning of this more prominent.

juanfra: Love the large comment bars to make it more prominent. I feel like if we're going to format our files to have the overrides at the end of the file we should follow that. Maybe this file is a bit short and doesn't change so much if we put it here or there. But let's think of other file with overrides here and there, for the different themes, if we don't follow the format we (ourselves) defined then it'll be hard to keep things consistent.

bordoni: We should define these formats for overwrite headers with comments.

src/resources/postcss/base/typography/_anchors.pcss
```diff
@@ -3,12 +3,22 @@
 	a {
 		color: var(--color-text-primary);
 		cursor: pointer;
+		outline: none;
```
ryanurban: Same q from above

src/resources/postcss/base/_buttons.pcss
```diff
@@ -4,7 +4,12 @@
 	 * ------------------------------------------------------------------------- */
 
 	button {
-		background-color: var(--color-background);
+		background-color: transparent;
```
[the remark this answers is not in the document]
paulmskim: Yup definitely. the only thing that we've split so far for skeleton / full style is the reset. So base, components, and views have not yet been split given the amount of design updates we've been having.

src/resources/postcss/base/typography/_anchors.pcss
paulmskim: @juanfra my main concerns in putting the override after the `.tribe-common-cta` classes are the order of the CSS and the organization. Now in this specific case it might not make a difference, but if we're doing color or font overrides, the order will matter due to the cascade. I'm not opposed to putting `a` overrides below `tribe-common-cta` classes, but then we'll also need to put `tribe-common-cta` override classes below that as well, which can get messy if we're not thinking of order (which touches upon my organization concern). I'm open to options for a good organizational structure which keeps in mind cascading and easy understanding of overriding.

@bordoni how do you mean? do you have an example?

paulmskim: @juanfra moved the theme overrides to its own selector and added a comment block for theme overrides, check the diff and let me know what you think.

PR #1083, opened 2019-07-05
VR > DQA Refinements

juanfra:
🎫 https://central.tri.be/issues/130194

### DIFF
--- a/src/resources/postcss/base/typography/_body.pcss
+++ b/src/resources/postcss/base/typography/_body.pcss
@@ -55,4 +55,14 @@
 	.tribe-common-b3--bold {
 		font-weight: var(--font-weight-bold);
 	}
+
+	/* -------------------------------------------------------------------------
+	 * Body 1 for --viewport-medium
+	 * ------------------------------------------------------------------------- */
+
+	.tribe-common-b1--min-medium {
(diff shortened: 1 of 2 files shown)


src/resources/postcss/base/typography/_body.pcss
```diff
@@ -55,4 +55,14 @@
 	.tribe-common-b3--bold {
 		font-weight: var(--font-weight-bold);
 	}
+
+	/* -------------------------------------------------------------------------
+	 * Body 1 for --viewport-medium
+	 * ------------------------------------------------------------------------- */
+
+	.tribe-common-b1--min-medium {
```
paulmskim: nice! :+1:

src/resources/postcss/base/typography/_headings.pcss
```diff
@@ -75,7 +75,11 @@
 
 	.tribe-common-h6 {
 		@mixin heading;
-		@mixin heading-6;
+		@mixin mobile-heading-6;
+
+		@media (--viewport-medium) {
+			@mixin desktop-heading-6;
```
paulmskim: this is good, there is also a `h6--min-medium` below that should use `desktop-heading-6` mixin below as well.

PR #1086, opened 2019-07-09
fix(tooltips assets): replace `tribe_asset` with `wp_enqueue_style`.

Camwyn:
The tooltip assets aren't getting enqueued properly, so replacing them with the wp built-in.

This needs a release and likely a ticket - at which point someone can dig in and see if there is actually a bug in `tribe_asset`

### DIFF
--- a/src/Tribe/Service_Providers/Tooltip.php
+++ b/src/Tribe/Service_Providers/Tooltip.php
@@ -35,12 +35,9 @@ class Tribe__Service_Providers__Tooltip extends tad_DI52_ServiceProvider {
 	 * @since 4.9.8
 	 */
 	public function add_tooltip_assets() {
-		tribe_asset(
-			Tribe__Main::instance(),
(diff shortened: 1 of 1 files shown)


bordoni:
@Camwyn i would love for us to figure this out, it should be a simple fix, in which pages does tool tip need to be enqueued that currently is not.

sc0ttkclark:
I've looked this over and I'm not sure why the original code isn't working -- everything I've been looking at seems to point to that it should just work.

Camwyn:
@bordoni:
> @Camwyn i would love for us to figure this out, it should be a simple fix, in which pages does tool tip need to be enqueued that currently is not.

All of the ticket report pages in the admin, and the corresponding ones on the FE for CT

Camwyn:
Closing in favor of https://github.com/moderntribe/tribe-common/pull/1097

PR #1089, opened 2019-07-16
Update freemius vendor library to 2.3.0

mitogh:
Ticket: https://central.tri.be/issues/130281


mitogh:
Need to run a smoke test locally before an **official** code review.

PR #1098, opened 2019-07-24
fix white links in wp-admin Plugins List

cliffordp:
https://central.tri.be/issues/131353


src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -85,6 +85,11 @@ table.plugins {
...
 		margin:     6px 0;
 		padding:    10px 12px;
+
+		a {
+			color:           white;
+			text-decoration: underline;
+		}
 	}
 
 	h4 {
```
sc0ttkclark: Should this have been under the update message too?

PR #1102, opened 2019-08-02
update hover focus active color for caret right and left

paulmskim:
https://central.tri.be/issues/131554


paulmskim:
Well I'm merging it anyway!

PR #1110, opened 2019-08-12
Release/b19.10

sc0ttkclark:


src/Tribe/PUE/Notices.php
```diff
@@ -387,7 +387,7 @@ protected function render_notice( $slug, $inner_html ) {
 		$html =
 			'<div class="api-check">
 				<div class="tribe-mascot">
-					<img src="' . $mascot . '"/>
+					<img src="' . $mascot . '" style="max-height: 150px; max-width: 150px; height: 100%; width: auto;"/>
```
sc0ttkclark:
```suggestion
					<img src="' . esc_url( $mascot ) . '" style="max-height: 150px; max-width: 150px; height: 100%; width: auto;"/>
```

It would be great to late escape here and remove the esc_url() on line 385 @bordoni

PR #1122, opened 2019-08-22
feat(version checks): Add functions to test installed version of plugins.

Camwyn:
Add `tribe_installed_before` `tribe_installed_after` and `tribe_installed_on` to test the install version against a passed version. 
Utilizes the stored versions ( example: `previous_event_tickets_versions` )
Requires the plugin have the `VERSION` and `VERSION_HISTORY_SLUG` constants set. `VERSION_HISTORY_SLUG` is a new constant being added specifically for this testing.

:ticket: https://central.tri.be/issues/133048


Camwyn:
@bordoni I added you as this is something you might be able to use in TEC, etc. So I wanted you to be aware of it.

@borkweb I added you because of the conversation in the scoping doc. Plus, 🐫 💨

src/functions/utils.php
```diff
@@ -638,4 +638,58 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
 		$args = apply_filters( 'tribe_register_rest_route_args', $args, $namespace, $route, $override );
 		return register_rest_route( $namespace, $route, $args, $override );
 	}
-}
\ No newline at end of file
+}
+
+if ( ! function_exists( 'tribe_installed_before' ) ) {
```
bordoni: I don't think we need these anymore. These are not supposed to replaced right?

src/functions/utils.php
```diff
@@ -638,4 +638,58 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
...
+	 */
+	function tribe_installed_before( $class, $version ) {
+		$pervious_versions = Tribe__Settings_Manager::get_option( $class::VERSION_HISTORY_SLUG );
+		$install_version = !  empty( $pervious_versions[1] ) ? $pervious_versions[1] : $class::VERSION;
+
+		return 0 > version_compare ( $install_version, $version );
+	}
+}
+
+if ( ! function_exists( 'tribe_installed_after' ) ) {
```
bordoni: I don't think we need these anymore. These are not supposed to replaced right?

src/functions/utils.php
```diff
@@ -638,4 +638,58 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
...
+	 *
+	 * @param object $class the main plugin class.
+	 * @param string $version a SemVer version string.
+	 * @return boolean
+	 */
+	function tribe_installed_before( $class, $version ) {
+		$pervious_versions = Tribe__Settings_Manager::get_option( $class::VERSION_HISTORY_SLUG );
+		$install_version = !  empty( $pervious_versions[1] ) ? $pervious_versions[1] : $class::VERSION;
+
+		return 0 > version_compare ( $install_version, $version );
```
bordoni: No space between method and `()`

src/functions/utils.php
```diff
@@ -638,4 +638,58 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
...
+	 *
+	 * @param object $class the main plugin class.
+	 * @param string $version a SemVer version string.
+	 * @return boolean
+	 */
+	function tribe_installed_after( $class, $version ) {
+		$pervious_versions = Tribe__Settings_Manager::get_option( $class::VERSION_HISTORY_SLUG );
+		$install_version = !  empty( $pervious_versions[1] ) ? $pervious_versions[1] : $class::VERSION;
+
+		return 0 < version_compare ( $install_version, $version );
```
bordoni: No space between method and `()`

src/functions/utils.php
```diff
@@ -638,4 +638,58 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
...
+	 *
+	 * @since TBD
+	 *
+	 * @param object $class the main plugin class.
+	 * @param string $version a SemVer version string.
+	 * @return boolean
+	 */
+	function tribe_installed_before( $class, $version ) {
+		$pervious_versions = Tribe__Settings_Manager::get_option( $class::VERSION_HISTORY_SLUG );
+		$install_version = !  empty( $pervious_versions[1] ) ? $pervious_versions[1] : $class::VERSION;
```
bordoni: We should create method for this piece that is being repeated; like `tribe_plugin_version_compare()`

src/functions/utils.php
Camwyn: No, I assumed they were in case of both versions of common being run - but that has been handled so 🔪

PR #1124, opened 2019-08-23
Fix/131080 improve dependencies messaging

cliffordp:
https://central.tri.be/issues/131080

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -5,6 +5,7 @@
 = [4.9.16] TBD =
 
 * Tweak - Changed the 'url' validation error text to just say it needs to be valid, not that it has to be a valid *absolute* URL [72214]
+* Fix - Enqueue Thickbox script on all admin pages when needed
 
 = [4.9.15] 2019-08-22 =
 
@@ -17,7 +18,6 @@
 * Tweak - Update Lodash version on Block editor to prevent any possibility of a security issue with the package. From v4.17.11 to v4.17.15 [131421]
 * Fix - Prevent mascot image to get blown up out of proportions to a larger size on buggy CSS loading. [131910]
 * Language - 0 new strings added, 66 updated, 1 fuzzied, and 4 obsoleted
-
 
 = [4.9.13] 2019-07-25 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -5,6 +5,7 @@
 = [4.9.16] TBD =
 
 * Tweak - Changed the 'url' validation error text to just say it needs to be valid, not that it has to be a valid *absolute* URL [72214]
+* Tweak - Smarter plugin dependency checking with more accurate admin notices if not all requirements are satisfied [131080]
 * Fix - Enqueue Thickbox script on all admin pages when needed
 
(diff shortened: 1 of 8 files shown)


readme.txt
```diff
@@ -5,6 +5,8 @@
 = [4.9.16] TBD =
 
 * Tweak - Changed the 'url' validation error text to just say it needs to be valid, not that it has to be a valid *absolute* URL [72214]
+* Tweak - Smarter plugin dependency checking with more accurate admin notices if not all requirements are satisfied [131080]
+* Fix - Enqueue Thickbox script on all admin pages when needed
```
```suggestion
* Fix - Enqueue Thickbox script on all admin pages when needed [131080]
```

src/Tribe/PUE/Checker.php
```diff
@@ -1003,6 +1003,9 @@ public function ajax_validate_key() {
 		 * @return string
 		 */
 		private function get_api_message( $info ) {
+			// Make sure Thickbox is available and consistent appearance regardless of which admin page we're on
+			add_thickbox();
```
sc0ttkclark: This seems like the wrong place to do this. Where we use the message to add a notice, we should do that there.

src/Tribe/Plugins.php
```diff
@@ -9,71 +9,71 @@
...
-		 * A list of tribe plugin's details in this format:
+		 * A list of tribe plugin's details in this array format:
 		 *
-		 * array(
+		 * [
 		 *  'short_name'   => Common name for the plugin, used in places such as WP Admin messages
 		 *  'class'        => Main plugin class
 		 *  'thickbox_url' => Download or purchase URL for plugin from within /wp-admin/ thickbox
-		 *  )
+		 *  ]
```
sc0ttkclark:
```suggestion
		 * ]
```

src/Tribe/Dependency.php
```diff
@@ -4,7 +4,7 @@
```
sc0ttkclark: The more I look at this class, I wonder if we can leverage it for the notices in each plugin that we have custom `admin_notice` hooks for and all the extra custom code.

sc0ttkclark: Like in https://github.com/moderntribe/tribe-common/blob/acace8801038dc7fbfd7c8a6c3f5de5ca6b3da31/src/Tribe/Dependency.php#L319

We could let the `$version` support the same data that sets `$dependent_plugin` a few lines below. If plugin not registered, it could be added to the list of plugins right there automatically.

src/Tribe/PUE/Checker.php
cliffordp: Yeah... `ajax_validate_key()` is the only place I traced this usage back to, but enqueuing within an ajax response probably isn't right either... so I just undid this.

src/Tribe/Dependency.php
cliffordp: Are you saying to process just `$plugin` instead of the dependencies and therefore avoid 3 separate calls to `$this->has_valid_dependencies()`?

PR #1125, opened 2019-08-26
#132279 - add trigger to update modal cart when get tickets clicked

jesseeproductions:
_Ref:_ [C#132279](https://central.tri.be/issues/132279)

Not sure this is the best way to hook into this, but what I am trying to do is every time the modal opens I have an event to connect to from another script to update cart information in the modal.


sc0ttkclark:
Great thinking @jesseeproductions!

PR #1134, opened 2019-09-02
feat(functions/utils.php) add the tribe_get_request_vars function

lucatume:
This PR adds some support functions for Alpha 2 work.

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -639,3 +639,35 @@ if ( ! function_exists( 'tribe_register_rest_route' ) ) {
 		return register_rest_route( $namespace, $route, $args, $override );
 	}
 }
+
+if ( ! function_exists( 'tribe_get_request_vars' ) ) {
+	/**
+	 * Returns the sanitized version of the `$_REQUEST` super-global array.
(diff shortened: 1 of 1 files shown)


src/functions/utils.php
```diff
@@ -638,4 +638,36 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
 		$args = apply_filters( 'tribe_register_rest_route_args', $args, $namespace, $route, $override );
 		return register_rest_route( $namespace, $route, $args, $override );
 	}
+}
+
+if ( ! function_exists( 'tribe_get_request_vars' ) ) {
+	/**
+	 * Returns the sanitized version of the `$_REQUEST` super-global array.
```
sc0ttkclark: We should probably note that this is cached and reused. Any manual changes to $_REQUEST will not be reflected.

lucatume: This is a good idea, thanks

lucatume: I've added a note and added support for a `$refresh` parameter to allow refreshing the cache.

PR #1139, opened 2019-09-03
make tribe_get_install_version() handle array

cliffordp:
https://central.tri.be/issues/130974 but not really

These edits were needed to get ET+'s PR working: https://github.com/moderntribe/event-tickets-plus/pull/845

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -694,7 +694,7 @@ if ( ! function_exists( 'tribe_get_class_instance' ) ) {
 	}
 }
 
-if ( ! function_exists( 'tribe_get_lowest_version_ever_installed' ) ) {
+if ( ! function_exists( 'tribe_get_least_version_ever_installed' ) ) {
 	/**
 	 * Gets the lowest version number ever installed for the specified class of a plugin having a
 	 * `version_history_slug` property or a `VERSION` constant (i.e. Main classes).
@@ -710,7 +710,7 @@ if ( ! function_exists( 'tribe_get_lowest_version_ever_installed' ) ) {
 	 *
 	 * @return string|boolean The SemVer version string or false if no info found.
 	 */
-	function tribe_get_lowest_version_ever_installed( $class ) {
+	function tribe_get_least_version_ever_installed( $class ) {
 		$instance = tribe_get_class_instance( $class );
 
 		if ( $instance ) {
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -49,13 +49,14 @@ if ( ! function_exists( 'tribe_check_plugin' ) ) {
 	 *
 	 * @since 4.9
 	 *
-	 * @param string $main_class   The Main/base class for this plugin
+	 * @param string $main_class The Main/base class for this plugin
 	 *
 	 * @return bool Indicates if plugin should continue initialization
 	 */
 	function tribe_check_plugin( $main_class ) {
 
-		$tribe_dependency    = Tribe__Dependency::instance();
+		$tribe_dependency = Tribe__Dependency::instance();
+
 		return $tribe_dependency->check_plugin( $main_class );
 
 	}
@@ -66,12 +67,13 @@ if ( ! function_exists( 'tribe_append_path' ) ) {
 	 * Append a path fragment to a URL preserving query arguments
 	 * and fragments.
 	 *
-	 * @param string $url A full URL in the `http://example.com/?query=var#frag` format.
+	 * @since 4.3
+	 *
 	 * @param string $path The path to append to the existing, if any, one., e.g. `/some/path`
 	 *
-	 * @return mixed|string
+	 * @param string $url  A full URL in the `http://example.com/?query=var#frag` format.
 	 *
-	 * @since 4.3
+	 * @return mixed|string
 	 */
 	function tribe_append_path( $url, $path ) {
 		$path = trim( $path, '/' );
@@ -142,7 +144,7 @@ if ( ! function_exists( 'tribe_get_request_var' ) ) {
 	 * @return mixed
 	 */
 	function tribe_get_request_var( $var, $default = null ) {
-		return Tribe__Utils__Array::get_in_any( array( $_GET, $_POST, $_REQUEST ), $var, $default );
+		return Tribe__Utils__Array::get_in_any( [ $_GET, $_POST, $_REQUEST ], $var, $default );
 	}
 }
 
@@ -190,7 +192,7 @@ if ( ! function_exists( 'tribe_is_truthy' ) ) {
 		 *
 		 * @param array $truthy_strings
 		 */
-		$truthy_strings = (array) apply_filters( 'tribe_is_truthy_strings', array(
+		$truthy_strings = (array) apply_filters( 'tribe_is_truthy_strings', [
 			'1',
 			'enable',
 			'enabled',
@@ -198,7 +200,7 @@ if ( ! function_exists( 'tribe_is_truthy' ) ) {
 			'y',
 			'yes',
 			'true',
-		) );
+		] );
 		// Makes sure we are dealing with lowercase for testing
 		if ( is_string( $var ) ) {
 			$var = strtolower( $var );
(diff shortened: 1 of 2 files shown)


src/functions/utils.php
```diff
@@ -645,19 +645,41 @@ function tribe_register_rest_route( $namespace, $route, $args = array(), $overri
...
+	}
+
+	if ( ! empty( $instance->version_history_slug ) ) {
+		$history = (array) Tribe__Settings_Manager::get_option( $instance->version_history_slug );
+
+		// '0' may be logged as a version number, which isn't useful, so we remove it
+		$history = array_filter( $history );
+
+		// Sort the array so smallest version number is first (likely how the array is stored anyway)
+		usort( $history, 'version_compare' );
```
sc0ttkclark: Why sort? Why not just use the last one?

cliffordp: You mean the first one that's not a zero?

We're looking for the oldest-registered version number.

sc0ttkclark: The zero was removed via array_filter() already though, right?

So wouldn't you just check `$history[0]`?

Camwyn:
```
// array_filter removes the '0' item from the beginning if it exists
$history = array_filter( $history );

if ( empty( $history ) ) {
	return false;
}

// Gets the first item in the array - the last added, therefore the oldest.
$intall_version = array_shift($history);

if (  ! empty( $intall_version ) ) {
	return $intall_version;
}

return false;
```

Camwyn: You could also just:
```
while ( ! empty( $history ) && empty( $history[0] ) ) {
    unset( $history[0] );
}

// just in case ;)
if ( empty( $history[0] ) ) {
	return false;
}

// We found it, return it
if (  ! empty( $history[0] ) ) {
	return $history[0];
}

// fallback
return false;
```

sc0ttkclark: I mean you could just take out the 'just in case' conditional at that point :)

Camwyn: True - edited 😛

cliffordp: @sc0ttkclark, is it possible someone installs version 2, finds a known issue, downgrades to version 1, then first entry is `2` but oldest version (from sorting) is actually `1`?

sc0ttkclark: That depends on our intention here, if we want the oldest/first installed version on a new installation, that would be v2. It just depends on what we want intent to be on this function.

cliffordp: i guess that's a question for @Camwyn, as he was the one who originally wrote these new functions... the function name and docblock should be very clear either way you/team decide.

Camwyn: That depends on how the history is used - do we record changes back and forth? Do we _use_ that info?

Maybe the best answer here is to modify where we store it to sort and only store unique values?

cliffordp: For my purpose, I want to know the earliest version ever installed, although the first version ever installed is likely the same for nearly all installs. If someone's initial install is v4.11.0, but for some reason decided to use 4.10.99 instead for a period of time, I _do_ want to send them a notice because they were using a version prior to 4.11.0

If this is a different use case for why these functions were created, I can do it my own way - just need to know these comparison functions' final decision so I can decide whether or not to use them.

Camwyn: @cliffordp I'm curious, what's the use case?

cliffordp: The attached Central ticket and ET+ pull request

Camwyn: At this point, for sanity's sake, I think the comparison should do something like:
```
$history = array_filter( $history );
$history = array_unique($history );
usort( $history, 'version_compare' );
```
Before doing comparisons.

Probably in a separate small function? So we can just call `version_history_clean($history);` and then `array_shift($history)` and do our comparison?

sc0ttkclark: As per latest slack discussion: https://tribe.slack.com/archives/CHXD8BSH4/p1567698140042700

- [ ] Clarify that this function will return the *first* installed version, not the *earliest* version number and rename function to `tribe_get_first_installed_version`
- [ ] Make the change that @Camwyn suggested in https://github.com/moderntribe/tribe-common/pull/1139#discussion_r320798585 to this function
- [ ] Adjust the code formatting in this function, the tabs/spaces look off
- [ ] Replace any usage of `tribe_get_install_version` to use `tribe_get_first_installed_version` instead
- [ ] Create a new function that will return the earliest version number using the sorting logic `tribe_get_earliest_installed_version`

src/functions/utils.php
```diff
@@ -595,138 +602,343 @@ function has_blocks( $post = null ) {
...
+
+						break;
+					}
+				}
+			}
+		}
 
-	// No version set.
-	return false;
+		return $instance;
```
Camwyn: This is _way_ better :)
If I may offer one more suggestion: instead of setting and carrying `$instance` - just return what you find, and return null here at the end?

If you do that, you should also use `empty( $instance )` instead of `! $instance` so it won't choke if `$instance` isn't uninitialized. 😉

src/functions/utils.php
```diff
@@ -595,138 +602,343 @@ function has_blocks( $post = null ) {
...
- * @since TBD
- *
- * @param object $class The plugin object.
- * @param string $version The SemVer version string.
- *
- * @return boolean Whether the plugin was installed at/on the passed version.
- */
-function tribe_installed_on( $class, $version ) {
-	$install_version = tribe_get_install_version( $class );
+if ( ! function_exists( 'tribe_get_currently_installed_version' ) ) {
```
cliffordp: @sc0ttkclark `tribe_get_currently_installed_version()` is a better name for what I understood `tribe_get_latest_installed_version()` should be (last item on your checklist)

However, I don't really like it. I think it is only really that useful if we were to add "current version greater/less than" functions. Do we want those? If not, should I remove this function?

src/functions/utils.php
sc0ttkclark: This whole foreach block should be replaced with a simply try/catch:

```php
try {
	return tribe( $class );
} catch ( RuntimeException $exception ) {
	return null;
}
```

src/functions/utils.php
sc0ttkclark: I think there's still value in this function, we can keep it.

src/functions/utils.php
cliffordp: @sc0ttkclark did you try this? It didn't work for me, which is why I did it the way I did.

src/functions/utils.php
cliffordp: @sc0ttkclark OK. Should I add "if current version is greater/less than" functions?

sc0ttkclark: I’m not sure we need it right away, but if you’d like to add it you can

src/functions/utils.php
sc0ttkclark: This is supposed to work based on what I’ve read into tribe() code

Camwyn: I am not so sure, but will have to test.
From the `tribe()` readme:
> @return mixed|object|Tribe__Container     The instance of the requested class. Please note that the cardinality of the class is controlled registering it as a singleton using `tribe_singleton` or `tribe_register`; if the `$slug_or_class` parameter is null then **the container itself will be returned**.

_Goes to test_
```
try {
	return tribe( 'foo.bar.baz' );
} catch ( RuntimeException $exception ) {
	return 'exception';
}
```

Returns `exception` - so it appears Scott is correct.

src/functions/utils.php
```diff
@@ -251,20 +254,21 @@ function tribe_sort_by_priority( $a, $b ) {
 	/**
 	 * Normalizes a list of terms to a list of fields.
 	 *
-	 * @param $terms A term or array of terms to normalize.
+	 * @since 4.5
+	 *
 	 * @param string $taxonomy The terms taxonomy.
-	 * @param string $field Teh fields the terms should be normalized to.
+	 * @param string $field    Teh fields the terms should be normalized to.
```
Camwyn:
```suggestion
	 * @param string $field    The fields the terms should be normalized to.
```

src/functions/utils.php
```diff
@@ -251,20 +254,21 @@ function tribe_sort_by_priority( $a, $b ) {
...
 	 *
```
Camwyn:
```suggestion
```

src/functions/utils.php
Camwyn: They're not _really_ necessary, since they'll both essentially be 
`version_compare( tribe_get_currently_installed_version( $class ), $version )`

sc0ttkclark:
One thing this PR would substantially benefit from is unit tests for these new functions. Can you focus on those tomorrow @cliffordp?

cliffordp:
@sc0ttkclark @Camwyn 
image

sc0ttkclark:
@cliffordp That specific call in `src/Tribe/Admin/Notices.php:58` should be using the correct alias name which is `tickets.main`. We should be using the aliases in most places.

cliffordp:
The point of this function is to be able to get an instance given a slug, class name, or instantiated class. Failing on ET's Main feels really wrong

sc0ttkclark:
ET's main is using the antiquated singleton class structure instead of the containerized singleton common allows for. We haven't migrated it over, so consider this an edge case there.

PR #1140, opened 2019-09-04
fix tooltip dependencies

paulmskim:
Fix tooltip dependencies

### DIFF
--- a/src/Tribe/Service_Providers/Tooltip.php
+++ b/src/Tribe/Service_Providers/Tooltip.php
@@ -32,14 +32,14 @@ class Tribe__Service_Providers__Tooltip extends tad_DI52_ServiceProvider {
 	/**
 	 * Register assets associated with tooltip
 	 *
-	 * @since 4.9.8
+	 * @since TBD
 	 */
 	public function add_tooltip_assets() {
 		tribe_asset(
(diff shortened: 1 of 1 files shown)


paulmskim:
@Camwyn please check to make sure that these are the dependencies you intended.

src/Tribe/Service_Providers/Tooltip.php
```diff
@@ -32,22 +32,22 @@ private function hook() {
 	/**
 	 * Register assets associated with tooltip
 	 *
-	 * @since 4.9.8
```
juanfra: I think this shouldn't be modified if the function is already there. For the rest is good to go!

src/Tribe/Service_Providers/Tooltip.php
```diff
@@ -32,22 +32,22 @@ private function hook() {
...
+	 * @since TBD
 	 */
 	public function add_tooltip_assets() {
 		tribe_asset(
 			Tribe__Main::instance(),
 			'tribe-tooltip-css',
 			'tooltip.css',
-			[],
+			[ 'tribe-common-style' ],
```
Camwyn: Locally (G19.11), this prevents the styles from being loaded in the admin?

PR #1156, opened 2019-09-25
Feat - void_query support

lucatume:
Ticket: https://central.tri.be/issues/134426 

This PR:

* ensures the request var values read from the context are sanitized (as strings)
* adds support, in the repository, for the `void_query` query var; this will void the query in an args compat way

### DIFF
--- a/src/Tribe/Repository.php
+++ b/src/Tribe/Repository.php
@@ -559,6 +559,10 @@ abstract class Tribe__Repository
 	public function build_query( $use_query_builder = true ) {
 		$query = null;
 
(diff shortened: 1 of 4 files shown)


src/Tribe/Repository.php
```diff
@@ -1068,7 +1072,9 @@ public function filter_by_simple_tax_schema( $value ) {
 	 * {@inheritdoc}
 	 */
 	public function by( $key, $value = null ) {
-		if ( $this->void_query ) {
+		if ( $this->void_query || ( $key === 'void_query' && false !== $value ) ) {
```
bordoni:
```suggestion
		if ( $this->void_query || ( 'void_query' === $key && false !== $value ) ) {
```

PR #1164, opened 2019-10-03
Initial coding for idea to intelligently append to localize script.

Camwyn:
See https://tribe.slack.com/archives/C0258NY1C/p1570054706033400


src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
+		$data = $wp_scripts->get_data( $enqueued_script, 'data');
+
+		if ( empty($data ) ) {
+			// If no data, then just do the default
+			wp_localize_script($enqueued_script, 'obj', $localized_data );
```
sc0ttkclark:
```suggestion
			wp_localize_script( $enqueued_script, 'obj', $localized_data );
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 * @since TBD
+	 *
+	 * @param string|object $enqueued_script The script handle we're saving data for - must already be loaded!
+	 * @param array         $localized_data The data we want to save.
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
+		$data = $wp_scripts->get_data( $enqueued_script, 'data');
```
sc0ttkclark:
```suggestion
		$data = $wp_scripts->get_data( $enqueued_script, 'data' );
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+
+		if ( empty($data ) ) {
+			// If no data, then just do the default
+			wp_localize_script($enqueued_script, 'obj', $localized_data );
+		} else {
+			// If we have data, we append to the existing data!
+			if ( ! is_array( $data ) ) {
+				$data = json_decode(str_replace('var obj = ', '', substr($data, 0, -1)), true);
```
sc0ttkclark:
```suggestion
				$data = json_decode( str_replace( 'var obj = ', '', substr( $data, 0, -1 ) ), true );
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+			}
+			foreach($data as $key => $value) {
```
sc0ttkclark:
```suggestion

			foreach( $data as $key => $value ) {
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+				$localized_data[$key] = $value;
```
sc0ttkclark:
```suggestion
				$localized_data[ $key ] = $value;
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+			}
+
+			$wp_scripts->add_data($enqueued_script, 'data', '');
```
sc0ttkclark:
```suggestion
			$wp_scripts->add_data( $enqueued_script, 'data', '' );
```

src/functions/utils.php
```diff
@@ -976,3 +976,39 @@ function tribe_get_request_vars( $refresh = false ) {
...
+
+			wp_localize_script($enqueued_script, 'obj', $localized_data);
```
sc0ttkclark:
```suggestion
			wp_localize_script( $enqueued_script, 'obj', $localized_data );
```

src/functions/utils.php
```diff
@@ -976,3 +976,41 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
+
+		$data = $wp_scripts->get_data( $enqueued_script, 'data' );
+
+		if ( empty($data ) ) {
+			// If no data, then just do the default
```
sc0ttkclark:
```suggestion
			// If no data, then just do the default.
```

src/functions/utils.php
```diff
@@ -976,3 +976,41 @@ function tribe_get_request_vars( $refresh = false ) {
...
+			wp_localize_script( $enqueued_script, 'obj', $localized_data );
+		} else {
+			// If we have data, we append to the existing data!
```
sc0ttkclark:
```suggestion
			// If we have data, we append to the existing data.
```

Camwyn:
@sc0ttkclark @bordoni I updated this branch - how do we move forward on this?

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 * @param array         $localized_data The data we want to save.
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
+
+		$data = $wp_scripts->get_data( $enqueued_script, 'data' );
+
+		if ( empty($data ) ) {
```
sc0ttkclark, line 999:
```suggestion
		if ( empty( $data ) ) {
```

sc0ttkclark:
Still liking this, we should try and find a path to release for this @lucatume @neillmcshea

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string|object $enqueued_script The script handle we're saving data for - must already be loaded!
+	 * @param array         $localized_data The data we want to save.
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
```
lucatume, line 995: Add a check to make sure this is, in fact, an instance of `WP_Scripts`.

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 * for our script. Then we edit existing elements rather than duplicating them.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|object $enqueued_script The script handle we're saving data for - must already be loaded!
+	 * @param array         $localized_data The data we want to save.
+	 *
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
```
lucatume, line 994: Personally, I do like functions or methods that leak knowledge of their implementation out of of their API.
The "maybe" in the name suggests *how* the function works, that is not really relevant to the client code.
I would rename this to `tribe_localize_script` to mean "this is tribe replacement for wp_localize_script"; which seems to be what we're going for here.

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+		if ( empty( $data ) ) {
+			// If no data, then just do the default.
+			wp_localize_script( $enqueued_script, 'obj', $localized_data );
+		} else {
+			// If we have data, we append to the existing data.
+			if ( ! is_array( $data ) ) {
+				$data = json_decode( str_replace( 'var obj = ', '', substr( $data, 0, -1 ) ), true );
+			}
+
+			foreach ( $data as $key => $value ) {
```
lucatume, line 1008: Check if `$data` is not false or null at this point, `json_decode` might return that.

As for any other check made by this function: I would handle failure (to localize a script data for whatever reason) using a `bool` (as the `wp_localize_script` does).

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+	 * @return void
+	 */
+	function tribe_localize_maybe_append( $enqueued_script, $localized_data = [] ) {
+		global $wp_scripts;
+
+		$data = $wp_scripts->get_data( $enqueued_script, 'data' );
+
+		if ( empty( $data ) ) {
+			// If no data, then just do the default.
+			wp_localize_script( $enqueued_script, 'obj', $localized_data );
```
lucatume, line 1001: Return a value to make this testable.
```suggestion
			return wp_localize_script( $enqueued_script, 'obj', $localized_data );
```

src/functions/utils.php
```diff
@@ -978,6 +978,44 @@ function tribe_get_request_vars( $refresh = false ) {
...
+				$data = json_decode( str_replace( 'var obj = ', '', substr( $data, 0, -1 ) ), true );
+			}
+
+			foreach ( $data as $key => $value ) {
+				$localized_data[ $key ] = $value;
+			}
+
+			$wp_scripts->add_data( $enqueued_script, 'data', '' );
+
+			wp_localize_script( $enqueued_script, 'obj', $localized_data );
```
lucatume, line 1014:
```suggestion
			return wp_localize_script( $enqueued_script, 'obj', $localized_data );
```

src/functions/utils.php
Camwyn: I think I'd rather let `wp_localize_script` handle it if data is false/null - we still get a boolean back and it adds a level of safety net in that the data still gets localized, just not appended??

lucatume (changes requested):
All fine and dandy, but this will need test coverage.

src/functions/utils.php
```diff
@@ -987,6 +987,54 @@ function tribe_get_request_vars( $refresh = false ) {
...
+				return wp_localize_script( $enqueued_script, 'obj', $localized_data );
+			}
+
+			foreach ( $data as $key => $value ) {
+				$localized_data[ $key ] = $value;
+			}
+
+			$wp_scripts->add_data( $enqueued_script, 'data', '' );
+
+			return wp_localize_script( $enqueued_script, 'obj', $localized_data );
```
lucatume, line 1033: Whatever test you write, it will end up having to check on if, and with which values, the `wp_localize_script` function is called.

To test check what's in the `WP_Dependencies::registered` property (it's `public`) after the code executed.
Get hold of the global scripts object `wp_localize_script` will use by means of `$GLOBALS['wp_scripts']`.

borkweb:

PR #1170, opened 2019-10-08
VR > Loader

juanfra:
🎫 https://central.tri.be/issues/135105
🎥 https://www.loom.com/share/e6518b7c3ef64688b4a86842aa8a8f66

### DIFF
--- a/src/resources/postcss/components/_loader.pcss
+++ b/src/resources/postcss/components/_loader.pcss
@@ -1,0 +1,93 @@
+.tribe-common {
+	/* -----------------------------------------------------------------------------
+	 *
+	 * Component: Loader
+	 *
+	 *
+	 * Example:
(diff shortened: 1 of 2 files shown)


src/resources/postcss/components/_loader.pcss
```diff
@@ -0,0 +1,93 @@
...
+		padding-top: calc(var(--spacer-11) * 3);
+
+		@media (--viewport-medium) {
+			padding-top: calc(var(--spacer-13) * 3 );
+		}
+	}
+
+	.tribe-common-c-loader__dot {
+		background-color: var(--color-accent-primary-background);
+		float: left;
```
paulmskim: I'd suggest moving away from floats as much as possible for positioning. There are specific use cases for floats, such as floating an image with text wrapping it for example, but positioning is not it's best use case. It's a very old use case for floats and given browser support for things such as flex, we should probably be using other options instead.

PR #1177, opened 2019-10-09
Add normalize function

Camwyn:
🎫 https://central.tri.be/issues/130974 (sorta)

Adds a function to "normalize" version numbers (no semver pre-release portion) for better comparison with QA `dev` zips.

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -689,6 +689,25 @@ if ( ! function_exists( 'tribe_get_class_instance' ) ) {
 	}
 }
 
+
+
+if ( ! function_exists( 'tribe_get_normalized_version' ) ) {
+	/**
+	 * Strips off anything that isn't a "standard" version character [1-0.].
+	 * Used in our version comaprisons - not the "get version" functions.
+	 *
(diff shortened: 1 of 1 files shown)


src/functions/utils.php
```diff
@@ -885,7 +904,10 @@ function tribe_installed_before( $class, $version ) {
 			return true;
 		}
 
-		return 0 > version_compare( $install_version, $version );
+		return 0 > version_compare(
```
borkweb: I believe this could potentially introduce an issue for users (particularly ones that installed an alpha or beta as their first version). This normalize function removes any awareness of pre-release versions and if there is a change that relies on checking versions to execute code added between a pre-release and a release, then that code would not execute.

Camwyn: hm...good point. 
I mean, that was kind of the point - but I was thinking from a QA perspective (there is a work-around). 
Pre-release versions are fairly rare here (although that may change). Suggestion retracted.

borkweb:
@borkweb is so incredibly brilliant.

PR #1179, opened 2019-10-10
Feature/135510 common base

paulmskim:
https://central.tri.be/issues/135510

Requires https://github.com/moderntribe/tribe-common/pull/1178 to be merged first.


paulmskim:
Note here: SVG icons and their use throughout TEC for skeleton v. full will be discussed with MattCR and Leah, will leave note here on results.

PR #1181, opened 2019-10-11
tribe_sanitize_deep didnt handle booleans correctly

bordoni:
_Ref:_ [C#135234](http://central.tri.be/issues/135234)
---
📹 http://i.bordoni.me/2194beb6994e

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -695,7 +695,6 @@ if ( ! function_exists( 'tribe_sanitize_deep' ) ) {
 	 */
 	function tribe_sanitize_deep( &$value ) {
 		if ( is_bool( $value ) ) {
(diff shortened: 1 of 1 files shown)


src/functions/utils.php
```diff
@@ -695,16 +694,20 @@ function tribe_get_request_vars( $refresh = false ) {
 	 *                    array.
 	 */
 	function tribe_sanitize_deep( &$value ) {
+		if ( is_bool( $value ) ) {
+			$value = filter_var( $value, FILTER_VALIDATE_BOOLEAN );
```
borkweb: This is less restrictive than the is_bool call so I don’t know how useful the filter_var is right here. I think you can just return here instead.

bordoni: That is true. haha

PR #1200, opened 2019-11-06
feat(Utils/Query.php) add Query utils

lucatume:
Ticket: https://central.tri.be/issues/136622

This PR adds the `Tribe\Utils\Query` class.
The class now provides just one method: `Query::for_posts( array $posts )`.
The method builds and returns a `WP_Query` object that will return the posts passed to it behaving like a query, but not running a query.

### DIFF
--- a/src/Tribe/Utils/Query.php
+++ b/src/Tribe/Utils/Query.php
@@ -22,7 +22,7 @@ class Query {
 	 * Builds a new `WP_Query` object and sets the post, and accessory flags, on it.
 	 *
 	 * The query is built to yield to run a query that will yield no result and to have a `request` property that
-	 * will never yield results; calls on the `WP_Query::get_posts` method are filtered to return always the post set.
+	 * will never yield results; calls on the `WP_Query::get_posts` method are filtered to always return the post set.
 	 * Queries built by this function can be spotted by looking for the `tribe_mock_query` property.
 	 *
 	 * @since TBD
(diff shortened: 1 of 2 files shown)


src/Tribe/Utils/Query.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @package Tribe\Utils
+ */
+class Query {
+
+	/**
+	 * Builds a new `WP_Query` object and sets the post, and accessory flags, on it.
+	 *
+	 * The query is built to yield to run a query that will yield no result and to have a `request` property that
+	 * will never yield results; calls on the `WP_Query::get_posts` method are filtered to return always the post set.
```
borkweb:
```suggestion
	 * will never yield results; calls on the `WP_Query::get_posts` method are filtered to always return the post set.
```

src/Tribe/Utils/Query.php
```diff
@@ -0,0 +1,73 @@
...
+		$query->post             = reset( $posts );
+		$query->query            = [ 'p' => 0 ];
+		$query->tribe_mock_query = true;
+		global $wpdb;
+		// Use a query that will never yield results.
+		$query->request = "SELECT ID FROM {$wpdb->posts} WHERE 1=0";
+
+		// Return the same set of posts on each method requiring posts.
+		$filter_posts_pre_query = static function ( $the_posts, $the_query ) use ( $posts, $query )
+		{
```
borkweb: Move this brace onto the previous line.

PR #1218, opened 2019-11-25
Correct missing ticket block when switching 

Camwyn:
...from blocks to classic editor.

🎫 https://central.tri.be/issues/131493

📸 http://p.tri.be/5VrtPI

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.10.1] TBD =
+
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.10.1] TBD =
+
+* Fix - correct missing block when swithcing from blocks to classic editor. [131493]
```
```suggestion
* Fix - Correct missing block when switching from blocks to classic editor. [131493]
```

PR #1222, opened 2019-11-26
 feat(attendees API): Add a new function to filter by event meta keys.

Camwyn:
Specific use case is for not showing attendees that are for events where "Show attendees" is
unchecked". More generally, allows filtering attendee results via an event meta key->value pair.

🎫 https://central.tri.be/issues/133333

📷 http://p.tri.be/8hslf3 (as admin, 72 is "private")
📸 http://p.tri.be/gWpFmE (as non-admin)


src/Tribe/Repository.php
```diff
@@ -1609,6 +1609,60 @@ public function where_meta_related_by( $meta_keys, $compare, $field = null, $val
 
 		return $this;
 	}
+	/**
+	 * Filters the query to only return posts that are related, via a meta key, to posts
+	 * that satisfy a condition.
+	 *
```
bordoni: Add `@since`

src/Tribe/Repository.php
```diff
@@ -1609,6 +1609,60 @@ public function where_meta_related_by( $meta_keys, $compare, $field = null, $val
...
+	 *                                  to another post type.
+	 * @param string       $compare     The SQL comparison operator.
+	 * @param string       $meta_field  One (a column in the `postmeta` table) that should match
+	 *                                  the comparison criteria; required if the comparison operator is not `EXISTS` or
+	 *                                  `NOT EXISTS`.
+	 * @param string|array $meta_values One or more values the post field(s) should be compared to;
+	 *                                  required if the comparison operator is not `EXISTS` or `NOT EXISTS`.
+	 *
+	 * @return $this
+	 * @throws Tribe__Repository__Usage_Error If the comparison operator requires
```
bordoni: I think that should come before the `@params`, right after the `@since`.

Camwyn: Was following the format of the function above it - but happy to change

src/Tribe/Repository.php
```diff
@@ -1609,6 +1609,63 @@ public function where_meta_related_by( $meta_keys, $compare, $field = null, $val
...
+	 * @return $this
+	 */
+	public function where_meta_related_by_meta( $meta_keys, $compare, $meta_field = null, $meta_values = null ) {
+		$meta_keys = Tribe__Utils__Array::list_to_array( $meta_keys );
+
+		if ( ! in_array( $compare, array( 'EXISTS', 'NOT EXISTS' ), true ) ) {
+			if ( empty( $meta_field ) || empty( $meta_values ) ) {
+				throw Tribe__Repository__Usage_Error::because_this_comparison_operator_requires_fields_and_values( $meta_keys, $compare, $this );
+			}
+			$meta_field = esc_sql( $meta_field );
```
sc0ttkclark: Only escapes `$meta_field` in this conditional but further down for `EXISTS` it uses the `$meta_field` var unescaped

src/Tribe/Repository.php
```diff
@@ -1609,6 +1609,63 @@ public function where_meta_related_by( $meta_keys, $compare, $field = null, $val
...
+		$this->filter_query->join( "LEFT JOIN {$wpdb->postmeta} {$pmm} ON {$pm}.meta_value = {$pmm}.post_id" );
+
+		$keys_in = $this->prepare_interval( $meta_keys );
+
+		if ( 'EXISTS' === $compare ) {
+			$this->filter_query->where( "{$pm}.meta_key IN {$keys_in} AND {$pmm}.meta_key = '{$meta_field}' AND {$pmm}.meta_value IS NOT NULL" );
+		} elseif ( 'NOT EXISTS' === $compare ) {
+			$this->filter_query->where( "{$pm}.meta_key IN {$keys_in} AND {$pmm}.meta_key = '{$meta_field}' AND {$pmm}.meta_value IS NULL" );
+		} else {
+			if ( in_array( $compare, self::$multi_value_keys, true ) ) {
```
sc0ttkclark:
```suggestion
			if ( in_array( $compare, static::$multi_value_keys, true ) ) {
```

Camwyn: Not sure I understand the _why_ of this one? (and should it be changed for the function above as well?)

Camwyn: @sc0ttkclark I've committed the suggestion - should we do the same for the other function as well?
Thanks @bordoni for the explanation!

sc0ttkclark: It ought to be changed there too, thanks!

sc0ttkclark:
This really needs a test, can we try and get a test made for it Friday (tomorrow) in time for cut off?

Camwyn:
Keeping branch up to date with release - test is not done yet

Camwyn:
@sc0ttkclark - added a read test.

sc0ttkclark:
@lucatume Review this when you can when you start your day. Thanks!

sc0ttkclark:
Due to the needs of https://github.com/moderntribe/event-tickets/pull/1491 we should update this so it can support a comparison operator + EXISTS check.

Example: meta_value = 1 OR meta_value IS NULL

sc0ttkclark:
I've updated this PR with the ability to lookup posts that have a specific related meta value OR that related meta IS NULL, for example:

`Books that have reviews with a field set to 'XYZ' OR no field set at all (no meta row for that key)`

Expanded the tests to cover even more functionality and cleaned up the test logic in that method so it's easier to read.

src/Tribe/Repository.php
```diff
@@ -1562,7 +1562,7 @@ public function set_formatter( Tribe__Repository__Formatter_Interface $formatter
 	 * Filters the query to only return posts that are related, via a meta key, to posts
 	 * that satisfy a condition.
 	 *
-	 * @param string|array $meta_keys One ore more `meta_keys` relating the queried post type(s)
+	 * @param string|array $meta_keys One or more `meta_keys` relating the queried post type(s)
 	 *                                to another post type.
 	 * @param string       $compare   The SQL comparison operator.
 	 * @param string       $field     One (a column in the `posts` table) that should match
```
lucatume: What about flipping the position of the second and third parameters?

This way the signature would become:

```php
public function where_meta_related_by( $meta_keys, $field = null, $compare = 'EXISTS', $values = null )
```

Allowing for more readable calls like:

```php
$r->where_meta_related_by( '_review', '_short_line', '=', 'Would read it again!' );

// It has a review.
$r->where_meta_related_by( '_review' )

// Find books w/ reviews with a short line.
$r->where_meta_related_by( '_review', '_short_line', 'EXISTS' );

// Find books w reviews w/o a short line.
$r->where_meta_related_by( '_review', '_short_line', 'NOT EXISTS' );

// Find books w/o reviews.
$r->where_meta_related_by( '_review', null, 'NOT EXISTS' );
```

The last example is the one that would become less "elegant" yet, I think, would keep being readable. I think the change would make the method more readable in its most common use.

This is, but, a personal opinion and I would respect a decision in the existing direction.

sc0ttkclark: I think you meant this comment about `where_meta_related_by_meta()`, not `where_meta_related_by()`, right?

We kept the same ordering for `where_meta_related_by_meta()` so it wouldn't differ too much.

PR #1223, opened 2019-11-26
Look for Templates in the correct folder inside of themes

bordoni:
After a call with Victor I realized that we were looking into the wrong folder inside of the Users theme, so this Pull Request aims to fix that in a simple way.

### DIFF
--- a/src/Tribe/Template.php
+++ b/src/Tribe/Template.php
@@ -345,16 +345,16 @@ class Tribe__Template {
 	 *
 	 * @since  TBD
 	 *
-	 * @return array
+	 * @return array The base folders we look for templates in the origin plugin.
 	 */
 	public function get_template_origin_base_folder() {
 		/**
-		 * Allows filtering of the base path for templates
+		 * Allows filtering of the base path for templates.
 		 *
 		 * @since TBD
 		 *
 		 * @param array  $namespace Which is the base folder we will look for files in the plugin.
-		 * @param self   $template  Current instance of the Tribe__Template
+		 * @param self   $template  Current instance of the Tribe__Template.
 		 */
 		return apply_filters( 'tribe_template_origin_base_folder', $this->template_origin_base_folder, $this );
 	}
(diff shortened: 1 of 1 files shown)


src/Tribe/Template.php
```diff
@@ -341,9 +369,17 @@ protected function get_template_public_namespace() {
 	 * @return string
 	 */
 	protected function get_template_public_path( $base ) {
+
 		// Craft the plugin Path
 		$path = array_merge( (array) $base, (array) $this->get_template_public_namespace() );
 
+		// Pick up if the folder needs to be aded to the public tempalte path
```
zbtirrell: minor comment typo

src/Tribe/Template.php
```diff
@@ -331,6 +340,25 @@ protected function get_template_public_namespace() {
 		return apply_filters( 'tribe_template_public_namespace', $namespace, $this );
 	}
 
+	/**
+	 * Fetches which base folder we look for templates in the origin plugin.
+	 *
+	 * @since  TBD
+	 *
+	 * @return array
```
sc0ttkclark:
```suggestion
	 * @return array The base folders we look for templates in the origin plugin.
```

src/Tribe/Template.php
```diff
@@ -331,6 +340,25 @@ protected function get_template_public_namespace() {
...
+	 */
+	public function get_template_origin_base_folder() {
+		/**
+		 * Allows filtering of the base path for templates
```
sc0ttkclark:
```suggestion
		 * Allows filtering of the base path for templates.
```

src/Tribe/Template.php
```diff
@@ -331,6 +340,25 @@ protected function get_template_public_namespace() {
...
+		 *
+		 * @since TBD
+		 *
+		 * @param array  $namespace Which is the base folder we will look for files in the plugin.
+		 * @param self   $template  Current instance of the Tribe__Template
```
sc0ttkclark:
```suggestion
		 * @param self   $template  Current instance of the Tribe__Template.
```

src/Tribe/Template.php
sc0ttkclark:
```suggestion
		// Pick up if the folder needs to be aded to the public template path.
```

PR #1231, opened 2019-12-08
Include Translation methods to the DateTime and DateTimeImmutable

bordoni:
_Ref:_ [C#138196](http://central.tri.be/issues/138196)
---
📹 https://i.bordoni.me/YEuddmrP

Currently, we use the `DateTime` and `DateTimeImmutable` on our class handling when using `Date_Utils::build_date_object()`, which creates problems when translations are needed on the other end, because the code looks messy since you need to pass all strings that need to be translated to the `date_i18n` or `wp_date` functions.

The goal of this is to include two classes that will extend the `DateTime` and `DateTimeImmutable` so we can have a method for the translation to happen.

Example of Usage:
```php
Dates::build_date_object( $selected_start_date_value )->format_i18n( $date_formats->compact );
```


src/Tribe/Utils/Date_I18n.php
```diff
@@ -0,0 +1,21 @@
...
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string $string Format to be used in the translation.
+	 *
+	 * @return string         Translated date.
+	 */
+	public function format_i18n( $string ) {
+		$formatted = $this->format( $string );
+		$unix_date =$this->format( 'U' );
```
sc0ttkclark:
```suggestion
		$unix_date = $this->format( 'U' );
```

src/Tribe/Utils/Date_I18n_Immutable.php
```diff
@@ -0,0 +1,21 @@
...
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string $string Format to be used in the translation.
+	 *
+	 * @return string         Translated date.
+	 */
+	public function format_i18n( $string ) {
+		$formatted = $this->format( $string );
+		$unix_date =$this->format( 'U' );
```
sc0ttkclark:
```suggestion
		$unix_date = $this->format( 'U' );
```

src/Tribe/Date_Utils.php
```diff
@@ -1386,7 +1389,7 @@ static function immutable( $datetime = 'now', $timezone = null, $with_fallback =
 			}
 
 			if ( $datetime instanceof DateTime ) {
-				return DateTimeImmutable::createFromMutable( $datetime );
+				return ( new Date_I18n_Immutable() )->setTimestamp( DateTimeImmutable::createFromMutable( $datetime )->getTimestamp() );
```
lucatume: The conversion to timestamp is one that will drop the timezone information defaulting on the use of the system one.
I would rather implement a `Date_I18n_Immutable::createFromMutable` method (in `camelCase` to override the default PHP method and keep the API consistent) where we're never conflating the time information and the timezone information into a timestamp.

src/Tribe/Date_Utils.php
```diff
@@ -1402,7 +1405,7 @@ static function immutable( $datetime = 'now', $timezone = null, $with_fallback =
 				return $cached;
 			}
 
-			$immutable = DateTimeImmutable::createFromMutable( $mutable );
+			$immutable = ( new Date_I18n_Immutable() )->setTimestamp( $mutable->getTimestamp() );
```
lucatume: The same comment about `createFromMutable` applies here.

src/Tribe/Date_Utils.php
bordoni: Nice I like that.

bordoni:
Timezone failing due to the time in which this ran.

src/Tribe/Utils/Date_I18n.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
lucatume: file header is missing

src/Tribe/Utils/Date_I18n.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Tribe\Utils;
+
+use Tribe__Date_Utils as Dates;
+use DateTime;
+use DateTimeImmutable;
+
+class Date_I18n extends DateTime {
```
lucatume: class header is missing

src/Tribe/Utils/Date_I18n_Immutable.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
lucatume: File header is missing

src/Tribe/Utils/Date_I18n_Immutable.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Tribe\Utils;
+
+use Tribe__Date_Utils as Dates;
+use DateTime;
+use DateTimeImmutable;
+
+class Date_I18n_Immutable extends DateTimeImmutable {
```
lucatume: Class header is missing

src/Tribe/Utils/Date_I18n_Immutable.php
```diff
@@ -0,0 +1,32 @@
...
+	/**
+	 * Returns a translated string usign the params from this Immutable DateTime instance.
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string $string Format to be used in the translation.
+	 *
+	 * @return string         Translated date.
+	 */
+	public function format_i18n( $string ) {
```
lucatume: Rename `$string` to `format`.

src/Tribe/Utils/Date_I18n.php
```diff
@@ -0,0 +1,32 @@
...
+	/**
+	 * Returns a translated string usign the params from this DateTime instance.
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string $string Format to be used in the translation.
+	 *
+	 * @return string         Translated date.
+	 */
+	public function format_i18n( $string ) {
```
lucatume: Rename `$string` to `$format`.

lucatume: Even better, to stick w/ what core uses, `$dateformatstring` or `$date_format`.

src/Tribe/Utils/Date_I18n.php
```diff
@@ -0,0 +1,32 @@
...
+		$unix_date  = $this->format( 'U' );
```
lucatume: Codex says the `$unix_date` is "Unix timestamp, offset by the desired timezone."
Is the timezone offset taken into account here?

bordoni: Yeah it takes in cosideration.

PR #1234, opened 2019-12-10
fix(Date_Utils) correctly use timezone when building from timestamp

lucatume:
Ticket: n/a

This PR fixes an issue introduced during the work done in `Date_Utils` where, when using date timestamps to build the date, the output would not be the one we expect.


sc0ttkclark:
I'd like to test this fix against the work we did for ET to see if it conflicts or requires more work on our side there.

PR #1237, opened 2019-12-11
Set the datepicker (compact) default to a more friendly setting

borkweb:
This sets the default datepicker format to a friendly US-centric setting.

🎥 http://p.tri.be/wKeyp9

Related:
* https://github.com/moderntribe/the-events-calendar/pull/2916
* https://github.com/moderntribe/events-pro/pull/1405

[see/136789](https://central.tri.be/issues/136789)

### DIFF
--- a/tests/wpunit/Tribe/Utils/Date_I18n_ImmutableTest.php
+++ b/tests/wpunit/Tribe/Utils/Date_I18n_ImmutableTest.php
@@ -20,10 +20,10 @@ class Date_I18n_Immutable_Test extends WPTestCase {
 
 	public function data_dates_and_timezones() {
 		return [
-			'America/Sao_Paulo' => [ '2018-02-01 18:00:00', 'America/Sao_Paulo', '2018-02-01 16:00:00' ],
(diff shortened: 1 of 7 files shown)


tests/wpunit/Tribe/Utils/Date_I18n_ImmutableTest.php
```diff
@@ -34,9 +34,8 @@ public function data_dates_and_timezones() {
 	 */
 	public function it_should_properly_include_timezones( $datetime, $timezone, $expected ) {
```
barryhughes: I know this test pre-dates the change you made in this PR, but I'll comment anyway: coming into this raw (I haven't really looked at this body of work before) I took a moment before I grokked what we're testing for (that we expect the result in UTC). The name of the test doesn't _quite_ express that for me. 

Anyway, that's a bit of a subjective comment perhaps.

borkweb: That is totally a worthwhile mention. I've updated the tests to be named: `it_should_properly_convert_to_utc_from_provided_timezone`

PR #1241, opened 2019-12-17
fix(Collection_Trait.php) allow for pre (un)serialize methods

lucatume:
Ticket: https://central.tri.be/issues/138249 

This PR adds support, in the `Collection_Trait` for custom `serialize` and `unserialize` handling by classes using the trait.

### DIFF
--- a/src/Tribe/Utils/Collection_Trait.php
+++ b/src/Tribe/Utils/Collection_Trait.php
@@ -152,14 +152,27 @@ trait Collection_Trait {
 	 * {@inheritDoc}
 	 */
 	public function serialize() {
-		return serialize( $this->all() );
+		$to_serialize = $this->all();
+
+		if ( method_exists( $this, 'before_serialize' ) ) {
+			$to_serialize = $this->before_serialize( $this->all() );
+		}
+
+		return serialize( $to_serialize );
 	}
 
 	/**
 	 * {@inheritDoc}
 	 */
 	public function unserialize( $serialized ) {
-		$this->items = unserialize( $serialized );
+		$to_unserialize = $serialized;
+
+		if ( method_exists( $this, 'custom_unserialize' ) ) {
+			$this->items = $this->custom_unserialize( $to_unserialize );
+			return;
+		}
+
+		$this->items = unserialize( $to_unserialize );
 	}
 
 	/**


src/Tribe/Utils/Collection_Trait.php
```diff
@@ -152,14 +152,27 @@ public function count() {
...
 	public function unserialize( $serialized ) {
-		$this->items = unserialize( $serialized );
+		$to_unserialize = $serialized;
+
+		if ( method_exists( $this, 'custom_unserialize' ) ) {
+			$this->items = $this->custom_unserialize( $to_unserialize );
+			return;
+		}
+
+		$this->items = unserialize( $to_unserialize );
```
barryhughes: Got a two part question (...I'm unfamiliar with where these elements are going to be used, so this may not be a concern/may not be valid, but will ask anyway).

(1/2) For added safety on runtimes that support it, should we specify `allowed_classes` via `unserialize()`'s optional 2nd param? On early runtimes this won't do anything, but on more recent PHPs we'd be able to stop direct wakeup of arbitrary classes (which can be a security risk).

src/Tribe/Utils/Collection_Trait.php
```diff
@@ -152,14 +152,27 @@ public function count() {
...
 	 */
 	public function serialize() {
-		return serialize( $this->all() );
+		$to_serialize = $this->all();
+
+		if ( method_exists( $this, 'before_serialize' ) ) {
+			$to_serialize = $this->before_serialize( $this->all() );
+		}
+
+		return serialize( $to_serialize );
```
barryhughes: (2/2) If we _do_ need to perform wakeups and have to support direct serialization/unserialization of arbitrary classes etc this won't be tenable but, following my earlier question, might an alternative serialization technique (to JSON, for example) be worthwhile on the basis of improved security (noting that the mitigation given in the first part of this question won't fly on PHP 5.6).

src/Tribe/Utils/Collection_Trait.php
lucatume: The problem is that we cannot know the classes used in a general use collection implementation before-hand.

To keep track of those we would need to pass them at constructor time.
But this is a trait and we cannot control the `__construct` method. If special care should be taken to `unserialize`, then I think it should fall into classes using the trait, not the trait.

src/Tribe/Utils/Collection_Trait.php
lucatume: The problem with JSON is that the information about the serialized class is lost.

This is the result of a call to `json_encode` on a post:

```json
{"ID":381,"post_author":"1","post_date":"2019-12-17 08:51:58","post_date_gmt":"2019-12-17 07:51:58","post_content":"Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.","post_title":"CoworkingRepublic","post_excerpt":"","post_status":"publish","comment_status":"closed","ping_status":"closed","post_password":"","post_name":"coworkingrepublic","to_ping":"","pinged":"","post_modified":"2019-12-17 08:51:58","post_modified_gmt":"2019-12-17 07:51:58","post_content_filtered":"","post_parent":0,"guid":"http:\/\/mtvr.local\/?post_type=tribe_venue&#038;p=381","menu_order":0,"post_type":"tribe_venue","post_mime_type":"","comment_count":"0","filter":"raw"}
```

Furthermore, we attach objects to post properties, with `serialize` and `unserialize` we either get those implementations for free, or we can have them handled *if* we need to.

I understand the security concern here, yet we're reading from cached values, not user inputs.
I fully back the "trust no-one" policy, but I think we're pretty safe here.

lucatume: Losing the class information at serialization time means we, now, would have to pass that information in the class `__construct` method.
Which would make the use of the generic collection not so convenient.

lucatume:
For an example usage: https://github.com/moderntribe/the-events-calendar/pull/2933 -- here we're also handling unserialization.

barryhughes:
For sure. JSON was just an example of an alternative format, not a recommendation. But, cool! Just wanted to raise the question. This otherwise looks shabba.

PR #1244, opened 2019-12-19
If we tribe_catch_and_throw, we need to restore error handler in catch

borkweb:
When we use our `tribe_catch_and_throw()` function, we need to be sure to restore the proper error handler upon catching the exception so we don't take over error handling for other errors in and out of our control.

The particular error in question was related to PHP 7.3 alongside WP 4.9.12 (and lower) where a notice is thrown within the `class-wp-comment-query.php` class file due to PHP 7.3's handling of `compact()`. We had an overly aggressive error handler that was never relinquishing control once it caught a date error. This fixes that problem.

Screen Shot 2019-12-18 at 8 09 14 PM

see/138785

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -1245,6 +1245,9 @@ if ( ! class_exists( 'Tribe__Date_Utils' ) ) {
 				$date = new Date_I18n( $datetime, $timezone_object );
 				restore_error_handler();
 			} catch ( Exception $e ) {
+				// if we encounter an error, we need to restore after catching
+				restore_error_handler();
+
 				if ( $timezone_object === null ) {
 					$timezone_object = Tribe__Timezones::build_timezone_object( $timezone );
 				}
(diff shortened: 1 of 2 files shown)


src/Tribe/Date_Utils.php
```diff
@@ -1245,6 +1245,9 @@ public static function build_date_object( $datetime = 'now', $timezone = null, $
 				$date = new Date_I18n( $datetime, $timezone_object );
 				restore_error_handler();
 			} catch ( Exception $e ) {
+				// if we encounter an error, we need to restore after catching
```
sc0ttkclark:
```suggestion
				// If we encounter an error, we need to restore after catching.
```

borkweb:
Those lines were added in this sprint, so we should be good on the changelog side of things.

borkweb:
Test errors are the classic time-of-day-tests-are-executed errors.

sc0ttkclark:
All good

lucatume:
This is a  great addition!

lucatume:
I meant... a great catch.

PR #1246, opened 2019-12-19
Replace immutable/mutable date handling changes and resolve test issues

borkweb:
This puts back the datetime fixes that bordoni put in place and fixes the tests so that their values are appropriately validated.

:movie_camera: http://p.tri.be/gV3Jcy

Related: https://github.com/moderntribe/the-events-calendar/pull/2945

see/138795

### DIFF
--- a/src/Tribe/Utils/Date_I18n.php
+++ b/src/Tribe/Utils/Date_I18n.php
@@ -31,7 +31,7 @@ class Date_I18n extends DateTime {
 	}
 
 	/**
-	 * Returns a translated string usign the params from this DateTime instance.
+	 * Returns a translated string using the params from this DateTime instance.
 	 *
 	 * @since  TBD
 	 *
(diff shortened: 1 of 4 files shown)


sc0ttkclark:
There’s a lot of movement on this part of the code in regards to U and timestamp handling. I recommend caution and further review on this PR. Are there any angles we are missing here? Timestamps and time zones and PHP DateTime handling has bit me in the past for Event Tickets.

borkweb:
I think those thoughts are _generally_ valid and further review is worthwhile.

It should be stated that these classes and tests are **all new as of this sprint** and were created by @bordoni, modified by @bordoni (which caused the tests to fail), then modified by me (which caused the date tests to pass, but some snapshot tests to fail–though I generated new snapshots for them not understanding what I was seeing), and now modified by me again with both tests and _real_ snapshots passing.

lucatume:
I've added some further test methods to state, in clear, what the "weird" default behavior of `DateTime` is: *when built from timestamp the DateTimezone will be ignored and the DateTimezone of the built object will be set to the GMT+0 one*.

This is quite important as the timezone can only be applied later, to localize a date built from timestamp, calling `DateTime::setTimezone`.

lucatume:
I'm not merging this one to wait from feedback from @borkweb on this.

borkweb:
Woo! Those changes look great, @lucatume!

PR #1255, opened 2020-01-10
VR > Template cache improvements fix.

juanfra:
🎫 https://moderntribe.atlassian.net/browse/TEC-3076
🎥 https://www.loom.com/share/09eb0102421d4838a04ee2ca5ec44b4f

The template method is also being used from the ET backend to display the tickets/rsvp sections within the event/post/page editor. Using the last item of the array as the cache identifier for those cases.

### DIFF
--- a/src/Tribe/Template.php
+++ b/src/Tribe/Template.php
@@ -507,7 +507,7 @@ class Tribe__Template {
 		static $template_names = [];
 
 		// Key we'll use for in-memory caching of expensive operations.
-		$cache_name_key = is_array( $name ) ? end( $name ) : $name;
+		$cache_name_key = is_array( $name ) ? implode( '/', $name ) : $name;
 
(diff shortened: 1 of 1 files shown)


src/Tribe/Template.php
```diff
@@ -507,7 +507,7 @@ public function template( $name, $context = array(), $echo = true ) {
 		static $template_names = [];
 
 		// Key we'll use for in-memory caching of expensive operations.
-		$cache_name_key = $name;
+		$cache_name_key = is_array( $name ) ? end( $name ) : $name;
```
bordoni:
```suggestion
		$cache_name_key = is_array( $name ) ? implode( '/', $name ) : $name;
```

juanfra: It's called a few times and the one that's different is the last item, that's why I'm grabbing that one. We can go either way.

PR #1268, opened 2020-01-23
Folder list with Namespacing correct

bordoni:
When including files from themes the `Tribe__Templates` class was looking into the incorrect file path depending on how the class was instantiated.

If we wanted to include an Event Tickets file from a class that was using the Origin as The Events Calendar, it would point to `[your-theme]/events/v2/event.php` instead of  `[your-theme]/tickets/v2/event.php`.

📹 https://i.bordoni.me/Koue8qN5

### DIFF
--- a/src/Tribe/Template.php
+++ b/src/Tribe/Template.php
@@ -148,7 +148,7 @@ class Tribe__Template {
 	 *
 	 * @since 4.11.0
 	 *
-	 * @return array
+	 * @return array Current folder we are looking for templates.
 	 */
 	public function get_template_folder() {
 		return $this->folder;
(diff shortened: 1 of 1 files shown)


src/Tribe/Template.php
```diff
@@ -143,6 +143,17 @@ public function set_template_folder( $folder = null ) {
 		return $this;
 	}
 
+	/**
+	 * Returns the array for which folder this template instance is looking into.
+	 *
+	 * @since 4.11.0
+	 *
+	 * @return array
```
lucatume: Missing comment for the `@return`.

src/Tribe/Template.php
```diff
@@ -369,18 +382,20 @@ public function get_template_origin_base_folder() {
...
 	 *
 	 * @since  4.7.20
+	 * @since  4.11.0 Added the param $namespace.
 	 *
-	 * @param  mixed  $base  Base path to look into
+	 * @param  mixed  $base      Base path to look into.
+	 * @param  string $namespace Adds the plugin namespace to the path returned.
 	 *
-	 * @return string
+	 * @return string  Returns the public path for a given base.
```
lucatume:
```suggestion
	 * @return string  The public path for a given base.˙˙
```

PR #1280, opened 2020-02-04
Add a helper method for fetching the datepickerFormat index

borkweb:
This adds a filter for the fetching of the datepickerFormat while providing the expected default if settings has not yet been saved.

Related: https://github.com/moderntribe/the-events-calendar/pull/3046

:movie_camera:
* Event Edit: http://p.tri.be/NHqoXF
* Aggregator: http://p.tri.be/VkkDoC
* Calendar View: http://p.tri.be/ewgdS7

:ticket: [TEC-3229]

[TEC-3229]: https://moderntribe.atlassian.net/browse/TEC-3229

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -26,10 +26,37 @@ if ( ! class_exists( 'Tribe__Date_Utils' ) ) {
 		const DBTIMEFORMAT          = 'H:i:s';
 		const DBYEARMONTHTIMEFORMAT = 'Y-m';
 
+		/**
+		 * Default datepicker format index.
+		 *
+		 * @since 4.11.0.1
+		 *
+		 * @var int
+		 */
+		private static $default_datepicker_format_index = 1;
+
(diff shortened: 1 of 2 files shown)


src/Tribe/Date_Utils.php
```diff
@@ -99,7 +126,7 @@ public static function datepicker_formats( $translate = null ) {
 				return $formats;
 			}
 
-			return isset( $formats[ $translate ] ) ? $formats[ $translate ] : $formats[1];
+			return isset( $formats[ $translate ] ) ? $formats[ $translate ] : $formats[ static::get_datepicker_format_index() ];
```
sc0ttkclark: Is there any fault tolerance for the result of this method? What happens if the format isn't in the `$formats` array?

borkweb: No, not really. But That isn't a part of this hotfix. This is just replacing the number 1 with a static variable.

sc0ttkclark: ✅

PR #1287, opened 2020-02-05
AR Modal > Add theme compatibility for overlaps.

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ETP-156
🎥 https://www.loom.com/share/346c2e6a7dd142588e6830060b2affa5


juanfra:
Thanks for the review @sc0ttkclark! How are we handling the version numbers for G&B and syncing? Under which version should I put the readme?

readme.txt
```diff
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.11.1] TBD =
```
```suggestion
= [4.11.2] TBD =
```

src/resources/postcss/dialog.pcss
```diff
@@ -184,4 +184,38 @@ Adding these here for now until common-styles are live
...
 	}
 
+	/* -------------------------------------------------------------------------
+	 * Theme Overrides - Avada
+	 * ------------------------------------------------------------------------- */
+
+	 .tribe-theme-avada & {
+
+		div.tribe-dialog {
+			z-index: 99999;
```
bordoni: Can we get a variable for that? Possibly something related to theme compatibility.

src/resources/postcss/dialog.pcss
```diff
@@ -184,4 +184,38 @@ Adding these here for now until common-styles are live
...
+	}
+
+	/* -------------------------------------------------------------------------
+	 * Theme Overrides - Divi
+	 * ------------------------------------------------------------------------- */
+
+	 .tribe-theme-divi & {
+
+		div.tribe-dialog {
+			z-index: 99999;
```
bordoni: Can we get a variable for that? Possibly something related to theme compatibility.

src/resources/postcss/dialog.pcss
```diff
@@ -184,4 +184,38 @@ Adding these here for now until common-styles are live
...
+
+	/* -------------------------------------------------------------------------
+	 * Theme Overrides - Enfold
+	 * ------------------------------------------------------------------------- */
+
+	#top.tribe-theme-enfold &,
+	#top .main_color & {
+
+		div.tribe-dialog {
+			z-index: 99999;
```
bordoni: Can we get a variable for that? Possibly something related to theme compatibility.

src/resources/postcss/dialog.pcss
juanfra: Thanks! We normally don't use variables in the pcss for the theme compatibility, because they're normally overrides. Do you feel we should do an exception here and having different from the rest?

Camwyn: Knowing that we've had some issues with variable replacement in the build (for compatibility, as done in TEC) it should be fine to skip the variables here.

That would be a _great_ thing for you to look into though @juanfra - see if you can figure out what the problem is (separate from this)?

bordoni: I really would love to avoid arbitrary numbers. What were the problems with the building?

Since now Tribe Common Styles is being used on Tickets too we can import variables from there at any time correct?

Maybe a variable for Theme compatibility overall? So if in the future we need to bump it, it applies to all themes.

juanfra: In the end it's a `z-index` race. What we're doing here is making sure that we have a higher one for the themes we provide compatibility with. Given that they use arbitrary numbers, we have to use arbitrary numbers to add their compatibility. These are "special cases", and we don't normally set variables for these special cases. There's no silver bullet for theme compatibility, each theme can have whatever `z-index` they want for their header, footer, or the element they want. We ended up giving compatibility with the same themes we have for V2. And if somebody else comes with something, then we can provide a CSS snippet.

PR #1290, opened 2020-02-11
Add is_plain_permalink to Rewrite class

bordoni:
🎟 [EC-3120]
---
We needed a check that was cached for plain permalinks.

Related to https://github.com/moderntribe/the-events-calendar/pull/3061

PR #1292, opened 2020-02-11
Add `plain_permalink` to Context and rewrite.

bordoni:
Duplicate of #1290 that I reverted because I merged too early.

### DIFF
--- a/src/Tribe/Context/locations.php
+++ b/src/Tribe/Context/locations.php
@@ -9,13 +9,27 @@
  *
  * @since 4.9.11
  */
-
 return [
 	'post_id' => [
 		'read' => [
 			Tribe__Context::FUNC => static function () {
 				return get_the_ID();
 			}
+		],
+	],
+	'plain_permalink' => [
+		'read'  => [
+			Tribe__Context::FUNC => static function () {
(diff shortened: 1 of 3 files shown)


src/Tribe/Context/locations.php
```diff
@@ -18,6 +17,21 @@
 			}
 		],
 	],
+	'plain_permalink' => [
+		'read'  => [
+			Tribe__Context::FUNC => static function () {
```
lucatume: Since we're dealing with an option that will be fetched, by WordPress, on each request, does it make sense to cache?
I think the code could be simplified to this:
```php
'permalink_structure' => [
	'read' => [
		Tribe__Context::OPTION => [ 'permalink_structure' ],
	],
],
'plain_permalink' => [
	'read' => [
		Tribe__Context::LOCATION_FUNC => [
			'permalink_structure',
			static function( $struct ){
				return empty( $struct );
			},
		 ],
	],
],
```

This would add the `permalink_structure` and the `plain_permalink` locations and make the latter depend on the first.

bordoni: Oh I like that quite a bit.

PR #1293, opened 2020-02-11
fix (css): correct invalid attribute.

Camwyn:
[BTRIA-171] 

No visible difference, so no screenshot. But valid CSS! 🎈 

[BTRIA-171]: https://moderntribe.atlassian.net/browse/BTRIA-171


bordoni:
This is a case of no QA required unless you think otherwise @Camwyn, can you make sure @neillmcshea can move it into the release? I want to avoid these PRs hanging.

PR #1294, opened 2020-02-13
Add a default value for the datepicker format option. 

helgatheviking:
Without this dates cannot be parsed and do NOT save if the option value is not in the DB.

This is what was happening to me when trying to set the start/end sale dates.
https://share.getcloudapp.com/qGuo0XRY

I'm unsure why the option is not in the DB.

### DIFF
--- a/src/Tribe/Date_Utils.php
+++ b/src/Tribe/Date_Utils.php
@@ -69,7 +69,7 @@ if ( ! class_exists( 'Tribe__Date_Utils' ) ) {
 		 */
 		public static function maybe_format_from_datepicker( $date, $datepicker = null ) {
 			if ( ! is_numeric( $datepicker ) ) {
-				$datepicker = tribe_get_option( 'datepickerFormat' );
+				$datepicker = tribe_get_option( 'datepickerFormat', self::get_datepicker_format_index() );
 			}
 
 			if ( is_numeric( $datepicker ) ) {


lucatume:
@helgatheviking thanks for the PR.

Since we're in the immediate aftermath of a release I will not merge it right away and the PR will go into the next release.

In regards to the issue itself: we've fixed a similar issue in the latest release, but this will make the code more robust; once again: thanks.

I will keep you posted, thanks again,

Luca

lucatume:
@helgatheviking I've re-based your PR and included it in a new one that is merely a wrapper to have tests correctly run and add some required files.

helgatheviking:
@lucatume sounds good to me! I was informed that my [bridge plugin](https://github.com/helgatheviking/woocommerce-name-your-price-event-tickets/) for adding WooCommerce Name Your Price support to tickets (for a pay what you want ticket) no longer works. Looks like there have been some templating changes lately. I was trying to work on that but couldn't get past this. 

Is there an install routine that should have set that option? Might I have missed it by getting the plugin update via git? 

and just out of curiosity, could I ask why there's no issues board here? (or do I just not have access)

lucatume:
> @lucatume sounds good to me! I was informed that my [bridge plugin](https://github.com/helgatheviking/woocommerce-name-your-price-event-tickets/) for adding WooCommerce Name Your Price support to tickets (for a pay what you want ticket) no longer works. Looks like there have been some templating changes lately. I was trying to work on that but couldn't get past this.

In 5.0 (for The Events Calendar & co.) we've made some changes to the architecture.
We've got a [KB site up](https://theeventscalendar.com/knowledgebase/k/) to detail the changes; let me know if that helps.

> Is there an install routine that should have set that option? Might I have missed it by getting the plugin update via git?

There is none, a simple visit to the Events > Settings page and then a save should do it.
Yet bugs are a thing (like the one you fixed, thanks).

> and just out of curiosity, could I ask why there's no issues board here? (or do I just not have access)

We don't use GitHub to track issues.

sc0ttkclark:
@helgatheviking My apologies, I was out sick last week for a few days and missed this PR coming in. I was debugging something for another fix and found the issue. I applied essentially the same fix as your code did here. I'm sorry I didn't see this PR sooner :(

https://github.com/moderntribe/tribe-common/pull/1302

Thank you for your contribution, it indeed was on the right track!

sc0ttkclark:
I'll work on getting our fixes for the two separate tribe-common releases corrected.

sc0ttkclark:
I've pulled the B20.02 code into the G20.02 release branch and have this changelog reflected there now.

PR #1304, opened 2020-02-23
Cache purge shouldn't run multiple times in request.

bordoni:
🎟 [TCMN-38]
---
The goal is to prevent the Purging from happening when `set_last_occurence` was being called multiple times in the same request.

📹 https://i.bordoni.me/geu2Ypnk

[TCMN-38]: https://moderntribe.atlassian.net/browse/TCMN-38


src/Tribe/Cache.php
```diff
@@ -163,8 +166,14 @@ public function delete_transient( $id, $expiration_trigger = '' ) {
 	 * This uses a modification of the the query from https://core.trac.wordpress.org/ticket/20316
 	 *
 	 * @since 4.11.0
+	 *
+	 * @return void Just execute the database SQL no return required.
 	 */
 	public function delete_expired_transients() {
+		if ( tribe_get_var( 'has_deleted_expired_transients', false ) ) {
```
lucatume: Why use the global container here? 
If this class is the only one in charge of handling transient deletions, then a static var should do.

Is there any specific reason to use `tribe_(get|set)_var`?

If the idea is to allow external, third-party code, to handle this and then set `tribe_set_var( 'has_deleted_expired_transients' )`, then this should be documented in this method.

I would add a filter too, here, to bail out of the deletion entirely:

```
$delete_expired_transients = apply_filters( 'tribe_cache_delete_expired_transients', (bool) static::$transients_deleted );
if( false === (bool)$delete_expired_transients ) {
    return;
}

// ... here do the deletion

static::$transients_deleted = true;
```

src/Tribe/Cache.php
```diff
@@ -176,17 +185,63 @@ public function delete_expired_transients() {
...
+	 * Flag if we should delete
+	 *
+	 * @since TBD
+	 *
+	 * @param boolean $value If we should delete transients or not on shutdown.
+	 *
+	 * @return void No return for setting the flag.
+	 */
+	public function flag_required_delete_transients( $value = true ) {
+		tribe_set_var( 'should_delete_expired_transients', $value );
```
lucatume: Here I would set a static class value, not a value on the container.

src/Tribe/Cache.php
```diff
@@ -176,17 +185,63 @@ public function delete_expired_transients() {
...
+
+	/**
+	 * Runs on hook `shutdown` and will delete transients on the end of the request.
+	 *
+	 * @since TBD
+	 *
+	 * @return void No return for action hook method.
+	 */
+	public function maybe_delete_expired_transients() {
+		if ( ! tribe_get_var( 'should_delete_expired_transients', false ) ) {
```
lucatume: Same here with a static prop, not the container one.

lucatume: Probably this is a better place for the toggle filter I've suggested above.

src/Tribe/Cache.php
```diff
@@ -51,13 +55,14 @@ public function set( $id, $value, $expiration = 0, $expiration_trigger = '' ) {
...
-		 * @param int          $expiration         Cache expiration time.
+		 * @since 4.8
+		 *
 		 * @param string       $id                 Cache ID.
 		 * @param mixed        $value              Cache value.
 		 * @param string|array $expiration_trigger Action that triggers automatic expiration.
 		 * @param string       $key                Unique cache key based on Cache ID and expiration trigger last run time.
 		 *
-		 * @since 4.8
+		 * @param int          $expiration         Cache expiration time.
```
sc0ttkclark: Shouldn't this be the first param listed and without a blank line?

bordoni: Yes my bad.

src/Tribe/Cache.php
bordoni: The problem here is that you can have multiple instances of the class which the same class which was one of the big problems of the previous implementation.

lucatume:
@bordoni I've added tests to this PR and pushed them.
Running the tests I've made my mind up that using the global array cache (`tribe_(get|set)_var`) is safer in terms of globally controllably API.

PR #1305, opened 2020-02-25
Prevent addons that are later in the list from breaking the earlier ones

bordoni:
🎟 [FBAR-78]
---
📹 https://i.bordoni.me/v1urmx5b
Due to how the code was set up before if a given Addon Dependency that was on the end of the list returned `false` it would block the loading for all dependencies as if they all failed, instead of just warning us about the fail on for example Filterbar.

[FBAR-78]: https://moderntribe.atlassian.net/browse/FBAR-78

### DIFF
--- a/src/Tribe/Dependency.php
+++ b/src/Tribe/Dependency.php
@@ -521,18 +521,17 @@ if ( ! class_exists( 'Tribe__Dependency' ) ) {
 		 * @return bool  returns false if any dependency is invalid
 		 */
 		protected function check_addon_dependencies( $main_class ) {
-
-			$addon_dependencies = 0;
+			$addon_dependencies = [];
 
 			foreach ( $this->registered_plugins as $registered ) {
 				if ( empty( $registered['dependencies']['addon-dependencies'][ $main_class ] ) ) {
 					continue;
 				}
 
-				$addon_dependencies = $this->has_valid_dependencies( $registered, $registered['dependencies']['addon-dependencies'], true );
+				$addon_dependencies[ $main_class ] = $this->has_valid_dependencies( $registered, $registered['dependencies']['addon-dependencies'], true );
 			}
 
-			return $addon_dependencies;
+			return in_array( true,  $addon_dependencies, true );
 		}
 
 		/**
--- a/src/Tribe/Dependency.php
+++ b/src/Tribe/Dependency.php
@@ -409,11 +409,11 @@ if ( ! class_exists( 'Tribe__Dependency' ) ) {
 		 *
 		 * @since 4.9
 		 *
-		 * @param       $file_path
-		 * @param       $main_class
-		 * @param       $version
-		 * @param array $classes_req
-		 * @param array $dependencies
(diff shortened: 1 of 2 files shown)


src/Tribe/Dependency.php
```diff
@@ -409,11 +409,11 @@ public function get_pue_from_class( $class ) {
 		 *
 		 * @since 4.9
 		 *
-		 * @param       $file_path
-		 * @param       $main_class
-		 * @param       $version
-		 * @param array $classes_req
-		 * @param array $dependencies
+		 * @param string $file_path    Full file path to the base plugin file
```
lucatume:
```suggestion
		 * @param string $file_path    Full file path to the base plugin file.
```

src/Tribe/Dependency.php
```diff
@@ -409,11 +409,11 @@ public function get_pue_from_class( $class ) {
...
+		 * @param string $main_class   The Main/base class for this plugin
```
lucatume:
```suggestion
		 * @param string $main_class   The Main/base class for this plugin.
```

src/Tribe/Dependency.php
```diff
@@ -409,11 +409,11 @@ public function get_pue_from_class( $class ) {
...
+		 * @param string $version      The version
```
lucatume:
```suggestion
		 * @param string $version      The plugin version.
```

src/Tribe/Dependency.php
```diff
@@ -409,11 +409,11 @@ public function get_pue_from_class( $class ) {
...
+		 * @param array  $classes_req  Any Main class files/tribe plugins required for this to run
```
lucatume:
```suggestion
		 * @param array  $classes_req  Any Main class files/tribe plugins required for this to run.
```

src/Tribe/Dependency.php
```diff
@@ -409,11 +409,11 @@ public function get_pue_from_class( $class ) {
...
+		 * @param array  $dependencies an array of dependencies to check
```
lucatume:
```suggestion
		 * @param array  $dependencies an array of dependencies to check.
```

src/Tribe/Dependency.php
```diff
@@ -525,18 +521,17 @@ public function check_plugin( $main_class ) {
...
 				if ( empty( $registered['dependencies']['addon-dependencies'][ $main_class ] ) ) {
 					continue;
 				}
 
-				$addon_dependencies = $this->has_valid_dependencies( $registered, $registered['dependencies']['addon-dependencies'], true );
+				$addon_dependencies[ $main_class ] = $this->has_valid_dependencies( $registered, $registered['dependencies']['addon-dependencies'], true );
 			}
 
-			return $addon_dependencies;
+			return in_array( true,  $addon_dependencies, true );
```
lucatume: This code essentially means "if at least one of the dependencies is fulfilled, then return true"; why not just `return` from the `foreach` loop?

PR #1310, opened 2020-03-03
Fix/ecp 345 check doing filter

jesseeproductions:
_Ref:_ [ECP-345](https://moderntribe.atlassian.net/browse/ECP-345)


src/functions/template-tags/post.php
```diff
@@ -30,14 +30,18 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
...
+		 *
+		 * @since 0.71 of WordPress
+		 *
+		 * @param string $content Content of the current post.
+		 */
+
+		$content = apply_filters( 'the_content', $content );
+	}
+
 	$content = str_replace( ']]>', ']]&gt;', $content );
```
sc0ttkclark: This could also go into the above conditional IMO

lucatume: This piece of code was added to prevent some mangling of the content in our filtering, I think shortcodes, from elsewhere. I think this should stay outside of the conditional.

sc0ttkclark: I suppose it doesn't really matter if we run it and then WP runs the same line again but just noted for reference

tests/wpunit/Tribe/functions/templateTags/postTest.php
```diff
@@ -0,0 +1,40 @@
...
+class postTest extends \Codeception\TestCase\WPTestCase {
+
+	public function setUp() {
+		// before
+		parent::setUp();
+
+		// your set up methods here
+	}
+
+	public function tearDown() {
```
lucatume: Remove this method if not used.

tests/wpunit/Tribe/functions/templateTags/postTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+namespace Tribe\functions\templateTags;
+
+class postTest extends \Codeception\TestCase\WPTestCase {
+
+	public function setUp() {
```
lucatume: Remove this method if not used.

tests/wpunit/Tribe/functions/templateTags/postTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+	/**
+	 * @test
+	 */
+	public function it_does_not_run_apply_filters_when_doing_filter_and_create_nested_level_error() {
+		// Remove autop to make it easier to match content.
+		remove_filter( 'the_content', 'wpautop' );
+		$content = 'Test Content';
+		$post_id = $this->factory->post->create();
+		$expected = $content . tribe_get_the_content( null, false, $post_id ) . 'FILTERING';
```
lucatume: This is not readable and brittle.
If the `tribe_get_the_content` function does not work correctly, then we'll a bad expectation yielding either false positives or false negatives.

Refactor this to be a simple, readable, string of what we expect to get, e.g. `Test Content FILTERING`, and use that for comparison.

"Smart" expectations are destined to break, I speak from experience.

tests/wpunit/Tribe/functions/templateTags/postTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+		// Run tribe_get_the_content inside of the_content filter call.
+		add_filter( 'the_content', static function ( $content ) use ( $post_id ) {
```
lucatume: Alignment is weird here.

tests/wpunit/Tribe/functions/templateTags/postTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+				return $content . tribe_get_the_content( null, false, $post_id ) . 'FILTERING';
+			} );
+
+		$content = apply_filters( 'the_content', $content );
+
+		$this->assertEquals( $expected, $content );
+	}
```
lucatume: Please add another test to make sure the logic applies correctly when not doing the `the_content` filter.

PR #1326, opened 2020-03-23
Modify the way we handle SCRIPT_DEBUG for assets

bordoni:


src/Tribe/Assets.php
```diff
@@ -327,19 +327,24 @@ public static function maybe_get_min_file( $url ) {
 		// Strip the plugin URL and make this relative.
 		$relative_location = str_replace( $base_url, '', $url );
 
+		if ( defined( 'SCRIPT_DEBUG' ) && true === SCRIPT_DEBUG ) {
```
mitogh: We should use 
```suggestion
		if ( defined( 'SCRIPT_DEBUG' ) &&  SCRIPT_DEBUG ) {
```

In case someone set other value like `1`  in `define( 'SCRIPT_DEBUG', 1 );`

src/Tribe/Assets.php
```diff
@@ -327,19 +327,24 @@ public static function maybe_get_min_file( $url ) {
...
+			// Add the actual url after having the min file added.
+			$urls[] = $relative_location;
+		}
+
 		// If needed add the Min Files.
-		if ( ! defined( 'SCRIPT_DEBUG' ) || SCRIPT_DEBUG === false ) {
-			if ( substr( $relative_location, - 3, 3 ) === '.js' ) {
-				$urls[] = substr_replace( $relative_location, '.min', - 3, 0 );
-			}
+		if ( substr( $relative_location, - 3, 3 ) === '.js' ) {
```
mitogh: Space on unary operator. 
```suggestion
		if ( substr( $relative_location, -3, 3 ) === '.js' ) {
```

src/Tribe/Assets.php
```diff
@@ -327,19 +327,24 @@ public static function maybe_get_min_file( $url ) {
...
+			$urls[] = substr_replace( $relative_location, '.min', - 3, 0 );
+		}
 
-			if ( substr( $relative_location, - 4, 4 ) === '.css' ) {
-				$urls[] = substr_replace( $relative_location, '.min', - 4, 0 );
-			}
+		if ( substr( $relative_location, - 4, 4 ) === '.css' ) {
```
mitogh: Same as before.
```suggestion
		if ( substr( $relative_location, -4, 4 ) === '.css' ) {
```

src/Tribe/Assets.php
```diff
@@ -327,19 +327,24 @@ public static function maybe_get_min_file( $url ) {
...
+			$urls[] = substr_replace( $relative_location, '.min', - 4, 0 );
 		}
 
-		// Add the actual url after having the min file added.
-		$urls[] = $relative_location;
+		if ( ! defined( 'SCRIPT_DEBUG' ) || SCRIPT_DEBUG === false ) {
```
mitogh: I think the second part needs to check if `SCRIPT_DEBUG` as will fail generate a Notice if ` SCRIPT_DEBUG` not defined. 

```suggestion
		if ( ! defined( 'SCRIPT_DEBUG' ) || ( defined( 'SCRIPT_DEBUG' ) && ! SCRIPT_DEBUG )  ) {
```

PR #1333, opened 2020-04-06
Helper methods to check if specific Tribe plugins are active

Luc45:
This PR aims to provide a standard way of checking if certain Tribe plugins are active.

Feedbacks are welcome! :rocket:


src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
...
 
+		/**
+		 * Returns true if the plugin "Event Tickets" is active.
+		 * Only reliable after "plugins_loaded" has been called.
+		 *
+		 * @since TBD
+		 *
+		 * @return bool
+		 */
+		public static function is_event_tickets_active() {
```
mitogh, line 142: I think the methods are a great addition, however I would go with a more dynamic approach in order to favor the fact that we can add more plugins externally using filters and such, instead of creating custom functions for each plugin I think we can create a single function:

```php
public static function is_active( $plugin_name ) {
	$plugins = apply_filters( 'tribe_active_plugins', [] );
	return isset( $plugins[ $plugin_name ] );
}
```

So this one can be registered from the plugin side such as TEC or ET.

```php
// from Event tickets
add_filter( 'tribe_active_plugins', static function( $plugins ) {
	$plugins[ 'et' ] = true;
	$plugins[ Tribe__Events__Main::class ] = true;
	$plugins[ 'event_tickets' ] = true;
	$plugins[ 'tickets' ] = true; 
});
```

This allow to use different alias for different plugins and use the function such as: 

```php
new ( Tribe__Plugins )->is_active( 'et' );
new ( Tribe__Plugins )->is_active( 'tickets' );
```

Another benefit of this approach is the removal of concerns between common and external plugins, so in this way the plugins always have the control on how to respond to the `is_active` function.

Luc45: Interesting. The plugin declares itself as loaded, through the filter... Pretty clever.

What do you think we wait a little bit to see if someone else has thoughts on it before I update this PR?

bordoni: I would love to see something like what @mitogh suggested there. But I would possibly throw a `_doing_it_wrong` if called before `plugins_loaded` which would ensure we avoid problems in the fugure.

sc0ttkclark: Can we stick to one plugin key for each plugin instead of all of the aliases (the repo slug)?

* event-tickets
* event-tickets-plus
* the-events-calendar
* events-pro

etc..

mitogh: I think using the same slug from the containers would be ideal: 

- https://moderntribe.github.io/products-engineering/docs/plugins/namespacing/

To keep it consistent with the rest of containers where a plugin main container is registered.

sc0ttkclark: Container slugs is perfect here!

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
 			return apply_filters( 'tribe_plugins_get_list', $this->tribe_plugins );
 		}
 
+		/**
+		 * Returns true if the plugin "Event Tickets" is active.
```
lucatume, line 135:
```suggestion
		 * Returns true if the plugin "Event Tickets" is active.
		 *
```

Blank line between short and long/detail description.

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
...
+		 * Only reliable after "plugins_loaded" has been called.
+		 *
+		 * @since TBD
+		 *
+		 * @return bool
```
lucatume, line 140:
```suggestion
		 * @return bool Whether Event Tickets is active or not.
```

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
...
+		 */
+		public static function is_ical_importer_active() {
+			return class_exists( 'Tribe__Events__Ical_Importer__Main' );
+		}
+
+		/**
+		 * Returns true if the plugin "Eventbrite Tickets" is active.
```
sc0ttkclark, line 243:
```suggestion
		 * Returns true if the plugin "Eventbrite Tickets" is active.
		 *
```

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
...
+		 * @since TBD
+		 *
+		 * @return bool
+		 */
+		public static function is_eventbrite_tickets_active() {
+			return class_exists( 'Tribe__Events__Tickets__Eventbrite__Main' );
+		}
+
+		/**
+		 * Returns true if the plugin "Advanced Post Manager" is active.
```
sc0ttkclark, line 255:
```suggestion
		 * Returns true if the plugin "Advanced Post Manager" is active.
		 *
```

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,136 @@ public function get_list() {
...
+		 * @since TBD
+		 *
+		 * @return bool
+		 */
+		public static function is_facebook_events_importer_active() {
+			return class_exists( 'Tribe__Events__Facebook__Importer' );
+		}
+
+		/**
+		 * Returns true if the plugin "iCal Importer" is active.
```
sc0ttkclark, line 231:
```suggestion
		 * Returns true if the plugin "iCal Importer" is active.
		 *
```

src/Tribe/Plugins.php
sc0ttkclark: Methods will be redone

src/Tribe/Plugins.php
sc0ttkclark: Methods will be redone

Luc45:
I've refactored this according to the code-review feedback.

Now each plugin must hook into that filter to set itself as active. Should I take any action in that regard?

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,27 @@ public function get_list() {
...
+		public static function is_active( $plugin_name ) {
+			if ( ! did_action( "plugins_loaded" ) ) {
+				_doing_it_wrong(
+					__METHOD__,
+					__( 'Using this function before "plugins_loaded" action has fired can return unreliable results.', 'tribe-common' ),
+					'TBD'
+				);
+			}
+
+			$plugins = apply_filters( 'tribe_active_plugins', [] );
```
bordoni, line 152: Docblock missing here.

Luc45: Added docblock. Thank you!

Luc45:
> Code looks good to me, just pending a release/ticket

Thanks Scott. Should I talk with someone to get a ticket for this?

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,37 @@ public function get_list() {
...
+			if ( ! did_action( "plugins_loaded" ) ) {
+				_doing_it_wrong(
+					__METHOD__,
+					__( 'Using this function before "plugins_loaded" action has fired can return unreliable results.', 'tribe-common' ),
+					'TBD'
+				);
+			}
+
+			/**
+			 * The array that each Tribe plugin overrides to
```
lucatume, line 153:
```suggestion
			 * Filters the array that each Tribe plugin overrides to
```

src/Tribe/Plugins.php
```diff
@@ -131,5 +131,37 @@ public function get_list() {
...
+			 *
+			 * @example [ 'the-events-calendar' => true, 'event-tickets' => true ]
+			 *
+			 * @since   TBD
+			 *
+			 * @return array Plugin slugs as keys and bool as value for whether it's active or not.
+			 */
+			$plugins = apply_filters( 'tribe_active_plugins', [] );
+
+			return isset( $plugins[ $plugin_name ] ) && $plugins[ $plugin_name ];
```
lucatume, line 164:
```suggestion
			return !empty( $plugins[ $plugin_name ] );
```

lucatume: To make this even more robust we might even use `tribe_is_truthy` here:
```suggestion
			return isset( $plugins[ $plugin_name ] ) && tribe_is_truthy( $plugins[ $plugin_name ] );
```

This would allow filters to mark a plugin as active setting the value to `yes` (and your code already supported this), but it would support setting a plugin as inactive setting its value to `no`.

sc0ttkclark:
@Luc45 you can create a new ticket in triage for it and we can get it slated for the next sprint, ping Jeff with the ticket and we can do it for G20.05.

readme.txt
```diff
@@ -2,8 +2,9 @@
 
 == Changelog ==
 
-= [TBD] TBD =
+= [4.12.1] 2020-04-20 =
 
+* Feature - Added a helper method to check if a given plugin is active. Usually a Modern Tribe plugin.
```
```suggestion
* Feature - Added a helper method `Tribe__Plugins::is_active( 'slug' )` to check if a given plugin is active.
```

PR #1334, opened 2020-04-06
Helper method to determine if Blocks are being used for Events

Luc45:
This PR adds a helper method to determine if Events are using blocks.

This is useful to add conditional behavior for Blocks/Classical scenarios.

This PR ports some logic from TEC to make that available on Common https://github.com/moderntribe/the-events-calendar/blob/master/src/Tribe/Editor/Compatibility.php#L72-L83


sc0ttkclark:
@Luc45 Is this needed/used for a specific PR/ticket on ET/ET+?

Luc45:
> @Luc45 Is this needed/used for a specific PR/ticket on ET/ET+?

Yes. https://moderntribe.atlassian.net/browse/ETP-234

I'm porting the logic directly into a method in ET, but I thought it would be common enough to make a PR to Common.

sc0ttkclark:
Thanks for the update! Let's get a ticket for this and put it into a future sprint.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,17 @@ public function is_classic_editor() {
 
 		return $is_classic_editor_request || $disabled_by_plugin || $disabled_by_filter;
 	}
+
+	/**
+	 * Returns whether Blocks are being used for Events
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
```
lucatume, line 220:
```suggestion
	 * @return bool Whether Blocks are being used for Events or not.
```

Add doc comment to all `@return` and `@param` tags.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,17 @@ public function is_classic_editor() {
 
 		return $is_classic_editor_request || $disabled_by_plugin || $disabled_by_filter;
 	}
+
+	/**
+	 * Returns whether Blocks are being used for Events
```
lucatume, line 216:
```suggestion
	 * Returns whether Blocks are being used for Events.
```

Full stop at end of comments.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,17 @@ public function is_classic_editor() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function is_events_using_blocks() {
+		$is_blocks_active_for_events = tribe_is_truthy( tribe_get_option( 'toggle_blocks_editor', false ) );
```
lucatume, line 223: Pre-computing this here is not efficient.
Since the condition you're running after is an `&&` the second part will only be evaluated if the first is true. Pulling options involves some filters and, possibly, a db query.
I would do this on the fly.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,17 @@ public function is_classic_editor() {
...
+
+		return $this->should_load_blocks() && $is_blocks_active_for_events;
```
lucatume, line 225:
```suggestion
		return $this->should_load_blocks()
		    && tribe_is_truthy( tribe_get_option( 'toggle_blocks_editor' ) );
```

lucatume: This way the second, costly, check is done only if the first is passing.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,17 @@ public function is_classic_editor() {
...
+	/**
+	 * Returns whether Blocks are being used for Events
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function is_events_using_blocks() {
+		$is_blocks_active_for_events = tribe_is_truthy( tribe_get_option( 'toggle_blocks_editor', false ) );
+
```
sc0ttkclark, line 224:
```suggestion
```

bordoni:
We should aim to release this in tandon with https://github.com/moderntribe/event-tickets/pull/1633

src/Tribe/Editor.php
```diff
@@ -211,4 +211,35 @@ public function is_classic_editor() {
...
+		 * or the classical editor.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $is_using_blocks True if using blocks. False if using the classical editor.
+		 */
+		$is_using_blocks = apply_filters( 'tribe_is_using_blocks', null );
+
+		// Early bail: The filter was overridden to return either true or false.
+		if ( is_bool( $is_using_blocks ) ) {
```
Luc45, line 234: https://3v4l.org/umO4f

lucatume: While this will work, returning a value like `1` or (string version) `"1"` here would still mean that the filtered value is true.

WordPress will, typically, check if the early bail filter value is not `null`

```suggestion
		if ( null !== $is_using_blocks ) {
```

Any other value would be cast to true and returned; I would modify the logic to that.

src/Tribe/Editor.php
```diff
@@ -211,4 +211,35 @@ public function is_classic_editor() {
 
 		return $is_classic_editor_request || $disabled_by_plugin || $disabled_by_filter;
 	}
+
+	/**
+	 * Whether the events are being served using Blocks or the Classical Editor.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool True is using Blocks. False if using the Classical Editor.
```
sc0ttkclark, line 220:
```suggestion
	 * @return bool True if using Blocks. False if using the Classical Editor.
```

PR #1344, opened 2020-04-15
Bucket/block editor initial state

paulmskim:
Merging bucket into release


src/Tribe/Editor/Blocks/Abstract.php
```diff
@@ -203,5 +203,40 @@ public function assets() {
 	 */
 	public function hook() {
 	}
+
+	/**
+	 * Returns the block data for the block editor.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
```
lucatume, line 212:
```suggestion
	 * @return array<string,mixed> The block editor data.
```

PR #1346, opened 2020-04-20
Expose one more variable and fix loading from themes problem

bordoni:
Resolves a problem where we didn't have external access to a variable, which is important for extension development.

Also solves the problem where Child Themes are not loaded if a Parent Theme has the same Tribe template overwrite.


src/Tribe/Template.php
```diff
@@ -183,6 +183,17 @@ public function set_template_folder_lookup( $value = true ) {
 		return $this;
 	}
 
+	/**
+	 * Gets in this instance of the template engine whether we are looking public folders like themes.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool If we are looking into theme folders
```
sc0ttkclark, line 191:
```suggestion
	 * @return bool Whether we are looking into theme folders.
```

PR #1349, opened 2020-04-27
feat(Autoloader) support flat and nested namespaced classes

lucatume:
This PR adds support for namespaced classes in the Autoloader.

While we had a "kinda" working support that would rely on:

1. the prefix had to be set to one w/o the leading `\` slash.
2. the directory structure of the namespaced classes had to be falt, w/o sub-directories.

This PRs allows registering prefixes in all the diff. ways and adds support for nested
directory structures.

### DIFF
--- a/src/Tribe/Autoloader.php
+++ b/src/Tribe/Autoloader.php
@@ -163,11 +163,25 @@
 
 			protected function get_prefixed_path( $class ) {
 				foreach ( $this->prefixes as $prefix => $dirs ) {
+					$is_namespaced = false !== strpos( $prefix, '\\' );
+
+					if ( $is_namespaced ) {
+						// If the prefix is a namespace, then normalize it.
+						$prefix = trim( $prefix, '\\' ) . '\\';
+					}
+
 					if ( strpos( $class, $prefix ) !== 0 ) {
 						continue;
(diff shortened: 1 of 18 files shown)


src/Tribe/Autoloader.php
```diff
@@ -163,11 +163,25 @@ private function normalize_root_dir( $root_dir ) {
 
 			protected function get_prefixed_path( $class ) {
 				foreach ( $this->prefixes as $prefix => $dirs ) {
+					$is_namespaced = false !== strpos( $prefix, '\\' );
+
+					if ( $is_namespaced ) {
+						// If the prefix is a namespace, then normalize it.
+						$prefix = trim( $prefix, '\\' ) . '\\';
```
sc0ttkclark, line 170: On further review, this would be useful if `$this->prefixes` already had the prefix normalized by this point as it's referenced via `get_prefix_by_slug()` elsewhere and the returned prefix would need normalizing there manually for each use.

sc0ttkclark: Abstracted this in https://github.com/moderntribe/tribe-common/pull/1350 so we can remove this if conditional that normalizes here.

PR #1354, opened 2020-05-03
incorrect parameter for doing_filter 

llslim:
I use the events calendar and events calendar pro plugins for a client website where we display webinar information to members. The site uses Wordpress 5.x and blocks, and the 'default page template' setting for the calendar.  I tried to embed a youtube video in an event post using the core-embed/youtube block. The result doesn't embed the video, but displays the link. A common error and fixed by the 'do_blocks' filter on the 'the_content' hook. Due to the incorrect parameter to the doing_filter function none of the filters are being applied to the content of an event.

The issue:
'the_content' is the hook tag for filters to be added to and not an actual filter. The [WP documentation  for doing_filter](https://developer.wordpress.org/reference/functions/doing_filter/) gives examples of the correct parameter.

The fix:
I just picked a most common filter, 'wptexturize', that is added to the 'the_content' by default in the file /wp-includes/default-filters.php of wordpress core. The filter may be change to something more appropriate. 

From looking at the post.php file history and readme/commit messages, in commit `5aac02d4` the if statement was put in place due to filters being called multiple times. The issue the commit referenced is in Jira which I don't have access to. So I couldn't verify if my fix still satisfies that original issue as well. 

A side note:
a doing_filter check is usually done from within a filter callback function, and the core get_the_content function from which tribe_get_the_content is duplicated from doesn't have the check. Is it feasible to export the Jira issue into the github issue queue for this project so more investigation can be done of the original issue?


sc0ttkclark:
I'm not sure at what point you see that `doing_filter` working for the `wptexturize` change here, I can't seem to reproduce that working. I can confirm that `doing_filter( 'the_content' )` works as expected in WP itself.

bordoni:
@llslim we are working on a parallel fix that will resolve the bug you reported here. I will udpate this PR once I get the code merged.

llslim:
> @llslim we are working on a parallel fix that will resolve the bug you reported here. I will udpate this PR once I get the code merged.

@bordoni cool. Does the parallel fix involve removing the check entirely? Just wondering

sc0ttkclark:
@llslim Can you confirm that the issue is resolved for you with the latest stable release?

@bordoni when you return, can you follow up and close this if the changes we have in that area cover this?

bordoni:
@llslim, one of the possibilities was to remove the check all together, the problem we are facing is that removing it also creates other compatibility problems with Event Tickets.

bordoni:
I've pushed a modification to your code that merges what we applied as a fix to yours, we will need to get some testing around it, but hopefully it helps.

llslim:
I'll take a look at the changes and I'll close this issue once I have this
week. Thanks for your help.

PR #1357, opened 2020-05-05
Resolve when Plugins load a non-full version of Select2 too early.

bordoni:
🎟 [TEC-3448]
---
Couldn't necessarily reproduce the problem, but for our intends and purposes we need to have the full version of Select2 loaded using SelectWoo, so we need to enforce it on the admin.

[TEC-3448]: https://moderntribe.atlassian.net/browse/TEC-3448


src/Tribe/Assets_Pipeline.php
```diff
@@ -38,4 +40,26 @@ public function prevent_underscore_conflict( $tag, $handle ) {
 		}
 		return $tag;
 	}
+	/**
+	 * Before underscore is loaded to the FE we add two scripts on before and one after to prevent underscore from
```
lucatume, line 44:
```suggestion
	 * Before selectWoo is loaded to the FE we add two scripts on before and one after to prevent selectWoo from
```

src/Tribe/Assets_Pipeline.php
```diff
@@ -38,4 +40,26 @@ public function prevent_underscore_conflict( $tag, $handle ) {
...
+	 * taking place on the global namespace if lodash is present.
```
lucatume, line 45:
```suggestion
	 * taking place on the global namespace if select2 is present.
```

readme.txt
```diff
@@ -4,6 +4,7 @@
 
 = [TBD] TBD =
 
+* Fix - Prevent other Select2 that were no loaded with full version to create failing scenarios with our loading of SelectWoo [TEC-3448]
```
sc0ttkclark, line 7: This reads weird, can you improve?

src/Tribe/Assets_Pipeline.php
sc0ttkclark:
```suggestion
	 * taking place on the global namespace if Select2 is present.
```

readme.txt
bordoni, line 7:
```suggestion
* Fix - Prevent third-party Select2 usage from breaking SelectWoo for The Events Calendar and Event Tickets dropdowns [TEC-3448]
```

PR #1359, opened 2020-05-05
Template Entry Points

jesseeproductions:
_Ref:_ [TCMN-65](https://moderntribe.atlassian.net/browse/TCMN-65)

Add Entry Points for:
- inside template container, before content
- inside template container, after content


src/Tribe/Template.php
```diff
@@ -584,6 +619,90 @@ public function get_template_file( $name ) {
 		return false;
 	}
 
+	/**
+	 * Runs the entry point hooks and filters
```
lucatume, line 623:
```suggestion
	 * Runs the entry point hooks and filters.
```

src/Tribe/Template.php
```diff
@@ -584,6 +619,90 @@ public function get_template_file( $name ) {
...
+	 *
+	 * @param string  $entry_point_name The name of the entry point.
+	 * @param boolean $echo             If we should also print the entry point content.
+	 *
+	 * @return null|string
```
lucatume, line 628:
```suggestion
	 * @return null|string Either the entry point HTML, or `null` to indicate the entry point is not applicable.
```

src/Tribe/Template.php
```diff
@@ -584,6 +619,90 @@ public function get_template_file( $name ) {
...
+		/**
+		 * Generic entry point action for the current template.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $hook_name        For which template include this entry point belongs.
+		 * @param string $entry_point_name Which entry point specifically we are triggering.
+		 * @param self   $template         Current instance of the template class doing this entry point.
+		 */
+		do_action( "tribe_template_entry_point:$hook_name", $hook_name, $entry_point_name, $this );
```
lucatume, line 660: I would check if there's any listener for this action before running it using `has_action`
The same I would do for the rest of these actions.
Filters, and actions, are costly operations with a relevant performance overhead price even if nothing is listening on them; furthermore, the `all` filter will always fire.
The `has_filter` and `has_action` functions are cheap and can save a lot of wasted processing.

src/Tribe/Template.php
```diff
@@ -857,10 +979,80 @@ public function template( $name, $context = [], $echo = true ) {
...
 		}
 
+		// Revert the current hook name.
+		$this->set_template_current_hook_name( null );
+
+		return $html;
+	}
+
+	/**
+	 * @param $html
```
lucatume, line 996: Remove this left-over.

src/Tribe/Template.php
```diff
@@ -857,10 +979,80 @@ public function template( $name, $context = [], $echo = true ) {
...
+	 * Run the hooks for the container entry points.
+	 *
+	 * @since  TBD
+	 *
+	 * @param string $html The html of the current template.
+	 *
+	 * @return string|false Either the final entry point content HTML or `false` if no entry point could be found or set to false.
+	 */
+	private function template_hook_container_entry_points( $html ) {
+		$regexp = '/<(?<is_end>\/)*(?<tag>[A-Z0-9]*)(?:\b)*[^>]*>/mi';
```
lucatume, line 1011: Regular expression, while very powerful, are expensive operations.

The result should be instance-cached (cached for the current request duration only).

To cache something in an guaranteed ephemeral cache, use the following API:

```php
$key = md5(...);
$cache = tribe_cache();
$cached = $cache[$key];

if( false !== $cached ){
    return $cached;
}

$value = 'foo';

$cache[$key] = $value;

return $value;
```

src/Tribe/Template.php
```diff
@@ -857,10 +979,80 @@ public function template( $name, $context = [], $echo = true ) {
...
+			return $html;
+		}
+
+		$first_tag_html = reset( $html_matches );
+		$last_tag_html  = end( $html_matches );
+
+		$open_container_entry_point_html  = $this->do_entry_point( 'after_container_open', false );
+		$close_container_entry_point_html = $this->do_entry_point( 'before_container_close', false );
+
+		$html = \Tribe\Utils\Strings::replace_first( $first_tag_html, $first_tag_html . $open_container_entry_point_html, $html );
```
lucatume, line 1053: `use` this and get rid of the fully-qualified name.

src/Tribe/Utils/Strings.php
```diff
@@ -0,0 +1,65 @@
...
+namespace Tribe\Utils;
+
+/**
+ * Class Strings
+ *
+ * @since TBD
+ *
+ * @package Tribe\Utils
+ */
+class Strings {
```
lucatume, line 18: This needs tests.

tests/wpunit/Tribe/TemplateTest.php
```diff
@@ -57,4 +64,172 @@ public function should_allow_setting_contextual_values_without_overriding_the_pr
...
+		} );
+		add_action( 'tribe_template_entry_point:dummy/dummy-invalid-template-01:before_container_close', function () {
+			echo '%%before_container_close%%';
+		} );
+		$html = $template->template( 'dummy-invalid-template-01', [], false );
+
+		$this->assertNotContains( '%%after_container_open%%', $html );
+		$this->assertStringEndsNotWith( '%%before_container_close%%', $html );
+
+		// Invalid Test 2
```
lucatume, line 129: Break into diff. tests; this and any other.

tests/wpunit/Tribe/TemplateTest.php
```diff
@@ -57,4 +64,172 @@ public function should_allow_setting_contextual_values_without_overriding_the_pr
...
+		} );
+		add_action( 'tribe_template_entry_point:dummy/dummy-valid-template-01:before_container_close', function () {
+			echo '%%before_container_close%%';
+		} );
+		$html = $template->template( 'dummy-valid-template-01', [], false );
+
+		$this->assertContains( '<a href="https://tri.be" class="test" target="_blank" title="Test Link" data-link="automated-tests">%%after_container_open%%', $html );
+		$this->assertStringEndsWith( '%%before_container_close%%</a>', $html );
+
+		// Valid Test 2
```
lucatume, line 186: Break into diff. tests; this and below.

src/Tribe/Template.php
bordoni: So, in this case, it can be complicated to cache. Because the MD5 on itself can be an expensive operation for a long string like the whole HTML. This creates a problem because the improvements are very small.

And if we use the path for caching it can fail the regular expression because the variables in the context can also dictate what will be in the first and last line, specifically for the string replacement.

Need to play around with how this behaves to see if this actually bring any performance improvements.

jesseeproductions: I added the caching here using the html for the MD5.  So we can compare if this helps or not. We could change to use this to get the hook name:

`$hook_name = $this->get_template_current_hook_name();`

Then we could use something like this for the cache key `tribe_template_entry_point_cache:$hook_name`

jesseeproductions:
I took care of everything here. The outstanding point is what to do with the caching.

src/Tribe/Template.php
```diff
@@ -672,10 +801,13 @@ public function template( $name, $context = [], $echo = true ) {
 			$namespace        = array_merge( $origin_folder_appendix, $legacy_namespace );
 		}
 
-		// Setup the Hook name
+		// Setup the Hook name.
 		$legacy_hook_name = implode( '/', $legacy_namespace );
 		$hook_name        = implode( '/', $namespace );
 
+		// Store the current hookname for the purposes of entry-points.
+		$this->set_template_current_hook_name( $hook_name );
```
sc0ttkclark, line 809: Just curious, if a template is outputting another template, does that prevent being able to hook into the end of the parent template?

bordoni: Oh yeah that is a good question I will poke around that behavior. :real_heart:

bordoni: Modified the behavior here, to make sure it saves the previous locally and replaces before exiting. So it seems to do the job of having the entry points correctly even when you have child templates.

src/Tribe/Utils/Strings.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Replace the last occurrence of a given value in the string.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $search  The string to search for and replace.
+	 * @param string $replace The replacement string.
+	 * @param string $subject The string to do the search and replace from.
+	 *
+	 * @return string
```
sc0ttkclark, line 54:
```suggestion
	 * @return string The string with the last occurrence of a given value replaced.
```

src/Tribe/Utils/Strings.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Replace the first occurrence of a given value in the string.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $search  The string to search for and replace.
+	 * @param string $replace The replacement string.
+	 * @param string $subject The string to do the search and replace from.
+	 *
+	 * @return string
```
sc0ttkclark, line 29:
```suggestion
	 * @return string The string with the first occurrence of a given value replaced.
```

src/Tribe/Template.php
sc0ttkclark: Caching on hook name sounds useful and makes sense to me.

src/Tribe/Template.php
```diff
@@ -985,4 +1181,33 @@ public function get_local_values() {
...
+	 *
+	 * @param string $html The html of the current template.
+	 *
+	 * @return array An array of matches from the regular expression.
+	 */
+	private function get_entry_point_matches( $html ) {
+
+		$key    = md5( $html );
+		$cache  = tribe_cache();
+		$cached = $cache[ $key ];
```
sc0ttkclark, line 1198:
```suggestion
		// Set up cache key using current hook name.
		$hook_name = $this->get_template_current_hook_name();
		$key       = "tribe_template_entry_point_cache:{$hook_name}";

		// Get cache instance.
		$cache = tribe_cache();

		// Get cached value (if set).
		$cached = $cache[ $key ];
```

src/Tribe/Template.php
sc0ttkclark: 👍

PR #1365, opened 2020-05-08
generate Tag from Class Name & make filterable

cliffordp:
[TCMN-69]

2min video demo of it generating the shortcode tag from the extending/child class' name, regardless of namespace: https://share.getcloudapp.com/jkulWWED

[TCMN-69]: https://moderntribe.atlassian.net/browse/TCMN-69

### DIFF
--- a/src/Tribe/Shortcode/Shortcode_Abstract.php
+++ b/src/Tribe/Shortcode/Shortcode_Abstract.php
@@ -1,31 +1,17 @@
 <?php
-/**
- * The abstract all shortcodes should implement.
- *
- * @since  4.12.0
- *
- * @package Tribe\Shortcode
- */
+
 namespace Tribe\Shortcode;
 
 use Tribe__Utils__Array as Arr;
 
 /**
- * Interface Shortcode_Interface
- *
- * @since   4.12.0
+ * The abstract all shortcodes should implement.
  *
  * @package Tribe\Shortcode
+ *
+ * @since   4.12.0
  */
 abstract class Shortcode_Abstract implements Shortcode_Interface {
-	/**
-	 * Slug of the current shortcode.
-	 *
-	 * @since   4.12.0
-	 *
-	 * @var   string
-	 */
-	protected $slug;
 
 	/**
 	 * Default arguments to be merged into final arguments of the shortcode.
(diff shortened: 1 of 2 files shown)


src/Tribe/Shortcode/Shortcode_Abstract.php
```diff
@@ -63,6 +49,51 @@ abstract class Shortcode_Abstract implements Shortcode_Interface {
...
+
+		$tag = str_replace( '-', '_', sanitize_key( $tag ) );
+
+		return $tag;
+	}
+
+	/**
+	 * Get a shortcode tag based on this class' name.
+	 *
+	 * @return string
```
sc0ttkclark, line 87:
```suggestion
	 * @since TBD
	 *
	 * @return string
```

src/Tribe/Shortcode/Shortcode_Abstract.php
```diff
@@ -63,6 +49,51 @@ abstract class Shortcode_Abstract implements Shortcode_Interface {
 	 */
 	protected $content;
 
+	/**
+	 * Get the shortcode tag, forced to lowercase with underscores.
+	 *
+	 * If `$this->tag` exists, use it, else it will be created dynamically from this class' name.
+	 *
```
sc0ttkclark, line 56:
```suggestion
	 *
	 * @since TBD
	 *
```

src/Tribe/Shortcode/Shortcode_Abstract.php
```diff
@@ -63,6 +49,55 @@ abstract class Shortcode_Abstract implements Shortcode_Interface {
...
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_tag() {
+		if (
+			! empty( $this->tag )
```
sc0ttkclark, line 63: Should we define the protected tag property?

cliffordp: If you want to use the prop, you'd add it in the child class, not necessary in abstract class.
*Should we???* Personally, no because we'd like the shortcode class to be descriptively named and therefore the shortcode tag 99% of the time should be generated. Defining the prop in the abstract would encourage its use more...

bordoni:
As we talked yesterday @cliffordp, I would like us to avoid having two ways to determine the slug of the shortcode internally, the method `start_tag_from_class_name` creates confusion in my opinion.

Since there is code in Events Calendar Pro using the shortcodes already, so I would avoid refactoring from `get_registration_slug` to `get_tag`.

What we should improve on the abstract is:
- Move towards having the slug be a static value that gets exposed via a public static method.
- Add filters to this new static method, so the registration of the shortcode slug can be changed.
- Make sure `parse_arguments` use the new static method instead of calling `$this->slug` directly.

PR #1370, opened 2020-05-12
Spotfix - Filter: tribe_extension_is_disallowed

cliffordp:
See https://github.com/moderntribe/event-tickets-plus/pull/990 for a demo video

### DIFF
--- a/src/Tribe/Extension.php
+++ b/src/Tribe/Extension.php
@@ -177,7 +177,7 @@ abstract class Tribe__Extension {
 		 *
 		 * @since TBD
 		 *
-		 * @param                  $is_disallowed        False by default.
+		 * @param bool             $is_disallowed        False by default.
 		 * @param string           $extension_class_name This extension's class name string
 		 *                                               (without initial forward slash for namespaced classes).
 		 * @param Tribe__Extension $this_instance        This extension class' instance.
--- a/src/Tribe/Extension.php
+++ b/src/Tribe/Extension.php
@@ -186,8 +186,6 @@ abstract class Tribe__Extension {
 		 * @param string           $extension_class_name This extension's class name string
 		 *                                               (without initial forward slash for namespaced classes).
 		 * @param Tribe__Extension $this_instance        This extension class' instance.
-		 *
-		 * @return bool
 		 */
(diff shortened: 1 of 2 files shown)


src/Tribe/Extension.php
```diff
@@ -164,13 +164,40 @@ protected function construct() {}
 			$this->get_plugin_file(),
```
sc0ttkclark, line 164: Let's set this to a var above so it can be reused further below.

src/Tribe/Extension.php
```diff
@@ -164,13 +164,40 @@ protected function construct() {}
...
 			$this->get( 'class' ),
```
sc0ttkclark, line 165: Let's set this to a var above so it can be reused further below.

src/Tribe/Extension.php
```diff
@@ -164,13 +164,40 @@ protected function construct() {}
...
 			$this->get_version(),
-			$this->get( 'requires', array() )
```
sc0ttkclark, line 167: Let's set this to a var above so it can be reused further below.

src/Tribe/Extension.php
```diff
@@ -378,6 +405,21 @@ protected function construct() {}
 		);
 	}
 
+	/**
+	 * Gets the error message about being explicitly disallowed.
```
sc0ttkclark, line 409:
```suggestion
	 * Gets the error message about being explicitly disallowed.
	 *
	 * @since TBD
	 *
	 * @return string Notice text.
```

src/Tribe/Extension.php
cliffordp: oh, yeah, duh

src/Tribe/Extension.php
```diff
@@ -160,25 +160,56 @@ protected function construct() {}
...
+		 * Explicitly disallow an extension, such as a core plugin having absorbed/replaced its functionality.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool             $is_disallowed        False by default.
+		 * @param string           $extension_class_name This extension's class name string
+		 *                                               (without initial forward slash for namespaced classes).
+		 * @param Tribe__Extension $this_instance        This extension class' instance.
+		 *
+		 * @return bool
```
sc0ttkclark, line 190:
```suggestion
```

PR #1371, opened 2020-05-13
feature(tracker): Adds a method to get the change date of a key.

Camwyn:
The tracker's data is useful for determining when something happens.
This code adds a method that allows you to provide the post ID and a key and get back the timestamp of the change.
Useful for determining if something was changed before/after something else, such as publish date.

Part of [VE-17]

[VE-17]: https://moderntribe.atlassian.net/browse/VE-17

### DIFF
--- a/src/Tribe/Tracker.php
+++ b/src/Tribe/Tracker.php
@@ -96,11 +96,7 @@ class Tribe__Tracker {
 		$modified = get_post_meta( $post_id, self::$field_key, true );
 
 		// If the key is missing or empty/null return false - no recorded change.
-		if ( empty( $modified[ $meta_key ] ) ) {
-			return false;
-		}
-
-		return $modified[ $meta_key ];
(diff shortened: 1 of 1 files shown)


src/Tribe/Tracker.php
```diff
@@ -83,7 +83,28 @@ public function has_field_changed( $field, $new, $old ) {
...
+	 */
+	public function get_modified_date( $meta_key, $post_id ) {
+		$modified = get_post_meta( $post_id, self::$field_key, true );
+
+		// If the key is missing or empty/null return false - no recorded change.
+		if ( empty( $modified[ $meta_key ] ) ) {
+			return false;
+		}
+
+		return $modified[ $meta_key ];
```
bordoni, line 103:
```suggestion
		return Tribe__Utils__Array::get( $modified, $meta_key, false );
```

PR #1372, opened 2020-05-15
hen using Block Editor we ensure that `apply_filters` for `the_content` on `tribe_get_the_content`

bordoni:
🎟 [TEC-3456]
---
This bug was introduced on [ECP-345], I still feel like this feels a bit hacky compared to just running the `apply_filters` every time like the default WordPress `the_content()` does.

[TEC-3456]: https://moderntribe.atlassian.net/browse/TEC-3456
[ECP-345]: https://moderntribe.atlassian.net/browse/ECP-345

### DIFF
--- a/src/functions/template-tags/post.php
+++ b/src/functions/template-tags/post.php
@@ -30,7 +30,7 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
 		$content = get_the_content( $more_link_text, $strip_teaser );
 	}
 
-	$has_blocks = false;
+	$has_blocks = function_exists( 'has_blocks' ) && has_blocks( $content );
 	// If blocks are present we need to run the content filter.
 	if ( function_exists( 'has_blocks' ) ) {
 		$has_blocks = has_blocks( $content );
--- a/src/functions/template-tags/post.php
+++ b/src/functions/template-tags/post.php
@@ -31,7 +31,7 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
 	}
 
 	$has_blocks = false;
-	// If blocks are present we need to for sure pass the content filter.
+	// If blocks are present we need to run the content filter.
 	if ( function_exists( 'has_blocks' ) ) {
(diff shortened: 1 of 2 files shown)


src/functions/template-tags/post.php
```diff
@@ -30,7 +30,13 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
 		$content = get_the_content( $more_link_text, $strip_teaser );
 	}
 
-	if ( ! doing_filter( 'the_content' ) ) {
+	$has_blocks = false;
+	// If blocks are present we need to for sure pass the content filter.
```
lucatume, line 34:
```suggestion
	// If blocks are present we need to run the content filter.
```

src/functions/template-tags/post.php
```diff
@@ -30,7 +30,13 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
 		$content = get_the_content( $more_link_text, $strip_teaser );
 	}
 
-	if ( ! doing_filter( 'the_content' ) ) {
+	$has_blocks = false;
```
lucatume, line 33:
```suggestion
	$has_blocks = function_exists( 'has_blocks' ) && has_blocks( $content );
```

src/functions/template-tags/post.php
```diff
@@ -30,7 +30,13 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
...
+	// If blocks are present we need to for sure pass the content filter.
+	if ( function_exists( 'has_blocks' ) ) {
```
lucatume, line 35: Following the change suggested above, I would remove this piece of code to go with the simpler version.

src/functions/template-tags/post.php
```diff
@@ -30,7 +30,13 @@ function tribe_get_the_content( $more_link_text = null, $strip_teaser = false, $
...
+		$has_blocks = has_blocks( $content );
+	}
```
sc0ttkclark, line 37:
```suggestion
```

PR #1373, opened 2020-05-15
Element attributes to merge attributes into a string

paulmskim:
Add class for handling merging of attributes and values into html string

### DIFF
--- a/src/Tribe/Utils/Element_Attributes.php
+++ b/src/Tribe/Utils/Element_Attributes.php
@@ -1,0 +1,228 @@
+<?php
+namespace Tribe\Utils;
+
+/**
+ * Class Element_Attributes to handle HTML attributes for elements.
+ *
+ * @since  TBD
+ *
+ * @package Tribe\Utils
+ */
+class Element_Attributes {
+	/**
+	 * Store the results of parsing the attributes.
+	 *
+	 * @since  TBD
+	 *
+	 * @var array
+	 */
+	protected $results = [];
+
+	/**
+	 * Stores the arguments passed.
+	 *
+	 * @since  TBD
+	 *
+	 * @var array
+	 */
+	protected $arguments = [];
+
+	/**
+	 * Setups an instance of Element Attributes.
+	 *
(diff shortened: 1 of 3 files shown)


src/Tribe/Utils/Element_Attributes.php
```diff
@@ -0,0 +1,228 @@
...
+	protected $arguments = [];
+
+	/**
+	 * Setups an instance of Element Attributes.
+	 *
+	 * @since  TBD
+	 *
+	 * @return void
+	 */
+	public function __construct() {
```
lucatume, line 37: This will not inform about what the method does and the method cannot be called without, at the very least, one argument.

```suggestion
	public function __construct( $arg_1 ) {
```

src/Tribe/Utils/Element_Attributes.php
```diff
@@ -0,0 +1,228 @@
...
+	 *
+	 * @var array
+	 */
+	protected $arguments = [];
+
+	/**
+	 * Setups an instance of Element Attributes.
+	 *
+	 * @since  TBD
+	 *
```
lucatume, line 34: Add a note here about the function parameters.

```suggestion
	 * @param string $element One or more elements to build the validated and sanitized attributes. 
```

src/Tribe/Utils/Element_Attributes.php
```diff
@@ -0,0 +1,228 @@
...
+		foreach ( $this->results as $key => $val ) {
+			if ( ! $val && '0' !== $val ) {
+				continue;
+			}
+
+			if ( is_bool( $val ) ) {
+				$attributes[] = esc_attr( $key );
+			} else {
+				$attributes[] = esc_attr( $key ) . '="' . esc_attr( $val ) . '"';
+			}
```
paulmskim, line 117: My main concern here is around this and escaping the key + value. in `Element_Classes` we had a single attribute `class` and the function to sanitize html classes. I'm not sure what the best approach for attributes is here..

PR #1377, opened 2020-05-29
Add With_Meta_Handling trait

lucatume:
This PR adds the `With_Meta_Handling` trait to the repository, and uses it in it to
provide a re-usable way to deal with those situations where, in updates or creation,
meta provided as array should not be stored as a single array, but as multiple entries.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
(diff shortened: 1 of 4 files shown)


bordoni:
Lets add a changelog to this one.

PR #1378, opened 2020-06-03
Template legacy hookname duplicated

bordoni:


src/Tribe/Template.php
```diff
@@ -872,22 +872,30 @@ public function template( $name, $context = [], $echo = true ) {
...
+			 *
+			 * @since        4.7.20
+			 *
+			 * @deprecated   4.11.0
+			 *
+			 * @param string $file     Complete path to include the PHP File
+			 * @param array  $name     Template name
+			 * @param self   $template Current instance of the Tribe__Template
+			 */
+			do_action_deprecated(
```
lucatume, line 892: Do we have an idea about how old is *the most recent code* using the legacy hook names?
If that code is not, at least, 1.5yrs old, then I would add a todo task and not throw a deprecation notice for now.

src/Tribe/Template.php
```diff
@@ -919,22 +927,30 @@ public function template( $name, $context = [], $echo = true ) {
...
+			 *
+			 * @since      4.7.20
+			 *
+			 * @deprecated 4.11.0
+			 *
+			 * @param string $file     Complete path to include the PHP File
+			 * @param array  $name     Template name
+			 * @param self   $template Current instance of the Tribe__Template
+			 */
+			do_action_deprecated(
```
lucatume, line 947: Same note and question about code "age" as above hee.

src/Tribe/Template.php
```diff
@@ -968,23 +984,31 @@ public function template( $name, $context = [], $echo = true ) {
...
+			 * @since        4.7.20
+			 *
+			 * @deprecated   4.11.0
+			 *
+			 * @param string $html     The final HTML
+			 * @param string $file     Complete path to include the PHP File
+			 * @param array  $name     Template name
+			 * @param self   $template Current instance of the Tribe__Template
+			 */
+			$html = apply_filters_deprecated(
```
lucatume, line 1005: Same note and question about code and age here.

src/Tribe/Template.php
Camwyn: The hooks were added in TEC 5.0.0, so not 1.5 years old?

lucatume: So I would not deprecate them yet.

PR #1379, opened 2020-06-03
Common styles: Add breakpoints for common classes.

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ET-799

In ET we don't have containers, nor any JS changing them to have container based queries. Hence we can't take advantage of the common styles.

I'm adding breakpoints so we can make use of common classes in ET and other places, and effectively rely on the common design system.

### DIFF
--- a/src/resources/postcss/base/full/forms/_text.pcss
+++ b/src/resources/postcss/base/full/forms/_text.pcss
@@ -29,6 +29,12 @@
 			border: 0;
 		}
 
+		@media(--viewport-medium) {
+			@mixin desktop-body-2;
+
+			border: 0;
+		}
+
 		&::placeholder {
 			color: var(--color-text-secondary);
 			font-style: normal;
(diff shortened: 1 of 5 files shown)


paulmskim:
This will not work as expected. It might work forbET, but it'll affect TEC before the breakpoint JS adds the correct classes. We'll need to think of another way to do this.

juanfra:
Maybe a fade in effect or something for TEC and PRO. If not, I doubt having these as common is a good idea, as they only serve for TEC and PRO requirements. And we can't rely on them anywhere else.

paulmskim:
A thought might be to create a new pcss file with only the breakpoint styles and enqueue as needed. I haven't thought too much into this and what the benefits/drawbacks are, but worth a conversation with @bordoni

juanfra:
Good idea, that can make it! I'm thinking of creating `common-responsive.pcss`, using stuff from  `base/responsive/full` and `base/responsive/skeleton` and inside each one of those the same file structure as `base/full` and `base/skeleton`. Thoughts?

PR #1386, opened 2020-06-11
Fix rewrites handling on sub-directory installation.

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3460

This PR supports the one for TEC, this code is, as the rest of the
Rewrite code, indirectly tested in TEC.


paulmskim:
is there a test that can be updated to test for this case?

PR #1387, opened 2020-06-11
VE-135 Add attributes to inputs where they are missing

Camwyn:
So we can add IDs for js targets.

Correct some spelling errors. Change some of our uses of `array()` to `[]`

[VE-135]

[VE-135]: https://moderntribe.atlassian.net/browse/VE-135


src/Tribe/Settings_Tab.php
```diff
@@ -161,7 +161,7 @@ public function doContent() {
 			if ( is_array( $this->fields ) && ! empty( $this->fields ) ) {
 				foreach ( $this->fields as $key => $field ) {
 					if ( isset( $sent_data[ $key ] ) ) {
-						// if we just saved [or attempted to], get the value that was inputed
+						// if we just saved [or attempted to], get the value that was input
```
Camwyn, line 164:
```suggestion
						// If we just saved [or attempted to], get the value that was input.
```

PR #1390, opened 2020-06-12
Add With_Db_Lock trait

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-2967

This PR objective is to provide a re-usable solution to acquire and release
database locks.

Database locks are **not** the same as `LOCK TABLES` instruction and are
a way to delegate the management of locks where db data is read and written,
to the db itself.

The functions used by the trait are part of MySQL 5.6 and in line w/ WP min
requirements.

The traditional, option or transient based locks have the issue of not being
atomic, i.e. another PHP process might write and read the same value in the
meanwhile and that has proved an issue in many instances.

MySQL 5.6 has a quirk about this (see [here](https://dev.mysql.com/doc/refman/5.6/en/locking-functions.html#function_get-lock)),
where a session calling `GET_LOCK` on a lock it already holds would result in the lock being released and re-acquired.
To work around this the trait uses `IS_FREE_LOCK` that will return `1` on first call and `0` on any further call.

See [related TEC PR for an example of how the trait is used](https://github.com/moderntribe/the-events-calendar/pull/3198).

### DIFF
--- a/src/Tribe/DB_Lock.php
+++ b/src/Tribe/DB_Lock.php
@@ -152,19 +152,16 @@ class DB_Lock {
 
 		global $wpdb;
 
-		// phpcs:ignore
 		$free = $wpdb->get_var(
-			$wpdb->prepare( "SELECT IS_FREE_LOCK( SHA1( %s ) )", $lock_key )
-
+			$wpdb->prepare( 'SELECT IS_FREE_LOCK( SHA1( %s ) )', $lock_key )
 		);
 
 		if ( ! $free ) {
 			return false;
 		}
 
-		// phpcs:ignore
 		$acquired = $wpdb->get_var(
-			$wpdb->prepare( "SELECT GET_LOCK( SHA1( %s ),%d )", $lock_key, $timeout )
+			$wpdb->prepare( 'SELECT GET_LOCK( SHA1( %s ),%d )', $lock_key, $timeout )
 
 		);
 
--- a/src/Tribe/DB_Lock.php
+++ b/src/Tribe/DB_Lock.php
@@ -1,0 +1,355 @@
+<?php
+/**
+ * Manages database locks using MySQL fucntions or queries.
+ *
+ * The MySQL functions used by this class are `GET_LOCK`, `IS_FREE_LOCK` and `RELEASE_LOCK`.
+ * The functions are part of MySQL 5.6 and in line with WordPress minimum requirement of MySQL version (5.6).
+ *
+ * @see     https://dev.mysql.com/doc/refman/5.6/en/locking-functions.html#function_get-lock
+ *
+ * @since   TBD
+ *
+ * @package Tribe
+ */
+
+namespace Tribe;
+
+/**
+ * Class DB_Lock
+ *
+ * @since   TBD
+ *
+ * @package Tribe
+ */
+class DB_Lock {
+
+	/**
+	 * The prefix of the options used to manage the database lock without use of MySQL functions
+	 * in the options table.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $db_lock_option_prefix = 'tribe_db_lock_';
+
+	/**
+	 * A map, shared among all instance of this trait in the session, of the currently held locks the
+	 * time the locks where acquired, a UNIX timestamp w/ micro-seconds.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string,float>
+	 */
+	protected static $held_db_locks = [];
+
+	/**
+	 * Prunes the stale locks stored in the options table.
+	 *
+	 * @since TBD
+	 *
+	 * @return int|false The number of pruned locks, or `false` to indicate the query to prune the locks generated
+	 *                   an error (logged).
+	 */
+	public static function prune_stale_db_locks() {
+		global $wpdb;
+		$prefix        = static::$db_lock_option_prefix;
+		$affected_rows = $wpdb->query(
+			"DELETE FROM {$wpdb->options} 
+				WHERE option_name LIKE '{$prefix}%' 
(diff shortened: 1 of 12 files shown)


src/Tribe/Traits/With_Db_Lock.php
```diff
@@ -0,0 +1,120 @@
...
+
+		/**
+		 * Filters the timeout, in seconds, of the database lock acquisition attempts.
+		 *
+		 * @since TBD
+		 *
+		 * @param int    $timeout The timeout, in seconds, of the lock acquisition attempt.
+		 * @param static $this    The object that's trying to acquire the lock by means of the trait.
+		 */
+		$timeout = apply_filters( 'tribe_db_lock_timeout', 3, $this );
```
mitogh, line 48: I would add the `$lock_key` as part of the filter. 

```suggestion
		$timeout = apply_filters( 'tribe_db_lock_timeout', 3, $lock_key, $this );
```

src/Tribe/Traits/With_Db_Lock.php
```diff
@@ -0,0 +1,120 @@
...
+namespace Tribe\Traits;
+
+/**
+ * Trait With_Db_Lock
+ *
+ * @since   TBD
+ *
+ * @package Tribe\Traits
+ */
+trait With_Db_Lock {
```
mitogh, line 24:
```suggestion
trait With_DB_Lock {
```

mitogh: If we go with `DB` should we rename the file name as well?

lucatume:
After some Slack and products dev chat, I will improve this solution to fall-back on a second solution if the functions are not available.

See: https://tribe.slack.com/archives/C0SFG7E2W/p1591970985281300

lucatume:
I've reworked the logic here to:
1. Add a filter to control whether to use MySQL functions or to use queries for locks.
2. Add support for locks managed by queries in the `options` table.
3. Add a `Tribe\Service_Providers\Crons` class to try and provide the One Place where we handle crons for this general-purpose tools
4. Add a daily cron task to prune stale locks from the db.

src/Tribe/DB_Lock.php
```diff
@@ -0,0 +1,355 @@
...
+	 */
+	protected function release_db_lock_w_queries( $lock_key ) {
+		if ( ! isset( self::$held_db_locks[ $lock_key ] ) ) {
+			// Avoid sessions that do nothold the lock to release it.
+			return false;
+		}
+
+		global $wpdb;
+		$option_name = $this->get_db_lock_option_name( $lock_key );
+		//phpcs:disable
```
bordoni, line 328: Why using PHPCS here?

src/Tribe/DB_Lock.php
```diff
@@ -0,0 +1,355 @@
...
+		 * While this will not cause issues in the context of different db sessions (e.g. two diff. PHP
+		 * processes competing for a lock), it would cause issues when the lock acquisition is attempted
+		 * in the context of the same PHP process.
+		 * To avoid a read-what-you-write issue in the context of the same request, we check if the lock is
+		 * free, using `IS_FREE_LOCK` first.
+		 */
+
+		global $wpdb;
+
+		// phpcs:ignore
```
bordoni, line 155: Why Ignore?

src/Tribe/DB_Lock.php
```diff
@@ -0,0 +1,355 @@
...
+		$free = $wpdb->get_var(
+			$wpdb->prepare( "SELECT IS_FREE_LOCK( SHA1( %s ) )", $lock_key )
+
```
bordoni, line 158: Empty space?

src/Tribe/Main.php
```diff
@@ -611,6 +613,7 @@ public function bind_implementations() {
 		tribe_singleton( 'context', 'Tribe__Context' );
 		tribe_singleton( 'post-transient', 'Tribe__Post_Transient' );
 		tribe_singleton( 'db', 'Tribe__Db' );
+		tribe_singleton( 'db-lock', DB_Lock::class );
```
bordoni, line 616: Shouldn't we also start registering `DB_Lock::class` as the singleton name?

src/Tribe/DB_Lock.php
lucatume: Direct calls to `DELETE` methods or calls to the `wpdb::delete` method should be accompained by cache invalidation.
If phpcs was not disabled here, the error would be reported.
There is no caching component being managed here, so I've suppressed the warnings.

src/Tribe/DB_Lock.php
lucatume: `get_var` calls should, traditionally, be accompained by calls to invalidate/update cache; there is no caching component here, so I've suppressed them.

src/Tribe/Main.php
lucatume: I'm not sure that is the best approach. 
I would rather opt for some kind of auto-wiring in the container component that would understand that, when I say, `tribe_singleton('db-lock', DB_Lock::class)`, then a second registration should be implicitly done to register `DB_Lock` too.

src/Tribe/DB_Lock.php
bordoni: Uhn, interesting.

src/Tribe/Main.php
bordoni: I like that solution to the problem.

Camwyn: Me too - for :allthethings: singleton?

lucatume: See https://github.com/moderntribe/tribe-common/pull/1392 for a possible implementation of this.

PR #1392, opened 2020-06-15
Add implicit wiring to the container.

lucatume:
This PR changes the correct, although not intuitive, current behavior of
the `tribe` service locator and auto-wiring container.

A question, provided the code below, do you expect `$a_from_container`
and `$b->a` to be the same?

```php
class A{}

class B{

	public $a;

	public function __construct( A $a ){
		$this->a = $a;
	}
}

tribe_singleton( 'test.a', A::class );

$b = tribe( B::class );

// Is this true?
$a_from_container = tribe( 'test.a' );
assert( $b->a === $a_from_container );
```

No, they are currently not the same.
While correct, the current behavior does not fill the intention gap that sounds like this:

> I expect to have, at the most, one instance of A in the service-locator/container per request.

The intention is expressed, and here comes the question for the reviewer, in the call:

```php
tribe_singleton( 'test.a', A::class );
```

Or do we mean:

> Give me the same instance of `A` over and over when I request `some-a`, but provide me with
> new instances of `A` every time I request `A::class`.

### DIFF
--- a/src/Tribe/Container.php
+++ b/src/Tribe/Container.php
@@ -88,7 +88,7 @@ if ( ! function_exists( 'tribe_singleton' ) ) {
 			$container->singleton( $class, static function () use ( $container, $slug ) {
 				return $container->make( $slug );
 			} );
-		} elseif ( is_object( $class ) && $class instanceof \Closure ) {
+		} elseif ( is_object( $class ) && ! $class instanceof \Closure ) {
 			$obj_class = get_class( $class );
 			$container->singleton( $obj_class, static function () use ( $container, $slug ) {
 				return $container->make( $slug );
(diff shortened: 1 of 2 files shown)


bordoni:
The expected result I think is to have the same instance every time, no matter how it's called.

lucatume:
> The expected result I think is to have the same instance every time, no matter how it's called.

This is how singleton works, that expectation is already honored **once the singleton is correctly bound to all the ways it's called**.

The proposed change here is to **register additional bindings in the container, without requiring to register them explicitly**.

See code below:

```php
class B{
  public $a;

  public function __construct( A $a ){
    $this->a = $a;
  }
}

tribe_singleton( 'some-a', A::class );

$b = tribe( B::class );

// This is not true in the current implementation of the container!
assert( $b->a === tribe( 'some-a' ) );
```

Our code, currently, might or might not work exactly because `$b->a !== tribe( 'some-a' )`. That behavior might have been unknowingly implemented, but it's been implemented nonetheless.

The point of this change is to "connect" the auto-wiring done at construct time, by the container, to inject the dependencies specified in the `__construct` method with the slug-based registration used all-over the code.

PR #1401, opened 2020-07-07
TCMN-85: Update process to generate promoter keys

mitogh:
- Update process to generate promoter keys 
- Update WP Settings API to attach the promoter keys

Ticket: [TCMN-85]

[TCMN-85]: https://moderntribe.atlassian.net/browse/TCMN-85

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -6,6 +6,8 @@
 
 * Feature - Added the `tribe_normalize_orderby` function to parse and build WP_Query `orderby` in a normalized format.
 * Tweak - Adjust verbosity level to report connection issues with Promoter [PRMTR-404]
+* Tweak - Update process to generate promoter keys [TCMN-85]
+* Tweak - Register promoter key as part of the WP Settings API [CMN-85]]
 
 = [4.12.5] 2020-06-24 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -7,8 +7,8 @@
 * Feature - Added the `tribe_normalize_orderby` function to parse and build WP_Query `orderby` in a normalized format. [TEC-3548]
 * Feature - Added the `pluck`, `pluck_field`, `pluck_taxonomy` and `pluck_combine` methods to the `Tribe__Utils__Post_Collection` class to allow  more flexible result handling when dealing with ORM result sets. [TEC-3548]
 * Tweak - Adjust verbosity level to report connection issues with Promoter [PRMTR-404]
-* Tweak - Update process to generate promoter keys [TCMN-85]
-* Tweak - Register promoter key as part of the WP Settings API [CMN-85]]
+* Tweak - Update process to generate promoter keys. [TCMN-85]
+* Tweak - Register promoter key as part of the WP Settings API. [TCMN-85]
 
 = [4.12.5] 2020-06-24 =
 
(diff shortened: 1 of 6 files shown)


src/Tribe/Promoter/Auth.php
```diff
@@ -58,57 +89,24 @@ public function authorize_with_connector() {
...
-			$key = $this->generate_key();
+		$salt = wp_generate_password( 6 );
+
+		if ( defined( 'AUTH_KEY' ) ) {
+			$key = AUTH_KEY;
+		} else {
+			$key = wp_generate_password( 25 );
 		}
 
+		$key = sha1( $salt . get_current_blog_id() . $key . get_bloginfo( 'url' ) );
```
bordoni, line 106: Suggestion, instead of adding as a string, append each to an array, filter, and then `json_encode` to `sha1`. Would give someone the power to modify their base values for key.

mitogh: How about filtering the generated key instead so in case someone wants to use a custom pattern they can define a custom hashing mechanism as well.

readme.txt
```diff
@@ -7,6 +7,8 @@
 * Feature - Added the `tribe_normalize_orderby` function to parse and build WP_Query `orderby` in a normalized format. [TEC-3548]
 * Feature - Added the `pluck`, `pluck_field`, `pluck_taxonomy` and `pluck_combine` methods to the `Tribe__Utils__Post_Collection` class to allow  more flexible result handling when dealing with ORM result sets. [TEC-3548]
 * Tweak - Adjust verbosity level to report connection issues with Promoter [PRMTR-404]
+* Tweak - Update process to generate promoter keys [TCMN-85]
+* Tweak - Register promoter key as part of the WP Settings API [CMN-85]]
```
sc0ttkclark, line 11:
```suggestion
* Tweak - Update process to generate promoter keys. [TCMN-85]
* Tweak - Register promoter key as part of the WP Settings API. [TCMN-85]
```

readme.txt
```diff
@@ -7,6 +7,8 @@
 * Feature - Added the `tribe_normalize_orderby` function to parse and build WP_Query `orderby` in a normalized format. [TEC-3548]
 * Feature - Added the `pluck`, `pluck_field`, `pluck_taxonomy` and `pluck_combine` methods to the `Tribe__Utils__Post_Collection` class to allow  more flexible result handling when dealing with ORM result sets. [TEC-3548]
 * Tweak - Adjust verbosity level to report connection issues with Promoter [PRMTR-404]
+* Tweak - Update process to generate promoter keys. [TCMN-85]
+* Tweak - Register promoter key as part of the WP Settings API. [TCMN-85]
```
sc0ttkclark, line 11:
```suggestion
* Tweak - Update process to generate Promoter keys. [TCMN-85]
* Tweak - Register Promoter key as part of the WP Settings API. [TCMN-85]
```

src/Tribe/Promoter/Auth.php
sc0ttkclark: Why does this need to be filtered by people outside of Promoter's purposes? Just curious there

mitogh:
From my point of view I don't think this would be needed as this value is pretty much used only internally for Promoter. 

Unless @bordoni has a different opinion about this topic.

mitogh:
Would you mind take another look @bordoni looks like there were some .txt conflicts to be fixed before :bowing_man:

PR #1404, opened 2020-07-13
Add template tag to force printing of tribe_assets

lucatume:
Issue: https://moderntribe.atlassian.net/browse/ECP-376

This PR adds the `tribe_asset_print_group` function as a template tag
to expose the `Tribe__Assets::print_group` method.

The method supports the correct rendering of shortcodes scripts and styles
in the context of page builders by forcing the printing of the `script` (JS)
and `link` (CSS) HTML tags on the page when their "natural" condition would
not be satisfied.

For an asset registered using the `tribe_assets` function to print on the page
2 conditions must, normally, be satisfied:

1. The asset should have been registered with WordPress.
2. The asset conditional must return a truthy value.

The `tribe_asset_print_group` function will ignore both conditions to print the
required HTML tags on the page in contexts that would not satisfy both requirements.

In the specific case addressed by the issue, the scripts need to print along with a 
shortcode HTML in the context of a live preview when they have neither being registered
nor satisfy their condition.`


readme.txt
```diff
@@ -7,6 +7,7 @@
 * Feature - Added the `tribe_normalize_orderby` function to parse and build WP_Query `orderby` in a normalized format. [TEC-3548]
 * Feature - Added the `pluck`, `pluck_field`, `pluck_taxonomy` and `pluck_combine` methods to the `Tribe__Utils__Post_Collection` class to allow  more flexible result handling when dealing with ORM result sets. [TEC-3548]
 * Tweak - Adjust verbosity level to report connection issues with Promoter [PRMTR-404]
+* Tweak - Add the `tribe_asset_print_group` function to allow printing scripts or styles managed by the `tribe_assets` function in the page HTML.
```
sc0ttkclark, line 10: No TCM ticket?

lucatume: No, it was not a feature we required from common, just something that came up as part of the work on TEC.

src/Tribe/Assets.php
```diff
@@ -673,4 +673,52 @@ public function get( $slug = null, $sort = true ) {
...
+
+		// Make sure each script is registered.
+		foreach ($to_print as $slug => $data){
+			if($data->is_registered){
+				continue;
+			}
+			$data->type === 'js'
+				? wp_register_script( $slug, $data->file, $data->deps, $data->version )
+				: wp_register_style( $slug, $data->file, $data->deps, $data->version );
+		}
```
bordoni, line 711:
```suggestion
		foreach ( $to_print as $slug => $data ){
			if ( $data->is_registered ){
				continue;
			}
			'js' === $data->type
				? wp_register_script( $slug, $data->file, $data->deps, $data->version )
				: wp_register_style( $slug, $data->file, $data->deps, $data->version );
		}
```

PR #1405, opened 2020-07-13
Stop removing non-scalars so category gets passed.

Camwyn:
[FBAR-74]

In Rewrites we are removing non-scalars from the query vars.

Discussed with @lucatume who did the original pull and we couldn't think of a reason to continue to do so.

Let's see how the tests shake out.

[FBAR-74]: https://moderntribe.atlassian.net/browse/FBAR-74


src/Tribe/Rewrite.php
```diff
@@ -426,9 +426,6 @@ public function get_canonical_url( $url, $force = false ) {
 		$query         = (string) parse_url( $url, PHP_URL_QUERY );
 		wp_parse_str( $query, $query_vars );
 
-		// Drop any query var that is not a scalar; it should not be handled.
-		$query_vars = array_filter( $query_vars, 'is_scalar' );
-
 		if ( isset( $query_vars['paged'] ) && 1 === (int) $query_vars['paged'] ) {
```
jesseeproductions, line 429: What about doing recursion and going through a multidimensional array and check if the value is_scalar? @bordoni and I were discussing that idea for my solution too. Something similar to this method: 
https://github.com/moderntribe/tribe-common/blob/924035eadc6b90283e772ff62ebd411e74f40ea8/src/Tribe/Utils/Array.php#L210

bordoni:
I asked @Camwyn to add some Artifacts covering some URL management/canonical scenarios, especially to cover our bases around the change @lucatume added on [ECP-343] which gives us a good base for removal.

@jesseeproductions I dont think need the filtering if those artifacts are good.

lucatume:
I'm adding, in the TEC PR, more tests to cover the cases where we might try to build canonical URLs from pretty permalinks with multiple (Array) entries. That should remove the guesswork of "what happens if" and put us in a safer spot.

bordoni:
@Camwyn where are we with this particular ticket?

lucatume:
> @Camwyn where are we with this particular ticket?

@bordoni I think this ticket has been replaced by the work done [in this one](https://github.com/moderntribe/tribe-common/pull/1415).

PR #1415, opened 2020-07-27
Passthru, do not remove, non-scalar query args

lucatume:
Issues: 
* https://moderntribe.atlassian.net/browse/FBAR-74 
* https://moderntribe.atlassian.net/browse/FBAR-85
* https://moderntribe.atlassian.net/browse/FBAR-86

[Screencap of View change](https://drive.google.com/open?id=1AB-gPOiEF4p48eSggRbnqe-ww-Ms-Z8r&authuser=luca%40tri.be&usp=drive_fs)
[Screencap of View pagination](https://drive.google.com/open?id=1ABkBWEgqJAJH9Iw_9EQA81UuC254b6Pc&authuser=luca%40tri.be&usp=drive_fs)
[Screencap of Date Picker use](https://drive.google.com/open?id=1AFHXTXLljcPBPdNTUTeZKeem8iImGD85&authuser=luca%40tri.be&usp=drive_fs)

This is a second attempt at solving the issue that make Filter Bar
filters not be carried over from View to View when changing in any way.
In the current Rewrite code, non-scalar query vars are dropped as they
would never be correctly resolved to a "pretty" URL.
Filter Bar will use query vars like `?tribe_eventcategory[0]=23` that,
being non-scalar, would be dropped during the generation of clean URLs
from any View.

This PR changes the behavior to **not** drop those non-scalar query vars
to, simply, making them pass-through.

From my first tests, not nearly exhaustive enough, this seems to solve the issues
reported in the 3 tickets above.

Tests to cover this code are, due to the nature of the Rewrite component,
in [the TEC PR](https://github.com/moderntribe/the-events-calendar/pull/3248).

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -412,6 +412,17 @@ class Tribe__Rewrite {
 			return $home_url;
 		}
 
+		// Passthru vars are additional salts for the cache that would render it useless: parse them here.
+		$query = (string) parse_url( $url, PHP_URL_QUERY );
+		wp_parse_str( $query, $query_vars );
+		// Non-scalar value query vars should not be handled, but they should survive the resolution and not be cached.
+		$scalar_query_vars = array_filter( $query_vars, 'is_scalar' );
+		$passthru_vars     = array_diff_key( $query_vars, $scalar_query_vars );
+		// Remove the passthru query vars from the URL to match the correct cache.
+		$url = remove_query_arg( array_keys( $passthru_vars ), $url );
+		// Normalize the URL to make sure there's a trailing slash at the end of the path, before the query or fragment.
+		$url = preg_replace( '~(?<!/)([?#])~', '/$1', $url );
+
 		if ( ! $force ) {
 			$this->warmup_cache(
 				'canonical_url',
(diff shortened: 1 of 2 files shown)


src/Tribe/Rewrite.php
```diff
@@ -589,6 +591,11 @@ public function get_canonical_url( $url, $force = false ) {
 			$this->canonical_url_cache[ $url ] = $resolved;
 		}
 
+		if ( count( $passthru_vars ) ) {
+			// If there are pass-through query vars, then re-apply them to the clean URL after it's been cached.
+			$resolved = add_query_arg( $passthru_vars, $resolved );
```
Camwyn, line 596: By doing this here - won't it cache the URL with no passthrough vars, then later return the cached URL w/o them since it's cached w/o them?

lucatume: Excellent point, I've addressed it in my latest commit and improved the code a bit, thanks.

lucatume:
I've addressed @Camwyn's comment about caching (he was correct about it), and improved the rewrite code a bit increasing the chance of cache hits happening.

[New Screencast](https://drive.google.com/open?id=1ASRYsvkV0Bjz5M4Do34bVKuROrqNGFoT&authuser=luca%40tri.be&usp=drive_fs)

sc0ttkclark:
Nevermind, just saw https://github.com/moderntribe/the-events-calendar/pull/3248

lucatume:
> I still thing some generalized tests for this outside of TEC context would be nice, but this looks good!

I agree with you here, the issue here is that that piece of code specifically relies on "our" rewrite rules. 
Setting up tests to cover it would provide more robustness, but covering this code in the plugins using it (TEC and PRO currently) has the best ROI.

PR #1416, opened 2020-08-04
Toggle > Make the toggle compatible with IE11.

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ET-865
🎥 https://www.loom.com/share/b51136bb31224a61856f6a2bf64f557b

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [TBD] TBD =
 
-* Fix - Added IE11 compatibility for the toggles styles. [ET-865]
+* Fix - Added IE11 compatibility for the toggles styles using `tribe-common-form-control-toggle` CSS class. [ET-865]
 
 = [4.12.6] 2020-07-27 =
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Added IE11 compatibility for the toggles styles. [ET-865]
```
bordoni, line 7: Lets try to make it more clear to which elements this applies to, maybe:

```suggestion
* Fix - Added IE11 compatibility for the toggles styles using `tribe-common-form-control-toggle` HTML class. [ET-865]
```

paulmskim:
looks good to me, just the one thing G suggested

readme.txt
juanfra: Wouldn't it be CSS class in that case?

PR #1417, opened 2020-08-04
Draft common link object for discussion

Camwyn:
Utility object for link building.

"Intelligently" adds required attributes (rel, target) as needed.

Allows passing extra arguments as an array in the format:
```php
[
    'media'         => 'screen', // key => value for most attributes.
    'disabled      => true, // key => true for  no-value attributes.
    'disabled      =>'', // key => '' also works for no-value attributes.
    'crossorigin' => false, // key => false for forcing removal of attributes.
]
```

Also added snapshot tests to common so I could use them.

Note, this would make https://github.com/moderntribe/tribe-common/pull/1383 obsolete. This is more robust.


borkweb:

PR #1434, opened 2020-09-02
Add support for common fallback and aliases

lucatume:
Issue: https://moderntribe.atlassian.net/browse/FBAR-148 

This PR adds two features to the `Tribe__Template` class, both triggered by needs of the work done one FBAR in `release/B20.09`:

1. Add support for a "common fallback"; what this means is that the template classes will look up, as usual, the plugins to find the template files, but this PR adds a further fallback location, at priority `100`, in Common. This will allow some templates to be shipped with Common and used across plugins without the need for duplicated code. Such is the case, for example, for SVG icons.
2. Add support for "aliases"; as we use the View/Template system more and more we're coming across the need to version the templates and to do so while still leveraging the existence of a previous, compatible version. E.g. in FBAR we are working with `v2_1` that overhauls most, but not all templates and, furthermore, would like to load the correct templates from those; so from a `v2_1` template we might want to look for a `component/test` template that is, in TEC as an example, in the `views/v2/components/test.php` file. This feature would allow us not to have to duplicate the code of that component and provides a nice API for extending template classes to say "look for templates in the `v2_1` directory of plugins, but look into `v2` too if you cannot find the template`.

Artifact: there is really no artifact for this work in terms of screencasts or pictures, please look at the test code!


jesseeproductions (dismissed):

jesseeproductions:
Looks good.

paulmskim (dismissed):

readme.txt
```diff
@@ -5,6 +5,8 @@
 = [4.12.9] TBD =
 
 * Tweak - Move border style button styles to border-small and add various border button styles that match the solid button style. [FBAR-143]
+* Tweak - Add the common views folder to the `Tribe__Template` lookup folders, the folder will be searched for matching template files only if no plugin-provided template was found. [FBAR-148]
+* Tweak - Add the `tribe_template_common_path` filter to allow controlling the path of the template file provided by common [FBAR-148]
```
sc0ttkclark, line 9:
```suggestion
* Tweak - Add the `tribe_template_common_path` filter to allow controlling the path of the template file provided by common. [FBAR-148]
```

sc0ttkclark (dismissed):

PR #1447, opened 2020-09-22
[ECP-458] Register Block Editor translations when available

lucatume:
Issue: [ECP-458]

This PR leverages the `wp_set_script_translations` function to allow scripts to localize strings in Block Editor context.

[ECP-458]: https://moderntribe.atlassian.net/browse/ECP-458


src/Tribe/Assets.php
```diff
@@ -443,6 +448,8 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
 			// Bigger Variables at the end.
 			'localize'      => [],
 			'conditionals'  => [],
+
```
sc0ttkclark, line 451: separated by a line but no specific comment?

sc0ttkclark (commented):
This will be great to have!

src/Tribe/Assets.php
```diff
@@ -443,6 +448,8 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
...
+			'translations'  => [],
```
bordoni, line 452:
```suggestion
			// Used to handle Translations handled in the JavaScript side of the Assets.
			'translations'  => [],
```

bordoni (approved):

sc0ttkclark (changes requested):
Needs changelog, should this get a test too?

bordoni:
I am super glad @sc0ttkclark asked for the tests here, it sent me down a rabbit hole of why the actual translations were not loading.

We will need to include a new process into our packager, that for each translation file in a certain folder generate `JSON` file using [`wp i18n make-json`](https://developer.wordpress.org/cli/commands/i18n/make-json/).

The way that script works specifically is that it generates a diff JSON for each JS file that has translations as the origin it identified the file origin based on the comment inside of the `.po` file it digested, E.g.:

```
#: test-script-1.js:1
msgid "Translations MSGID"
msgstr "Translations MSGSTR"
```

So we will need to change the code here to point all of the translations in a single folder for uniformity, likely we won't allow via the Tribe Assets registering a separate domain then the origin plugin or it's the default path for translations, which simplifies the registration process of assets to a boolean variable like `include_translations` or `has_translations`.

cc @lucatume @borkweb  

We will need to chat about this tomorrow for more context and a path forward.

tests/wpunit/Tribe/AssetsTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+class AssetsTest extends \Codeception\TestCase\WPTestCase {
+	/**
+	 * @test
+	 *
+	 * @since TBD
+	 */
+	public function it_should_have_translations() {
+		$locale = 'en';
+		add_filter( 'pre_determine_locale', function() use ( $locale ) {
```
lucatume, line 15:
```suggestion
		add_filter( 'pre_determine_locale', static function() use ( $locale ) {
```

tests/wpunit/Tribe/AssetsTest.php
```diff
@@ -0,0 +1,50 @@
...
+use Tribe__Main as Plugin;
+
+class AssetsTest extends \Codeception\TestCase\WPTestCase {
+	/**
+	 * @test
+	 *
+	 * @since TBD
+	 */
+	public function it_should_have_translations() {
+		$locale = 'en';
```
lucatume, line 14: Since English is the default language, I would add a data provider for this test method and provide the language for it, say `en` and another `de`.
Scaffold the translation files for the second language (`de`) in my example and test those too.

While the fallback/default English logic might keep working, the logic for actual translation might be not down the road and I think this will protect the code a bit more from regression.

bordoni: I will add that I feel like it creates a bigger safety net, but the default language is not English, it's empty string. Looking for `en` is not the default I mean.

lucatume: Ah, thanks. I did not know that.
The strings are, by default in our code, in English and that could obscure a non-passing test.

lucatume:
I've refreshed this PR and added the changelog entry.

bordoni (approved):
It is ready to merge, but we will need to make changes to our packaging for this to fully work. Which we didnt have time yet to tackle

PR #1450, opened 2020-10-02
Don't nest the functions!

Camwyn:
Noticed that `tribe_upload_image` was nested inside the check for `tribe_normalize_terms_list`

That didn't make sense - so here's a fix 😉 (looks like a merge error)


lucatume (approved):
Took me a while to understand the issue, this is a great catch, thanks.

sc0ttkclark (approved):

paulmskim (approved):

PR #1455, opened 2020-10-14
POC - filter detection

Camwyn:
This is dumb code - more a POC than anything else. 

This will list out all filters in the `tribe_support_registered_filter_overrides` array that have hooks, as well as how many times they've been hooked.

📷 http://p.tri.be/ZmXxSk

**See:** https://docs.google.com/document/d/1VAolTyAvHmuXHRDaNhEyHOCvMTQXsajdI8pJXEsc-9w/


src/Tribe/Support/FIlter_Checker_Report.php
```diff
@@ -0,0 +1,126 @@
...
+		 *         filter_slug,
+		 *         another_filter_slug,
+		 * 		],
+		 *
+		 * @var array $plugin_filters
+		 */
+		return apply_filters(
+			'tribe_support_registered_filter_overrides',
+			[
+				'the-events-calendar' => [
```
lucatume, line 53: This seems a maintenance nightmare that will either end up with this being abandoned or this code never being used.
Why not make this dynamic and collect all the stuff attached to any filter starting with `tribe_`; that is consistent in our code and would not require maintenance.

lucatume (commented):
If this runs in the context of the Help page (admin, and a specific page) then many filters that would be attached in the context of a front-end request will not be attached at all making the list provided by this page of little value.

I like the idea (knowing what filters have something attached or not), but not this implementation.

What if we move to Debug Bar? There are similar panels in Common (search the Debug Bar directory) and they will show up in Query Monitor.
That would allow us to have that list to show what is *actually* acting on filters on which page.

Verbiage: nothing is "overridden" as much as "filtering" or "attached" to a filter/action, I would change the wording of methods and comments to reflect that.

src/Tribe/Support/FIlter_Checker_Report.php
Camwyn: That's the idea - I just put in some dummy data for now to illustrate and test.

sc0ttkclark: We also know that our plugins themselves extend each other, so we wouldn't want to count that as an "override" from the customer perspective.

src/Tribe/Support/FIlter_Checker_Report.php
```diff
@@ -0,0 +1,126 @@
+<?php
```
sc0ttkclark, line 1: Typo on file name with capitalization?

PR #1457, opened 2020-10-15
Add the tribe_suspending_filter function

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3587

This PR adds the `tribe_suspending_filter` function to be able to run a callback while removing and the re-attaching a callback to a filter.

The use case that originated this is like this:

```php
function make_full(){
		return greet() . ' world';
}

function hello(){
		apply_filter( 'greet', 'hello' );
}

add_filter( 'greet', 'hello' );
add_filter( 'greet', 'make_full' );

echo apply_filters( 'greet',  );
```

Following the code it's easy to understand that, when applying the `greet` filter, then `make_full` will call `greet` that will apply the same filter again, calling `make_full` and so on for an infinite loop.

To avoid this, in the `make_full` function, we can apply a logic like "return the greeting without applying myself":

```php
function make_full(){
		return tribe_suspending_filter(
			$for_filter = 'greet',
			$suspend = __FUNCTION__,
			$and_run = static function(){
				return greet() . ' world';
			}
		);
}

function hello(){
		apply_filter( 'greet', 'hello' );
}

add_filter( 'greet', 'hello' );
add_filter( 'greet', 'make_full' );

echo apply_filters( 'greet',  );
```

Leaving alone the absurdity of the example, this is quite useful as we already do something similar in other parts of the plugins.

The real case is one where `tribe_get_events` will serialize itself when real object caching is active and that will lead, in turn, to the early resolution of callbacks that might, in turn, call `tribe_get_event` in their body.


tests/wpunit/Tribe/functions/utilsTest.php
```diff
@@ -448,4 +448,53 @@ public function should_not_detach_filters_that_are_not_in_the_list_of_filters_to
...
+		$this->assertEquals( 'two', tribe_suspending_filter(
+				'test_filter',
+				$callback_one,
+				function () use ( $callback_one, $callback_two ) {
+					$this->assertFalse( has_filter( 'test_filter', $callback_one ) );
+					$this->assertEquals( 13, has_filter( 'test_filter', $callback_two ) );
+
+					return apply_filters( 'test_filter', 'original', ...range( 1, 10 ) );
+				},
+				3 )
```
bordoni, line 484: too many tabs.

tests/wpunit/Tribe/functions/utilsTest.php
```diff
@@ -448,4 +448,53 @@ public function should_not_detach_filters_that_are_not_in_the_list_of_filters_to
...
+		$this->assertEquals( 'one', tribe_suspending_filter(
+				'test_filter',
+				$callback_two,
+				function () use ( $callback_one, $callback_two ) {
+					$this->assertEquals( 23, has_filter( 'test_filter', $callback_one ) );
+					$this->assertFalse( has_filter( 'test_filter', $callback_two ) );
+
+					return apply_filters( 'test_filter', 'original', ...range( 1, 10 ) );
+				},
+				3 )
```
bordoni, line 497: too many tabs.

bordoni (approved):
Very minor stuff. But good to go.

I like the idea, sounded a bit confusing at first but it makes sense.
Only wonder about the name of the method, it's good but that was what confused me at first.

lucatume:
> Only wonder about the name of the method, it's good but that was what confused me at first.

I struggled a bit to find a name for the function and, finally, settled on one that would follow the convention set by the `tribe_without_filters` function.

PR #1458, opened 2020-10-15
Fix issue with Select2 in The Events Calendar.

mmcachran:
This fixes an issue in The Events Calendar when the Select2 container cannot be found.

**Original Error Message:**
Screen Shot 2020-10-15 at 1 36 59 PM


sc0ttkclark (commented):
Looks good to me, though I'm not entirely sure why TEC would call this and not have the element there. Might be good for follow up before this moves forward.

lucatume:
I've changed the assignee from myself to @bordoni as this is more of his area of operations.

src/resources/js/dropdowns.js
```diff
@@ -449,12 +449,19 @@ var tribe_dropdowns = window.tribe_dropdowns || {};
...
+			$fieldContainer = $( field );
+		}
+
+		if ( $fieldContainer ) {
+			// Propagating original input classes to the select2 container.
+			$fieldContainer.addClass( obj.getSelectClasses( $select ).join( ' ' ) );
 
-		// Propagating original input classes to the select2 container.
-		$container.data( 'select2' ).$container.removeClass( 'hide-before-select2-init' );
+			// Propagating original input classes to the select2 container.
```
lucatume, line 462: This comment should, probably, be removed or modified.
```suggestion

```

lucatume (changes requested):

paulmskim:
This is an issue of having 2 different versions of select2 loading. We run select2 4.X, but when select2 3.X is loaded as well, it creates this problem. We may need to look into how to prevent having more than 1 version of select2 loading at the same time or deal with potential conflicts gracefully.

It's known that ACF loads select2 3.X in earlier versions, and updating to a later version of ACF fixes this. This could very well also be the case with other plugins.

bordoni:
This problem will be resolved by #1573

PR #1452, opened 2020-10-08
Add return type to Canonical_Formatter::format() (to stop promoter breaking Woocommerce Checkout)

justlevine:
Tl;dr adds return type of `string` to LineFormatter::format, so it matches what monolog expects.

`Declaration of Tribe\Log\Canonical_Formatter::format(array $record) must be compatible with Monolog\Formatter\LineFormatter::format(array $record): string`.

---
Longer version:

The newer versions of Event Tickets Plus was breaking Woocommerce, because Promotor was throwing an exception. `['ticket_purchased', 'License Key must be present].

Normally this should be ignored (I for example use the nonprofit license, so we dont have promoter), but issue with monolog were causing a fatal error:

`PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Monolog\Handler\ErrorLogHandler::__construct() must be of the type int, null given, called in /var/www/thegsc/wp-content/plugins/the-events-calendar/common/src/Tribe/Log/Service_Provider.php on line 73.`

That's a separate issue, and i'm not familiar enough with Tribe's Logger, but I changed ` new ErrorLogHandler( null, $level_threshold );` to `new ErrorLongHandler( 0, $level_threshold)`, since [thats the default value provided by Monolog](https://github.com/Seldaek/monolog/blob/bb99e4c699bd5abe2609cace3c2daf0e45fdb673/src/Monolog/Handler/ErrorLogHandler.php#L25)...

Then I got hit with another error, and that's what this fix addresses

```
PHP Fatal error:  Declaration of Tribe\Log\Canonical_Formatter::format(array $record) must be compatible with Monolog\Formatter\LineFormatter::format(array $record): string in /var/www/thegsc/wp-content/plugins/the-events-calendar/common/src/Tribe/Log/Canonical_Formatter.php on line 27
[08-Oct-2020 16:30:43 UTC] PHP Stack trace:
[08-Oct-2020 16:30:43 UTC] PHP   1. {main}() /var/www/thegsc/index.php:0
[08-Oct-2020 16:30:43 UTC] PHP   2. require() /var/www/thegsc/index.php:17
[08-Oct-2020 16:30:43 UTC] PHP   3. require_once() /var/www/thegsc/wp-blog-header.php:19
[08-Oct-2020 16:30:43 UTC] PHP   4. do_action() /var/www/thegsc/wp-includes/template-loader.php:13
[...]

PR #1461, opened 2020-10-20
Allow our Monolog implementation to be 1.x and 2.x compatible | TEC-3638

borkweb:
### Core issue

Our Monolog implementation assumes Monolog 1.x. WooCommerce Bookings (the plugin whose conflict was flagged most recently) includes Monolog 2.x. Our implementation is not compatible with 2.x due to a couple of reasons.

#### Improper instantiation

Our instantiation of the `Monolog\Handler\ErrorHandler` is incorrect as per the docblock in both versions 1.x and 2.x of Monolog. The first argument should be an `int` and we were passing `null`. With Monolog 2.x, the type hinting enforces the use of an `int` where we should have been using one all along.

#### Our `Canonical_Formatter`

We appropriately extend the `Monolog\Formatter\LineFormatter` class for our `Canonical_Formatter` class—which is used when instantiating the `Monolog\Handler\ErrorLogHandler` object our `Log\Service_Provider`. However, extending the class as we do makes our `Canonical_Formatter` incompatible with 2.x.

We cannot update to Monolog 2.x (whose minimum PHP version is 7.2) and the Monolog 1.x & 2.x method signatures for the `LineFormatter` class are incompatible.

The fix that I have put in place allows us to skirt around the issue of method signatures between Monolog versions by changing `Canonical_Formatter` to instantiate `LineFormatter` rather than extending it.

### Artifacts

**The Monolog-related tests are passing and that should serve as our validation for the Monolog-related logging behaviors.**

For the display of the expected functionality as per [TEC-3638], here's a screenshot of the rendered event (rather than a WP Fatal):

Selection_066

See: [TEC-3638]
Related: #1452

[TEC-3638]: https://moderntribe.atlassian.net/browse/TEC-3638


sc0ttkclark (approved):

borkweb:
@sc0ttkclark, can you give this another look-see?

sc0ttkclark (approved):
Hey you're good at this or somethiny

bordoni (approved):

bordoni:
Damn, that is a good find. I did not think of this as possible, I looked at the two diff implementations and assumed. Thanks for finding this.

src/Tribe/Log/Canonical_Formatter.php
```diff
@@ -11,9 +11,37 @@
 namespace Tribe\Log;
 
 
+use Monolog\Formatter\FormatterInterface;
 use Monolog\Formatter\LineFormatter;
 
-class Canonical_Formatter extends LineFormatter {
+class Canonical_Formatter implements FormatterInterface {
```
lucatume, line 17: Composition over inheritance, one can rarely go wrong. Thanks for this fix.

lucatume (approved):

PR #1466, opened 2020-10-22
Allow removing non-persistent cache data

rafsuntaskin:
This alters the `delete` method to allow the removal of non-persistent data.

Related PR: https://github.com/moderntribe/event-tickets/pull/1805


lucatume (changes requested):
This seems like a nice idea, but it will require test coverage.

lucatume (approved):
Nice! And tests too, delightful.

sc0ttkclark:
Added changelog after-the-fact on my side

PR #1475, opened 2020-11-10
Context: Fallback to Global if Local is true but value of Local is empty

cliffordp:
[ETP-471]

For https://github.com/moderntribe/event-tickets-plus/pull/1089

[ETP-471]: https://moderntribe.atlassian.net/browse/ETP-471


sc0ttkclark (changes requested):
I sent my suggestion via Slack for this function with a better overall solution that uses inheritance for local >> global >> default handling. If `$is_local` is false, it will just be global >> default.

Slack convo: https://tribe.slack.com/archives/CHXD8BSH4/p1604981380349500?thread_ts=1604967404.339900&cid=CHXD8BSH4

sc0ttkclark:
Tests can go into https://github.com/moderntribe/tribe-common/blob/tests/ETP-471-v2-ar/tests/wpunit/Tribe/TemplateTest.php

PR #1477, opened 2020-11-10
fix(Cache.php) do not try to store strings larger than packet size

lucatume:
Issue: https://moderntribe.atlassian.net/browse/TEC-3615 

Artifacts: see tests. Due to the nature of this issue the test should happen on the sandbox, where we know we can kill the Db before killing PHP (due to lack of memory to handle the string).

The fix in this PR is quite simple in nature: if not using external object caching (e.g. Redis); then we should **not** try to write to the db, by means of a `set_transient` call, values that are larger than the set `max_allow_packet` size.

The tests simulate each part of the involved pieces making sure the check is made at the root level (in the `Tribe__Cache` object) and does not fall on the client code.
This means that, in the instance that created the issue, it does not fall on Month View to check the size of its HTML, but rather on the Cache abstraction that will try to cache the HTML string only if possible.

This will add a layer of protection in those instances where Month View Cache is active, there are many events (hundreds, at least) in the View and external object caching is **not** set up.
The above ingredients should never co-exist, but it might happen and we can mitigate the issue.


src/Tribe/Cache.php
```diff
@@ -496,4 +500,41 @@ public function warmup_post_caches( $post_ids, $update_post_meta_cache = false )
...
+			$size             = strlen( $serialized_value );
+		} catch ( Exception $e ) {
+			// The underlying function would run into the same issue, bail and do not set the transient.
+			return true;
+		}
+
+		/** @var Tribe__Feature_Detection $feature_detection */
+		$feature_detection = tribe( 'feature-detection' );
+
+		// If the size of the string si above 90% of the database `max_allowed_packet` setting, then it should not be written to the db.
```
paulmskim, line 537:
```suggestion
		// If the size of the string is above 90% of the database `max_allowed_packet` setting, then it should not be written to the db.
```

readme.txt
```diff
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Do not try to store too large values in transients when not using external object cache. [TEC-3615]
```
```suggestion
* Fix - Do not try to store too large of values in transients when not using external object cache. [TEC-3615]
```

src/Tribe/Cache.php
```diff
@@ -87,7 +87,11 @@ public function set( $id, $value, $expiration = 0, $expiration_trigger = '' ) {
 	 * @return bool
 	 */
 	public function set_transient( $id, $value, $expiration = 0, $expiration_trigger = '' ) {
-		return set_transient( $this->get_id( $id, $expiration_trigger ), $value, $expiration );
+		if ( $this->data_size_over_packet_size( $value, $serialized_value ) ) {
```
sc0ttkclark, line 90: Not seeing $serialized_value set up here

readme.txt
Camwyn:
```suggestion
* Fix - Do not try to store overly large values in transients when not using external object cache. [TEC-3615]
```

tests/wpunit/Tribe/CacheTest.php
```diff
@@ -186,4 +186,111 @@ public function should_not_clean_transients_more_than_once_per_request_when_trig
 
 		$this->assertEquals( 1, $passed );
 	}
+
+	/**
+	 * It should not cache too large strings in transients
```
Camwyn, line 191:
```suggestion
	 * It should not cache overly large strings in transients
```

tests/wpunit/Tribe/CacheTest.php
```diff
@@ -186,4 +186,111 @@ public function should_not_clean_transients_more_than_once_per_request_when_trig
...
+	 *
+	 * @test
+	 */
+	public function should_not_cache_too_large_strings_in_transients() {
```
Camwyn, line 195:
```suggestion
	public function should_not_cache_overly_large_strings_in_transients() {
```

Camwyn (changes requested):
Just some wording tweaks - and the missing praam @sc0ttkclark noted.

src/Tribe/Cache.php
lucatume: No need to. It's set by reference, like the `$matches` in `preg_match`.

lucatume: You can test this out on a PHP 5.6 machine with this:

```bash
php -r 'preg_match("/foo/", "foo bar", $m) && print_r($m);'
```

sc0ttkclark: OK so there's no PHP notices calling an undefined variable at this line then?

lucatume: I'm double-checking to make sure, spinning up now.

lucatume: Confirmed, no errors on a PHP 5.6 machine and 7.2 one.

sc0ttkclark: ✅

sc0ttkclark (approved):

bordoni (approved):

PR #1481, opened 2020-11-12
Correctly handle query arguments merge

lucatume:
Issue: https://moderntribe.atlassian.net/browse/ECP-357 

This fix originates from the issue above; while investigating it I ran into some issues
while trying to set up Week View and investigated.
	
The issues originated from mis-handling, in the Repository code, of the merging of query
args leading to wrongly built `meta_query` entries like this one:

```php
$args = [
	'meta_query' => [
		'_eventhiddenfromupcoming_not_exists => [
			'key' => [
				'_foo',
				'_foo',
			],
			'compare' => [
				'NOT EXISTS',
				'NOT EXISTS',
			],
			'value' => [
				'n/a',
				'n/a',
		],
	],
];
```

Each entry of that array should NOT be an array, of course.
That would be caused by simple code like this:

```php
tribe_events()->where('meta_not_exists', '_foo)->where('meta_not_exists', '_foo')->get_ids();
```

While the double call to `where` is nonsense, that might happen (and was happening) and our code should
survive it. It did not and would create wrongly built entries.

Look at the tests to have a better idea.

The PR adds another set of methods to the `Tribe__Utils__Array` class that leverage Closures and
the visitor pattern.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,11 @@
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Correctly handle mutltiple calls to the Repository `by` or `where` method that would cause issues in some Views [ECP-357]
+* Tweak - Add the `Tribe__Utils__Array::merge_recursive_query_vars` method to correctly recursively merge nested arrays in the format used by `WP_Query` [ECP-357]
+
 = [4.12.12] 2020-10-22 =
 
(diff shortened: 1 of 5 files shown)


sc0ttkclark (dismissed):

sc0ttkclark:
Great work!

bordoni (approved):

PR #1491, opened 2020-11-18
Add an auto-enabled Black Friday banner using our Notices system | TCMN-90 | ET-962 | TEC-3679

borkweb:
A couple of things are being done in here:

1. Removing old marketing notices that no longer apply (BF2018 and the Gutenberg release)
2. Add a new BF marketing notice with a `type` of `tribe-banner` and a priority of `-1` so that it is the first notice that is output and appears at the top of the page as a banner. (except on the Settings page where it looks different)

This stands on the shoulders of our existing Notices system and the Black Friday banner will follow the following rules for output:

1. If the `TRIBE_HIDE_UPSELL` constant is set to `true`, the banner will not display.
2. If the user has previously dismissed the banner, the banner will not display.
2. If the current UTC time falls within the dynamically chosen Black Friday date range (Monday of the week of Thanksgiving at 11am UTC until December 1st at 5am UTC), the banner _will_ display.

**Note:** the timestamps are filterable and QA will need to filter the values to test the notice. Here's my filter:
**Also note:** I don't have full control over all of the class names used within the notices system without making this a bigger change than it should be, but the classes within the marketing notices have been updated to use BEM.

```php
add_filter( 'tribe_black_friday_start_time', function() {
	return time();
} );
```

:movie_camera: http://p.tri.be/dvufuM
:ticket: [TCMN-90] [TEC-3679] [ET-962]

[TCMN-90]: https://moderntribe.atlassian.net/browse/TCMN-90
[TEC-3679]: https://moderntribe.atlassian.net/browse/TEC-3679
[ET-962]: https://moderntribe.atlassian.net/browse/ET-962


bordoni (dismissed):

src/Tribe/Admin/Notice/Marketing.php
```diff
@@ -36,179 +39,137 @@ public function __construct() {
...
-		 * Allows filtering of the default Gutenberg Release Notice's end time, mainly for testing purposes.
-		 *
-		 * @since 4.7.23
-		 *
-		 * @param int $gutenberg_release_end_time Defaults to Nov 17 2018 @ midnight, California time.
-		 */
-		return apply_filters( 'tribe_gutenberg_release_notice_end_time', 1542434400 );
-	}
+		// If not a valid screen, don't display.
+		if ( empty( $current_screen->id ) || ! in_array( $current_screen->id, $screens ) ) {
```
sc0ttkclark, line 139:
```suggestion
		if ( empty( $current_screen->id ) || ! in_array( $current_screen->id, $screens, true ) ) {
```

src/admin-views/notices/tribe-bf-general.php
```diff
@@ -0,0 +1,22 @@
...
+ * @var string $icon_url The local URL for the notice's image.
+ * @var string $cta_url The short URL for black friday.
+ */
+?>
+<div class="tribe-marketing-notice">
+	<div class="tribe-marketing-notice__icon">
+		<img src="<?php echo esc_url( $icon_url ); ?>"/>
+	</div>
+	<div class="tribe-marketing-notice__content">
+		<h3>Save 40% on Every. Single. Plugin.</h3>
```
sc0ttkclark, line 16: Confirming -- not interested in translations, correct?

sc0ttkclark (changes requested):
Minor tweak

sc0ttkclark (dismissed):

src/admin-views/notices/tribe-bf-general.php
borkweb: Correct! It will likely change each year in some capacity, so forcing folks to translate a splash-in-the-pan marketing message feels less than ideal.

src/Tribe/Admin/Notice/Marketing.php
borkweb: I think that tweak is a bit superfluous as the values being checked cannot be interpreted as anything other than strings. That third parameter really matters when dealing with truthy values, numbers, etc.

sc0ttkclark: I agree but phpcs catches it and every time I see it I enforce the third parameters so it never triggers a notice about it.

PR #1493, opened 2020-11-18
Function to add tribe-dependency attributes to admin elements

Camwyn:
adds `tribe_format_field_dependency()` which is used in the list widget `get_admin_data()` so we can add dependencies to admin form items.


src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
 		}
 	}
 }
+
+function tribe_format_field_dependency( $deps ) {
```
lucatume, line 136: Add a function doc block.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+    // Sanity check.
+    if ( empty( $deps ) ) {
+        return '';
+	}
```
lucatume, line 140: The alignment of this brace is wrong.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+        'is-empty',
+        'is-not-empty',
+        'is-numeric',
+        'is-not-numeric',
+        'is-checked',
+        'is-not-checked',
+    ];
+
+    foreach ( $deps as $attr => $value ) {
+        // Attributes are always lower case.
```
lucatume, line 168: Since you're assuming lower-case names here, I would just make all keys lowercase at the start of the function:

```php
$deps = array_combine( array_map( 'strtolower', array_keys( $deps ) ), $deps );
```

And not bother with upper or lowercase keys again in the method.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+        $attr = strtolower( $attr );
+
+        // Valid dependency check.
+        // Note this means the dependency has to begin with "is", except for "id"!
+        if ( ! in_array( $attr, $accepted ) ) {
```
lucatume, line 173: This is a good example of "filtering while interpolating" that could be avoided.

```php
$valid_deps = array_intersect_key( $deps, array_flip( $accepted ) );
```

And then iterate on the result; it's a small optimization, but it makes for leaner code and does not move a full array around.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+            continue;
+        }
+
+        // Handle the ID component.
+        if ( 'id' === $attr ) {
+			// "Missing" quotes is intentional - they are added automagically and we want to avoid doubles.
```
lucatume, line 179: There is something wrong in this alignment.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+            $dependency .= " data-depends={$value}";
+            continue;
+        }
+
+        // Handle boolean values.
+        if ( is_bool( $value ) ) {
+            if ( $value ) {
+                $dependency .= " data-{$attr}";
+            } else {
+                if ( false !== stripos( $attr, 'is-not-' ) ) {
```
lucatume, line 189: We're checking if `is-not-` is *at the start* of the string, not if *is somewhere in the string*.

Change to `stripos( $attr, 'is-not-') === 0`.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+                }
+
+                $dependency .= " data-{$attr}";
+                continue;
+            }
+
+            continue;
+        }
+
+		// Handle string and "empty" values
```
lucatume, line 202: Alignment.

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,83 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+
+        // Handle the ID component.
+        if ( 'id' === $attr ) {
+			// "Missing" quotes is intentional - they are added automagically and we want to avoid doubles.
+            $dependency .= " data-depends={$value}";
+            continue;
+        }
+
+        // Handle boolean values.
+        if ( is_bool( $value ) ) {
```
lucatume, line 185: Would this be covering values that are still considered booleans, but are not booleans, e.g. `0` or `1`?

lucatume (changes requested):
Updates to the function and tests.

src/functions/template-tags/html.php
Camwyn: No - since we could be checking that a value is (int|string) 1  or (int|string) 0 I didn't want to convert.
ie "Show this section if the value of #id is exactly 1, but not if it's 0 or 2"?

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,84 @@ function tribe_disabled( $disabled, $echo = true ) {
 		}
 	}
 }
+
+/**
+ * Generates a string for the Tribe-dependency attributes
```
sc0ttkclark, line 137:
```suggestion
 * Generates a string for the tribe-dependency attributes.
```

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,84 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+ *
+ * @since TBD
+ *
+ * @param array<string,mixed> $deps
+ * @return void
```
sc0ttkclark, line 142: This isn't correct

sc0ttkclark (changes requested):
Some tweaks here I think, also a test would be fantastic

Camwyn:
@sc0ttkclark Added tests for bool/int/string values - I could create about 2 dozen with all the potential combinations, not sure it's worth it?

sc0ttkclark:
I think bool/int/string are suitable here, float might come through a few places. Outside of that, I would assume that as a person using this function that I need to provide those types of values, not arrays or objects. Sounds safe!

src/functions/template-tags/html.php
```diff
@@ -132,3 +132,85 @@ function tribe_disabled( $disabled, $echo = true ) {
...
+
+				$dependency .= " data-{$attr}";
+				continue;
+			}
+
+			continue;
+		}
+
+		// Handle string and "empty" values
+		// "Missing" quotes is intentional - they are added automagically and we want to avoid doubles.
```
sc0ttkclark, line 205:
```suggestion
		/*
		 * Handle string and "empty" values.
		 *
		 * "Missing" quotes is intentional - they are added automagically and we want to avoid doubles.
		 */
```

sc0ttkclark: Are we 100% sure here, especially with escaping and values that may include spaces?

Camwyn: Let me fiddle with it a bit more, but the dependencies were getting double-quoted between here and the template output. This was my solution for that.

Camwyn: Think I've found a better solution...?

sc0ttkclark: ✅

src/functions/template-tags/html.php
lucatume: Ok, thanks for the clarification.

tests/wpunit/Tribe/functions/templateTags/dependenciesTest.php
```diff
@@ -0,0 +1,66 @@
...
+	public function it_should_return_correct_attributes_for_boolean() {
+		$deps = [
+			'id' => 1,
+			'is-not-empty' => true
+		];
+
+		$dependencies = tribe_format_field_dependency( $deps );
+
+		// Note the leading space - this is intentional, so we're not trim()-ing it to ensure it gets added!
+		$this->assertEquals( ' data-depends=1 data-is-not-empty', $dependencies );
```
sc0ttkclark, line 33: Tests probably need to be adjusted

sc0ttkclark (changes requested):
Needs testing adjustments

sc0ttkclark (approved):

lucatume (approved):

PR #1503, opened 2020-11-21
Add a submodule-sync workflow

borkweb:


.github/workflows/submodule-sync.yml
```diff
@@ -0,0 +1,70 @@
...
+        working-directory: ./tut
+        with:
+          envkey_GITHUB_USER: tr1b0t
+          envkey_GITHUB_OAUTH_TOKEN: ${{ secrets.GH_BOT_TOKEN }}
+          file_name: .env
+
+      - name: Synchronize submodules
+        working-directory: ./tut
+        run: |
+          ./tut submodule-sync
```
sc0ttkclark, line 69: I think this should only happen for the current branch. We should update tut to restrict to the branch it's currently on from the common workflow

sc0ttkclark (commented):
Looks great! One change might be even better/safe

.github/workflows/submodule-sync.yml
borkweb: Good call! The command doesn't support that yet. BUT...this action now works, so I'll go off and update the command to allow us to specify a specific branch :man_dancing:

sc0ttkclark: Heck yeah!!

borkweb:
Now that this is working swimmingly with a specific branch, I'll close this out and open a real PR

PR #1506, opened 2020-11-26
Add action before inline style print

lucatume:
This PR adds an action in the Customizer component before inline
styles are printed.

The action is leveraged in the other plugin tests to make the Customizer
output easier to test.

### DIFF
--- a/src/Tribe/Customizer.php
+++ b/src/Tribe/Customizer.php
@@ -396,11 +396,13 @@ final class Tribe__Customizer {
 	/**
 	 * Print the CSS for the customizer using wp_add_inline_style
 	 *
-	 * @return void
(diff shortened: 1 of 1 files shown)


Camwyn (approved):
This looks good. Is it worthwhile to add a sheet-specific hook as well?
`do_action( "tribe_customizer_before_{$sheet}_inline_style", $sheet, $inline_style );`?

I'm guessing not, but the thought crossed my mind.

lucatume:
> This looks good. Is it worthwhile to add a sheet-specific hook as well?
> `do_action( "tribe_customizer_before_{$sheet}_inline_style", $sheet, $inline_style );`?
> 
> I'm guessing not, but the thought crossed my mind.

I would say no.

PR #1522, opened 2020-12-14
Move TribeDb to common.

juanfra:
🎥 https://www.loom.com/share/4a182954b6a54eebaf2dde1f8c6364a7

Working back on https://github.com/moderntribe/event-tickets/pull/1772. Scott suggested to move TribeDb to common, so I'm opening this PR.


lucatume (commented):
The idea is a good one, the only question I have is if the tests that are actually using this, in TEC and PRO, are still working?

juanfra:
@lucatume I think so, given that they're still using the `TribeDb` from those repos. It'd be ideal to make them use this version from common (which was grabbed from what we already have in TEC). 

The only thing to change there would be the inclusion, something like [this](https://github.com/the-events-calendar/event-tickets/pull/1772/files#diff-f43957ff1990ab850481d9be76abbd0f6c804db95704fec5c901bea250e52eb6) and [this](https://github.com/the-events-calendar/event-tickets/pull/1772/files#diff-24754dc4db93d15e983f9f4cbab56349d8d80847f76431287938d2135a288449R5)

sc0ttkclark:
Let's rename the class now while we can. In that way, there's no conflicts with existing classes loading and it'll be easy to tell when things are changed over through subsequent PRs across the various repositories.

sc0ttkclark (approved):

juanfra:
@sc0ttkclark should we also update the namespace then?

sc0ttkclark:
Namespace is definitely something for a bigger PR effort (I think)

PR #1526, opened 2020-12-30
Tweaks for the countdown widget.

Camwyn:
Adds a "search events" feature to the dropdown object.
Tweaks `dependency.js` (gently!)
Corrects a few spelling issues.

Part of [ECP-542] & [ECP-546]
🎥 http://p.tri.be/xC4s7Z

See linked PRO PR for more screenshots.
TEC: https://github.com/the-events-calendar/the-events-calendar/pull/3415
PRO: https://github.com/the-events-calendar/events-pro/pull/1656

[ECP-542]: https://theeventscalendar.atlassian.net/browse/ECP-542

[ECP-546]: https://theeventscalendar.atlassian.net/browse/ECP-546


src/resources/js/dependency.js
```diff
@@ -306,7 +305,7 @@
 	}, obj.selectors.dependency );
 
 	// Configure on Document ready for the default trigger
-	$document.ready( obj.setup );
```
bordoni, line 309: Why the modification here? It seems to be the same no?

bordoni (changes requested):

lucatume:
@Camwyn I get the following failure in the `wpunit` suite (fast mode):
Screenshot 2021-01-04 at 10 41 35

Check, please.

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
...
 
 	/**
-	 * Sorts Hierarchically all the Terms for Select2
+	 * Search for Posts using Select2
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string|array        $search   Search string from Select2.
+	 * @param  int                 $page     When we deal with pagination.
+	 * @param  array,string,mixed> $args     Arguments to pass to the query.
```
lucatume, line 104:
```suggestion
	 * @param  array<string,mixed> $args     Arguments to pass to the query.
```

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
 	}
 
 	/**
-	 * Sorts Hierarchically all the Terms for Select2
+	 * Search for Posts using Select2
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string|array        $search   Search string from Select2.
```
lucatume, line 102: Add types to the array entry.

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
...
+	 * @param  int                 $page     When we deal with pagination.
+	 * @param  array,string,mixed> $args     Arguments to pass to the query.
+	 * @param  string|int          $selected Selected item ID.
+	 *
+	 * @return array
```
lucatume, line 107: Add types to array.

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
...
+				// Newer SelectWoo uses a new search format.
+				$args['s'] = $search['term']; // post?
+			} else {
+				// For older pieces that still use Select2 format.
+				$args['s'] = $search;
+			}
+		}
+
+		$args['posts_per_page'] = $posts_per_page;
+		$args['paged']          = $page;
```
lucatume, line 124: Since that will not be used at all, I would skip the caching of meta and taxonomy terms and make things a bit faster:

```suggestion
		$args['paged']          = $page;
		$args['update_post_meta_cache'] = false;
		$args['update_post_term_cache'] = false;
```

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
...
+
+		$results = new WP_Query( $args );
+
+		if ( ! $results->have_posts() ) {
+			return $data;
+		}
+
+		foreach( $results->posts as $post ) {
+			// Prep for Select2
+			$post->id        = $post->ID;
```
lucatume, line 134: What fields of the post object are **actually** being used in the JS?
If those among to the the post ID and title, then the full post object fetch could be skipped to use a direct query to fetch only the post ID and title.

Not required for this PR, but it's a 1% gain it's worth exploring.

lucatume: To be clear, this would mean running a query to fetch only the post IDs (fast) and then a second one to only fetch the post titles on the matches.

src/Tribe/Ajax/Dropdown.php
```diff
@@ -128,7 +177,7 @@ public function sort_terms_hierarchically( &$terms, &$into, $parent = 0 ) {
 	 *
 	 * @since  4.6
 	 *
-	 * @param  array  $results  The Select2
+	 * @param  array|object  $results  The Select2
```
lucatume, line 180: Add array type.

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,13 +95,62 @@ public function search_terms( $search, $page, $args, $source ) {
...
+
+	/**
+	 * Sorts all the Terms for Select2 hierarchically.
 	 *
 	 * @since  4.6
 	 *
-	 * @param  array   &$terms Array of Terms from `get_terms`
-	 * @param  array   &$into  Variable where we will store the
-	 * @param  integer $parent Used for the recursion
+	 * @param  array   &$terms Array of Terms from `get_terms`.
```
lucatume, line 151: Here and elsewhere: add array type.

src/functions/template-tags/html.php
```diff
@@ -178,31 +179,36 @@ function tribe_format_field_dependency( $deps ) {
 
 		// Handle the ID component.
 		if ( 'id' === $attr ) {
-			$dependency .= " data-depends=\"{$value}\"";
+			$dependency .= " data-depends=\"#{$value}\"";
```
lucatume, line 182: Tests need to be updated as this is breaking.

lucatume: This change makes sense in logic, but I'm not sure all of our code would correctly use it and it could result, in some instances, in an attribute value like `##some-id`.
Maybe detect the starting `#` here and add it if required?

src/functions/template-tags/html.php
```diff
@@ -178,31 +179,36 @@ function tribe_format_field_dependency( $deps ) {
...
+		if ( 'parent' === $attr ) {
+			$dependency .= " data-dependent-parent=\"{$value}\"";
 			continue;
 		}
 
 		// Handle boolean values.
 		if ( is_bool( $value ) ) {
 			if ( $value ) {
-				$dependency .= " data-{$attr}";
+				$dependency .= " data-condition-{$attr}";
```
lucatume, line 195: Could you explain this? It looks to me like the logic is changing here.

lucatume (changes requested):

src/functions/template-tags/html.php
paulmskim: I'm a bit worried if we have any code currently using this (and other cases of where we're modifying existing code). If so, this could break the current implementation of those instances.

src/resources/js/dependency.js
```diff
@@ -108,11 +108,10 @@
...
 
 		$dependents.each( function( k, dependent ) {
-			var $dependent         = $( dependent );
-			var hasDependentParent = $dependent.is( '[data-dependent-parent]' );
+			var $dependent = $( dependent );
 
-			if ( hasDependentParent ) {
-				var dependentParent  = $dependent.data( 'dependentParent' );
+			if ( $dependent.is( '[data-dependent-parent]' ) ) {
+				var dependentParent  = $dependent.data( 'dependent-parent' );
```
paulmskim, line 114: Do we have any code that depends on this (that uses `data-dependentParent`)? This will break any implementation of that if so.

Camwyn: We do not that I could find. Although @bordoni might know of an instance?

src/Tribe/Ajax/Dropdown.php
Camwyn: Yes - I was duplicating some existing code but I like the simpler approach. Let me check that we're not using the rest of the post and reply.

src/resources/js/dependency.js
Camwyn: Because most common/existing ways of using `$.ready` are now deprecated - it's back-compatible but forward-looking.
See: https://api.jquery.com/ready/#ready-handler Starting with "jQuery offers several ways to attach a function that will run when the DOM is ready."

src/resources/js/dependency.js
Camwyn: http://p.tri.be/MLPrhj no `data-dependent-parent (except here and in the js)
http://p.tri.be/WV1eor no `dependentParent` (except in the js)

src/functions/template-tags/html.php
Camwyn: This code did not work correctly as it was. It checks for `data-condition-foo` but we were setting `data-foo`

src/resources/js/dependency.js
bordoni: Hun crazy that they deprecated the `.ready()` method.

We will need to do a major change in our codebase.

src/functions/template-tags/html.php
Camwyn: Changed

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,15 +95,79 @@ public function search_terms( $search, $page, $args, $source ) {
...
+			return false;
+		}
+
+
+		$events = wp_list_pluck( $results->posts, 'ID' ) ;
+		if ( ! empty( $selected ) && ! in_array( $selected, $events) ) {
+			$events[] = $selected;
+		}
+
+		$events_list = implode( ',', $events );
```
bordoni, line 140: Can we also pass this to a sanitization, just in case? like `array_filter()`

bordoni (changes requested):

src/functions/template-tags/html.php
Camwyn: History: this was created and then _not used_ for the list widget.

src/Tribe/Ajax/Dropdown.php
Camwyn: Done

src/Tribe/Ajax/Dropdown.php
Camwyn: Done

src/Tribe/Ajax/Dropdown.php
Camwyn: Changed

src/Tribe/Ajax/Dropdown.php
Camwyn: Done

src/Tribe/Ajax/Dropdown.php
Camwyn: Done

src/resources/js/dependency.js
Camwyn: At some point, yeah. This is a a small one (I think) for now?

src/Tribe/Ajax/Dropdown.php
Camwyn: Done

src/Tribe/Ajax/Dropdown.php
```diff
@@ -95,15 +95,76 @@ public function search_terms( $search, $page, $args, $source ) {
...
+	 * Search for Posts using Select2
+	 *
+	 * @since  TBD
+	 *
+	 * @param  string|array        $search   Search string from Select2.
+	 * @param  int                 $page     Page we want when we're dealing with pagination.
+	 * @param  array<string,mixed> $args     Arguments to pass to the query.
+	 * @param  string|int          $selected Selected item ID.
+	 *
+	 * @return array<string|mixed>
```
sc0ttkclark, line 107: This could return false

src/Tribe/Ajax/Dropdown.php
```diff
@@ -227,7 +289,8 @@ private function success( $data ) {
 	 *
 	 * @since  4.6
 	 *
-	 * @param  array $data
+	 * @param array<string|mixed> $data
```
sc0ttkclark, line 292: This doesn't match method signature

src/functions/template-tags/html.php
sc0ttkclark: I want to say we did something with this for ET tickets in the classic editor, let me go check and see if we had built something extra based on that.

sc0ttkclark: Looks clear, just wanted to check!

sc0ttkclark (changes requested):
Looks good, a few minor things left I think

src/resources/js/dependency.js
paulmskim: OK. My other worry is if we have cases in the code where we construct these data attributes using variables, like `"data-$key"` or, less likely, `"data-dependent$key"`. But if we're fairly certain this was used for widgets, I'd say let's be cautious and keep an eye out for anything that looks amiss.

src/functions/template-tags/html.php
lucatume: @paulmskim please, resolve this conversation when you're satisfied.

lucatume (changes requested):
Adress the changes required by other reviewers (Common gets a lot of attention!).

src/Tribe/Ajax/Dropdown.php
Camwyn: Corrected - the return

src/Tribe/Ajax/Dropdown.php
Camwyn: corrected

sc0ttkclark (approved):

lucatume (approved):

bordoni (approved):

PR #1543, opened 2021-02-08
Simplify Widgets code and Move stuff into Common

bordoni:
🎫 [ECP-486]
---
📹 https://i.bordoni.me/Z4u6L7kY

[ECP-486]: https://theeventscalendar.atlassian.net/browse/ECP-486

### DIFF
--- a/src/admin-views/widgets/components/fields/section.php
+++ b/src/admin-views/widgets/components/fields/section.php
@@ -33,7 +33,7 @@ $section_classes = array_merge( [ 'tribe-widget-form-control', 'tribe-widget-for
 	?>
 >
 	<?php if ( ! empty( $label ) ) : ?>
-		<?php // Note: the actual widget title/handle is an <h3>. ?>
+		<?php // Note: the actual widget title/handle is an <h2>. ?>
 		<h4 class="tribe-widget-form-control__section-title"><?php echo esc_html( $label ); ?></h4>
 	<?php endif; ?>
 
--- a/src/admin-views/widgets/components/fields/section.php
+++ b/src/admin-views/widgets/components/fields/section.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Admin View: Widget Component Section field.
+ *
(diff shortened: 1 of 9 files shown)


src/admin-views/widgets/components/fields/section.php
```diff
@@ -0,0 +1,41 @@
...
+?>
+<div
+	<?php tribe_classes( $section_classes ); ?>
+	<?php
+	// Not escaped - contains html (data-attr="value").
+	echo $dependency; // phpcs:ignore
+	?>
+>
+	<?php if ( ! empty( $label ) ) : ?>
+		<?php // Note: the actual widget title/handle is an <h3>. ?>
```
Camwyn, line 36: No - it's an h2.
So maybe the below one should be an h3?

Camwyn (changes requested):

src/admin-views/widgets/components/fields/section.php
Camwyn: This is changing - but better to fix it here first?

src/admin-views/widgets/components/fields/section.php
```diff
@@ -0,0 +1,41 @@
...
+<div
+	<?php tribe_classes( $section_classes ); ?>
+	<?php
+	// Not escaped - contains html (data-attr="value").
+	echo $dependency; // phpcs:ignore
+	?>
+>
+	<?php if ( ! empty( $label ) ) : ?>
+		<?php // Note: the actual widget title/handle is an <h2>. ?>
+		<h4 class="tribe-widget-form-control__section-title"><?php echo esc_html( $label ); ?></h4>
```
Camwyn, line 37: as part of https://github.com/the-events-calendar/tribe-common/pull/1543#pullrequestreview-587870355
should this be an <h3>?

bordoni: No not in this case, this is the admin.

PR #1546, opened 2021-02-09
Update mascot to the manatee

borkweb:
Chubby water creature!


sc0ttkclark (approved):
Would look better with a hat but this is a good start

PR #1549, opened 2021-02-21
Assets allow Sync and Defer

bordoni:
Our Views V2 manager JS was causing a lot of priority problems due to it's requirement of begin loaded last, after a little bit of playing around I realized the best way to actually solve the problem was to use `defer` for the `<script>` tag.

WordPress makes it quite hard to `defer` or `async` a script, so we filter the HTML tag itself.


src/Tribe/Assets.php
```diff
@@ -38,6 +38,45 @@ public static function instance() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $tag    Tag we are filtering.
+	 * @param string $handle Which is the ID/Handle of the tag we are about to print.
+	 *
+	 * @return string Script tag with the defer and/or async attached.
+	 */
+	public function filter_tag_async_defer( $tag, $handle ) {
+		// Only filter for own own filters.
```
Camwyn, line 55: Can we put a sanity check in here to be sure it's a script tag - it's a public function, so someone may call it incorrectly (and a script and style asset _might_ be named the same)?

src/Tribe/Assets.php
```diff
@@ -38,6 +38,45 @@ public static function instance() {
...
+	 * Filters the Script tags to attach Async and/or Defer based on the rules we set in our Asset class.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $tag    Tag we are filtering.
+	 * @param string $handle Which is the ID/Handle of the tag we are about to print.
+	 *
+	 * @return string Script tag with the defer and/or async attached.
+	 */
+	public function filter_tag_async_defer( $tag, $handle ) {
```
Camwyn, line 54: Can we put a sanity check in here to be sure it's a script tag - it's a public function, so someone may call it incorrectly (and a script and style asset might be named the same)?

Camwyn (changes requested):

Camwyn (approved):

Camwyn (approved):

PR #1551, opened 2021-02-25
Context method for dangerously_repopulate_locations

bordoni:
Simple in concept but very dangerous, so I wanted to make sure it was very clear.

Used in the Pro Shortcodes:
https://github.com/the-events-calendar/events-pro/pull/1705/files#diff-dcb1c960a6bcd15eb9fb3e0da181cbab759450cf6295bed6b71aab497161b89aR306-R312


sc0ttkclark:
I'd really rather give a new method to add a new location to the list of locations instead here. This feels wrong in the implementation from ECP.

bordoni:
I don't think I follow what was your question there @sc0ttkclark. The goal of this method was to give the possibility of regenerating the locations which by default will not update once you started to mess with the context.

Ideally the goal is once you have plugins all loaded Locations are set in place and wont change.

But there are certain specific scenarios where the implementation requires us to be able to ignore global context.

lucatume (dismissed):
My gut feeling would be "NOOOOO", but the wealth of disclaimers make it something I could accept.

sc0ttkclark (dismissed):
Like Luca, I'm sad but OK

PR #1553, opened 2021-02-26
[ET-1023] Fix Tribe__Cache non_persistent_keys checks so they always reference the correct cache key

sc0ttkclark:
[ET-1023]

This is a simple PR in it's intent: Ensure non-persistent keys get set to the non-md5'd key properly in the `set()` method so subsequent calls do not lead to the incorrect **persistent** cache group used.

The `isset()` logic for non-persistent keys has also been updated to reduce overhead and complexity.

This PR includes a new test that confirms this fix and provides checks for various values/types to confirm they work as *currently* expected.

[ET-1023]: https://theeventscalendar.atlassian.net/browse/ET-1023

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [4.12.19] TBD =
 
-* Fix - Prevent problems when using longer array keys in `Tribe__Cache` so the correct non-persistent groups are referenced.
+* Fix - Prevent problems when using longer array keys in `Tribe__Cache` so the correct non-persistent groups are referenced. [ET-1023]
 
 = [4.12.18] 2021-02-24 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.12.19] TBD =
+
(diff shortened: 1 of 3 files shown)


sc0ttkclark:
Working on getting a ticket number for this, will get it and add it to the changelog once I have that.

sc0ttkclark:
The work here was directly related to findings from testing https://github.com/the-events-calendar/event-tickets/pull/1954

Specifically changes in `Ticket_Object.php` here: https://github.com/the-events-calendar/event-tickets/pull/1954/files#diff-7ebe2c382547021fbff65788caa8364a8e4be5ab4d6c2970bba860dc52ef28f3

The isset() would pass but the value was false so it would pass along an unexpected value (it was not intentionally supporting false).

readme.txt
```diff
@@ -2,6 +2,10 @@
 
 == Changelog ==
 
+= [4.12.19] TBD =
+
+* Fix - Prevent problems when using longer array keys in `Tribe__Cache` so the correct non-persistent groups are referenced.
```
```suggestion
* Fix - Prevent problems when using longer array keys in `Tribe__Cache` so the correct non-persistent groups are referenced. [ET-1023]
```

borkweb (approved):
:noice:

PR #1562, opened 2021-03-13
Move Taxonomy file from TEC to common, as it will aid all plugins on shortcode

bordoni:
🎫 [ECP-728]
---
All shortcodes should now be using the new system in V2, allows much more flexibility when passing params and validating them. 

[ECP-728]: https://theeventscalendar.atlassian.net/browse/ECP-728

### DIFF
--- a/src/Tribe/Utils/Taxonomy.php
+++ b/src/Tribe/Utils/Taxonomy.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace Tribe\Utils;
+
+/**
+ * Class Taxonomy.
+ *
+ * @since   TBD
+ *
+ * @package Tribe\Utils
+ */
+class Taxonomy {
+	/**
+	 * Match any operand.
+	 *
+	 * @since TBD
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const OPERAND_OR = 'OR';
+
+	/**
+	 * Match all operand.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const OPERAND_AND = 'AND';
+
+	/**
+	 * Default operand for taxonomy filters.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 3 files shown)


src/Tribe/Utils/Taxonomy.php
```diff
@@ -0,0 +1,112 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string       $taxonomy Which taxonomy we are using to setup.
+	 * @param string|array $terms    Which terms we are going to use here.
+	 * @param string       $operand  Which is the operand we should use.
+	 *
+	 * @return array A fully qualified `tax_query` array, merge using `array_merge_recursively`.
+	 */
+	public static function translate_to_repo( $taxonomy, $terms, $operand = self::OPERAND_OR ) {
```
lucatume, line 52: I would change the method name to either `translate_to_orm_args` or `to_orm_args` or `to_repository_args`.

src/Tribe/Utils/Taxonomy.php
```diff
@@ -0,0 +1,112 @@
...
+	 * Transform all Term IDs and Slugs into IDs of existing terms in a given taxonomy.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|int|array $terms Terms to be cleaned up.
+	 * @param string           $taxonomy Which taxonomy we are querying for.
+	 *
+	 * @return array List of IDs of terms.
+	 */
+	public static function sanitize_shortcode_terms_to_id( $terms, $taxonomy ) {
```
lucatume, line 90: This looks more like a normalization (all eligible will be IDs at the end); should this method be called `normalize_term_ids`?  Furthermore: is the notion of a shortcode required at all? It seems like a normalization function that would serve in a number of scenarios, not only shortcode, correct?

lucatume (commented):
I would change some method names to better represent or express what they do.

lucatume (approved):

PR #1578, opened 2021-04-27
Add changelog and Cache listener hooks

bordoni:
🎫 [ECP-826] [ECP-824]
---
Expiring the cache for Widgets when saving their settings, no demo here, since the problem was not happening locally.

[ECP-826]: https://theeventscalendar.atlassian.net/browse/ECP-826
[ECP-824]: https://theeventscalendar.atlassian.net/browse/ECP-824

### DIFF
--- a/src/Tribe/Cache_Listener.php
+++ b/src/Tribe/Cache_Listener.php
@@ -1,171 +1,214 @@
 <?php
+
+/**
+ * Listen for events and update their timestamps
+ */
+class Tribe__Cache_Listener {
+
 	/**
-	 * Listen for events and update their timestamps
+	 * The name of the trigger that will be fired when rewrite rules are generated.
 	 */
-	class Tribe__Cache_Listener {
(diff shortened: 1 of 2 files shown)


src/Tribe/Cache_Listener.php
```diff
@@ -1,171 +1,214 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string,bool> $triggers Which options will trigger this given action last occurrence.
+	 * @param string             $action   Which action this trigger will set.
+	 * @param array              $args     Which arguments from the updated option method.
+	 *
+	 * @return array
+	 */
+	public function filter_action_last_occurrence_triggers( $triggers, $action, array $args = [] ) {
```
mitogh, line 145: Might be a good idea to enforce and initialize some args. 

```suggestion
	public function filter_action_last_occurrence_triggers( array $triggers = [], $action = '', array $args = [] ) {
```

mitogh (approved):

PR #1579, opened 2021-04-27
Help  PHPStorm to resolve types from the container

mitogh:
Allow PHPStorm Editor to resolve the different types when using the
container to get an instance to an object, this works only when the
container is called with a string that is either the full quailified
class of the object or when the name is used on the container such as:

```php
tribe( Object_Type::class )->resolves_names_of_method_or_properties();
tribe( 'Object_Type' )->resolves_names_of_method_or_properties();
```

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace PHPSTORM_META {
-	// Allow PHP Storm Editor to resolve return types when calling tribe( Object_Type::class ) or tribe( `Object_type` )
+	// Allow PhpStorm IDE to resolve return types when calling tribe( Object_Type::class ) or tribe( `Object_Type` )
 	override(
 		\tribe( 0 ),
 		map( [
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace PHPSTORM_META {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Wow, wicked!

sc0ttkclark (approved):

.phpstorm.meta.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace PHPSTORM_META {
+	// Allow PHP Storm Editor to resolve return types when calling tribe( Object_Type::class ) or tribe( `Object_type` )
```
sc0ttkclark, line 4:
```suggestion
	// Allow PhpStorm IDE to resolve return types when calling tribe( Object_Type::class ) or tribe( `Object_type` )
```

sc0ttkclark:
This needs to ultimately be added to all plugin repos, correct?

mitogh:
> This needs to ultimately be added to all plugin repos, correct?

Not really as long as the plugin uses common the `.phpstorm.meta.php` file is recursively added  so if you have a specif configuration on other plugins those are merged into a single one.

sc0ttkclark:
Oh wow, interesting! Much wow, much autocomplete!

mitogh:
Oh for sure it was a pain to describe or create a variable just to get the autocomplete. 

Reference: https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html

bordoni (dismissed):

PR #1583, opened 2021-04-30
[Innovation] Separate theme overrides into own files

paulmskim:
**Why:** We serve up theme overrides with our skeleton and full styles, no matter what theme the user is using. This doesn't make sense. It increases the amount of CSS and data we pass from the server to the visitor.

**What:** This is the first pass at separating out theme overrides into their own styles. Based on the theme and display settings, we will conditionally serve up the proper theme overrides.


borkweb:

PR #1584, opened 2021-05-04
New common-centric Theme Compatibility

Camwyn:
Move from Body_Classes (not removed for back compat) to Compatibility_Classes.

While the object is to move away from just putting classes on the body, this in no way prevents that - rather it generalizes more that we're dealing with a queue of classes and we need to know which ones are appropriate.

Create common Theme Compatibility for TEC & ET to extend. Based off TEC.
Move all non-plugin specific functionality there.

Add some tests.
Passing tests are the artifact.

Required by: https://github.com/the-events-calendar/the-events-calendar/pull/3513


bordoni (approved):

src/Tribe/Utils/Theme_Compatibility.php
```diff
@@ -0,0 +1,266 @@
...
+	];
+
+	/**
+	 * Checks if the current theme needs a compatibility fix.
+	 *
+	 * @since  TBD
+	 *
+	 * @param string $theme Optionally, pass a specific theme name in to see if compatibility
+	 *                      is required for that theme.
+     *
```
paulmskim, line 40:
```suggestion
	 *
```

src/Tribe/Utils/Theme_Compatibility.php
```diff
@@ -0,0 +1,266 @@
...
+
+	/**
+	 * Contains the logic for if this object's classes should be added to the queue.
+	 *
+	 * @since TBD
+	 *
+	 * @param boolean $add   Whether to add the class to the queue or not.
+	 * @param array   $class The array of compatibility class names to add.
+	 * @param string  $queue The queue we want to get 'admin', 'display', 'all'.
+
```
paulmskim, line 68:
```suggestion
	 *
```

src/Tribe/Utils/Compatibility_Classes.php
```diff
@@ -0,0 +1,352 @@
...
+		$existing_classes = explode( ' ', $classes );
+		// Make sure they should be added.
+		if ( ! $this->should_add_compatibility_classes( $this->get_class_names( 'admin' ), (array) $existing_classes, 'admin' ) ) {
+			// Ensure we return the current string on false!
+			return $classes;
+		}
+
+		$element_classes = new Element_Classes( $this->get_class_names( 'admin' ) );
+
+		return implode( ' ', array_merge( $existing_classes, $element_classes->get_classes() ) );
```
bordoni, line 301:
```suggestion
		$element_classes = new Element_Classes( array_merge( $existing_classes, $this->get_class_names( 'admin' ) ) );

		return $element_classes->get_classes_as_string();
```

src/Tribe/Utils/Theme_Compatibility.php
```diff
@@ -0,0 +1,266 @@
...
+		if ( empty( $current_theme ) || empty( $current_theme->get_template() ) ) {
+			return $classes;
+		}
+
+		// Detect if we're using a child theme.
+		if ( $parent = $current_theme->parent ) {
+			$classes[] = sanitize_html_class( 'tribe-theme-' . $parent->get_template() );
+			$classes[] = sanitize_html_class( 'tribe-theme-child-' . $current_theme->get_template() );
+		} else {
+			$classes[] = sanitize_html_class( 'tribe-theme-' . $current_theme->get_template() );
```
bordoni, line 150: Will these be passed down to the `Element_Classes`?

bordoni (changes requested):

src/Tribe/Utils/Theme_Compatibility.php
Camwyn: They aren't right now.
Compatibility classes go on compatibility containers _only_ so it would be bad form (and a lot of noise) to add them to every element that we output.
From the other side, the container needs no additional classes, so I didn't see a need to add E_C at first, although I can see how the additional functionality might be useful?

bordoni: No I think we are good like that. Just double checking.

Camwyn: Adding a second thought: if, for some reason we _want_ to add compatibility classes (like body classes in the interim for backwards compatibility) we should do it manually at the point we'd call Element_Classes - easier to clean up?

bordoni (approved):

PR #1585, opened 2021-05-04
ECP-737: ECP > Conflict with Avada

mitogh:
Prevent to deal with conflicts when `selectWoo` is rendered and used
along side another plugins that uses `selectWoo` with
different versions.

Instead all the `selectWoo` references are replaced with the
existing `select2TEC`

Artifact: https://d.pr/v/nfjMNu
Ticket: [ECP-737]

[ECP-737]: https://theeventscalendar.atlassian.net/browse/ECP-737

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [4.13.3] TBD =
 
-
+* Fix - Compatibility with Avada themes and third party plugins or themes loading `selectWoo` at the same time. [ECP-737]
 
 = [4.13.2] 2021-04-29 =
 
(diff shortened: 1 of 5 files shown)


lucatume (approved):

sc0ttkclark (commented):
We should double check this doesn't have ramifications for any of the other plugins, in case they are looking for selectWoo themselves

mitogh:
Thanks, for the heads up we are increasing the QA level on this one due this reason, however it seems like all tribe dropdowns uses a wrapper to create new selects like https://github.com/the-events-calendar/tribe-common/blob/master/src/resources/js/dropdowns.js#L454 to create a new instance.

PR #1586, opened 2021-05-06
TEC-3796: Adjust the actions used to register styles for the common

mitogh:
The changes relies on the fact that only register the scripts, once
the script is required is loaded by the callers of the `group`
of assets.

Screencast: https://d.pr/v/dUJBur
Ticket: TEC-3796

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 
 = [4.13.3] TBD =
 
-
+* Tweak - Adjust the actions used to register and load the styles for the tooltip component [TEC-3796]
 
 = [4.13.2] 2021-04-29 =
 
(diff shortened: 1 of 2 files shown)


Camwyn (approved):
I'm approving, but I wonder if a hook needs to be defined at all? Won't the group enqqueue take care of that?

mitogh:
Great question, so basically what this change does is only register the scripts but does not include it on the assets the assets are loaded once this function is called:

https://github.com/the-events-calendar/tribe-common/blob/master/src/Tribe/Tooltip/View.php#L61

Camwyn (approved):

mitogh:
FYI: This PR does not have any relationship with the ticket TEC-3849  it was a mistake on the branching name.

PR #1591, opened 2021-05-11
Add support for chunkable transients

lucatume:
Issue: https://theeventscalendar.atlassian.net/browse/TEC-3627

This commit adds the code, and tests, to support "chunkable" transient.

A "chunkable" transient is one where we know that the cost of building
the transient value will, 80% of the times, be more than the one
required to fetch each transient row from the db.

The function will be completely skipped when object caching is active
and developers will have to make a choice to use the API with knowledge
of the above.

> Note: I've added tests using the `uopz` extension that will be skipped in CI due to the fact that it's not, currently, using `tric`.

Artifact: 
Screenshot 2021-05-11 at 17 29 29


readme.txt
```diff
@@ -1,12 +1,14 @@
...
 
 == Changelog ==
 
 = [4.13.3] TBD =
 
 * Feature - Create new functionality in Tribe__Customizer__Section to allow for simpler craetion of controls and sections. [TEC-3836]
+* Feature - Added the `set_chunkable_transient` and `get_chunkable_transient` functions to the Cache class, see doc-blocks. [TEC-3627]
 * Fix - Compatibility with Avada themes and third party plugins or themes loading `selectWoo` at the same time. [ECP-737]
 * Tweak - Adjust the actions used to register and load the styles for the tooltip component [TEC-3796]
+* Tweak - Adjust the actions used to register and load the styles for the tooltip component [TEC-3796]
```
borkweb, line 11: Accidental duplicate line.

```suggestion
```

borkweb (approved):
Clever! Other than the duplicated line in the changelog, this looks good. :kangaroo: :dash:

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+	public function set_chunkable_transient( $id, $value, $expiration = 0, $expiration_trigger = '' ) {
+		$transient = $this->get_id( $id, $expiration_trigger );
+
+		if ( wp_using_ext_object_cache() ) {
+			return $this->set_transient( $transient, $value, $expiration );
+		}
+
+		$inserted         = [];
+		$serialized_value = maybe_serialize( $value );
+		$chunk_size       = tribe( 'feature-detection' )->get_mysql_max_packet_size() * .9;
```
mitogh, line 608: Just to make it more clear. 
```suggestion
		$chunk_size       = tribe( 'feature-detection' )->get_mysql_max_packet_size() * 0.9;
```

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+
+		try {
+			$data          = implode( '', $chunks );
+			$is_serialized = preg_match( '/^[aO]:\\d+:/', $data );
+			$unserialized  = maybe_unserialize( implode( '', $chunks ) );
+			if ( is_string( $unserialized ) && $unserialized === $data && $is_serialized ) {
+				// Something was messed up.
+				return false;
+			}
+		} catch ( Exception $e ) {
```
mitogh, line 578: Maybe log the exception ?

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+		}
+
+		try {
+			$data          = implode( '', $chunks );
+			$is_serialized = preg_match( '/^[aO]:\\d+:/', $data );
+			$unserialized  = maybe_unserialize( implode( '', $chunks ) );
+			if ( is_string( $unserialized ) && $unserialized === $data && $is_serialized ) {
+				// Something was messed up.
+				return false;
+			}
```
mitogh, line 577:
```suggestion
				return true;
			}
```

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+
+		$chunks = [];
+		$i      = 0;
+		do {
+			$chunk_transient            = $transient . '_' . $i ++;
+			$chunk                      = get_transient( $chunk_transient );
+			$chunks[ $chunk_transient ] = (string) $chunk;
+		} while ( ! empty( $chunk ) );
+
+		if ( count( array_filter( $chunks ) ) === 0 ) {
```
mitogh, line 566:
```suggestion
		if ( empty( array_filter( $chunks ) ) ) {
```

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+		$transient = $this->get_id( $id, $expiration_trigger );
+
+		if ( wp_using_ext_object_cache() ) {
+			return get_transient( $transient );
+		}
+
+		$chunks = [];
+		$i      = 0;
+		do {
+			$chunk_transient            = $transient . '_' . $i ++;
```
mitogh, line 561:
```suggestion
			$chunk_transient            = $transient . '_' . $i++;
```

src/Tribe/Cache.php
```diff
@@ -536,4 +536,93 @@ public function data_size_over_packet_size( $value ) {
...
+	 */
+	public function get_chunkable_transient( $id, $expiration_trigger = '' ) {
+		$transient = $this->get_id( $id, $expiration_trigger );
+
+		if ( wp_using_ext_object_cache() ) {
+			return get_transient( $transient );
+		}
+
+		$chunks = [];
+		$i      = 0;
```
mitogh, line 559:
```suggestion
		$i      = 1;
```

As this is just an indicator for the transient number I think `1` makes more sense as start point

mitogh (changes requested):

src/Tribe/Cache.php
lucatume: I would say no; w/ the frequency we use transients, that would produce noise w/o some real value.

src/Tribe/Cache.php
```diff
@@ -536,4 +536,94 @@ public function data_size_over_packet_size( $value ) {
...
+
+		try {
+			$data          = implode( '', $chunks );
+			$is_serialized = preg_match( '/^[aO]:\\d+:/', $data );
+			$unserialized  = maybe_unserialize( implode( '', $chunks ) );
+			if ( is_string( $unserialized ) && $unserialized === $data && $is_serialized ) {
+				// Something was messed up.
+				return false;
+				return true;
+			}
```
mitogh, line 578: Sorry looks like this one was placed on the wrong place.

```suggestion
			if ( is_string( $unserialized ) && $unserialized === $data && $is_serialized ) {
				// Something was messed up.
				return false;
			}
			
			return true;
```

mitogh (changes requested):

mitogh (approved):

PR #1592, opened 2021-05-12
TCMN-107: Display `draft` posts as part of the events

mitogh:
The events are displayed only if the user is able to `read_private_posts`.

Ticket: TCMN-107

This would display all draft events to the current user only if the
user is able to `read_private_posts`.

Artifact: https://d.pr/v/zO4CH9


paulmskim (approved):

paulmskim (commented):
Curious here, because I'm not sure how it works, does this affect anything around password protected events? Or is that a separate system?

mitogh:
Great question this does not affect this behavior as a password protected password can be set as `publish` by default or it can be set to `draft` in which case (if is a `draft`) only a user that can `read_private_posts` should be able to see the title or the event on the list (which is the current behavior for `publish` status) but once accessing the password the password still need to be entered.

PR #1594, opened 2021-05-14
ECP-737: Minify changes on the vendor for selectWoo

mitogh:
Minifiy the production version for selectWoo file.

Ticket: ECP-737


vendor/tribe-selectWoo/dist/js/select2.full.js
```diff
@@ -6559,7 +6559,7 @@ S2.define('jquery.select2',[
   }
 
   // Also register select2TEC under select2 if select2 is not already present.
-  $.fn.select2 = $.fn.select2 || $.fn.select2TEC;
+  $.fn.select2TEC = $.fn.select2 || $.fn.select2TEC;
```
borkweb, line 6562: This change alters the purpose of this line, I think.  The goal was to assign select2TEC into select2 if select2 isn't defined.

mitogh: Actually the goal is to create a wrapper around `select2` meaning not to define `select2`  as if this file is executed before any other select2 from  themes and plugins this version of `select2` is going to be used as the upcoming files only use a reference that already exists on the browser.

With this change this variable of select2 is isolated from this change, dropdowns on TEC world are executed using a custom logic that calls  `select2TEC` meaning that this shouldn't affect any existing dropdown, and should allow for plugins and themes to use any version of `select2`. 

https://github.com/the-events-calendar/tribe-common/blob/master/src/resources/js/dropdowns.js#L454

borkweb (approved):

PR #1605, opened 2021-05-29
Make sure cache is retrived correctly

mitogh:
Make sure the pieces from the cache start at `0` as all the index start
on that number we need to make sure we get the first element in this
case the element with index `zero` from there we build the subsequent
pieces.

Then the function if was able to build the elements out of the cache it
should return the values from the cache instead of returning `true`.

As the caller to the `chunkable_transient` should build the results out
of the cache.

Requires: https://github.com/the-events-calendar/tribe-common/pull/1606

### DIFF
--- a/src/Tribe/Cache.php
+++ b/src/Tribe/Cache.php
@@ -564,7 +564,7 @@ class Tribe__Cache implements ArrayAccess {
 		} while ( ! empty( $chunk ) );
 
 		// Remove any piece of data that was added but is not relevant.
-		$chunks = array_filter($chunks);
+		$chunks = array_filter( $chunks );
 
 		if ( empty( $chunks ) ) {
 			return false;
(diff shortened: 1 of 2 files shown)


src/Tribe/Cache.php
```diff
@@ -556,27 +556,31 @@ public function get_chunkable_transient( $id, $expiration_trigger = '' ) {
...
+		$i      = 0;
 		do {
 			$chunk_transient            = $transient . '_' . $i++;
 			$chunk                      = get_transient( $chunk_transient );
 			$chunks[ $chunk_transient ] = (string) $chunk;
 		} while ( ! empty( $chunk ) );
 
-		if ( empty( array_filter( $chunks ) ) ) {
+		// Remove any piece of data that was added but is not relevant.
+		$chunks = array_filter($chunks);
```
bordoni, line 567:
```suggestion
		$chunks = array_filter( $chunks );
```

bordoni (changes requested):
One small change.

borkweb (approved):
Nice!

PR #1608, opened 2021-06-09
Fix/tec 3396 aria attributes

Camwyn:
These changes put an aria-compliant <title> element in the featured, recurring, and virtual icons.

Includes a default title, that can be overwritten via the passed param `icon_title`.


src/views/v2/components/icons/recurring.php
```diff
@@ -19,5 +20,12 @@
...
\ No newline at end of file
+<svg <?php tribe_classes( $svg_classes ); ?>  role="img" viewBox="0 0 14 14" xmlns="http://www.w3.org/2000/svg">
+<title><?php echo esc_attr( $icon_title ) ?></title>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M9 3.296c0 .039 0 .078-.012.104 0 .013-.012.04-.012.052a.54.54 0 01-.118.222L7.27 5.422a.479.479 0 01-.711 0 .61.61 0 010-.783l.734-.809H4.781c-1.53 0-2.785 1.37-2.785 3.066 0 .821.284 1.591.817 2.165a.61.61 0 010 .782.478.478 0 01-.71 0C1.39 9.061 1 8.017 1 6.91c0-2.296 1.695-4.161 3.78-4.161h2.525l-.735-.809a.61.61 0 010-.782.482.482 0 01.711 0L8.87 2.904c.059.066.094.144.118.222 0 .013.012.04.012.052v.118zM13 7.091c0 2.296-1.695 4.161-3.78 4.161H6.694l.735.809a.582.582 0 010 .783.479.479 0 01-.711 0l-1.577-1.761a.569.569 0 01-.118-.222c0-.013-.012-.04-.012-.052C5 10.769 5 10.743 5 10.704c0-.039 0-.078.012-.104 0-.013.012-.04.012-.052a.54.54 0 01.118-.222L6.73 8.578a.482.482 0 01.711 0 .582.582 0 010 .783l-.734.809H9.23c1.529 0 2.785-1.37 2.785-3.066 0-.821-.284-1.591-.818-2.165a.582.582 0 010-.782.482.482 0 01.712 0C12.609 4.927 13 5.97 13 7.09z" stroke-width=".25"/>
```
paulmskim, line 30: indent

src/views/v2/components/icons/virtual.php
```diff
@@ -19,5 +19,15 @@
 if ( ! empty( $classes ) ) {
 	$svg_classes = array_merge( $svg_classes, $classes );
 }
+
+if ( empty( $icon_title ) ) {
+	$icon_title = __( 'Virtual Event', 'tribe-common' );
```
paulmskim, line 24: Should this just be `Virtual`? We have replacement words for `Event` if users use a different word.

paulmskim (changes requested):
Should we add tests for passing in `icon_title` as an argument?

src/views/v2/components/icons/virtual.php
Camwyn: probably - but not in common?
These are meant to be fallbacks while we update all the calls to them across plugins.

Camwyn: We _could_ instead have them all default to _no_ label
TBH I am unsure which is the better approach for a few of these (most will fall back to no label, but the three in this PR are used extensively)

paulmskim: I know, I think keeping common generic would be better. We don't use `__( 'Featured Event', 'tribe-common' )` or `Recurring` on the other icons.

Camwyn: How about we invert it to empty by default, and in the PR for VE I'll change all the calls to it to pass the label?

More work, but best for users?

bordoni: I like your approach Stephen, but leaving it a generic name like `Featured` instead of empty is also ok I think.

bordoni (approved):

src/views/v2/components/icons/featured.php
```diff
@@ -19,5 +19,12 @@
...
 	$svg_classes = array_merge( $svg_classes, $classes );
 }
+
+if ( empty( $icon_title ) ) {
+	$icon_title = __( 'Featured', 'tribe-common' );
+}
 ?>
-<svg <?php tribe_classes( $svg_classes ); ?> viewBox="0 0 8 10" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M0 0h8v10L4.049 7.439 0 10V0z"/></svg>
+<svg <?php tribe_classes( $svg_classes ); ?> viewBox="0 0 8 10" xmlns="http://www.w3.org/2000/svg">
+	<title><?php echo esc_attr( $icon_title ) ?></title>
```
paulmskim, line 28: this and all other titles should be `esc_html()`

paulmskim (commented):
One more thing and should be good to go

bordoni (approved):

PR #1610, opened 2021-06-15
Feature/ve 352 ve 353 switch slide toggle

jesseeproductions:
This PR is to add two templates to common so they can be used with all the plugins. 

The switch is already in use on VE Multiple Zoom Accounts release ( 1.5 )

This adds a slide toggle: https://cloudup.com/cCL_dTxw-nL

And the switch:  https://cloudup.com/iFRC52-DKjh

This PR updates VE to use common: https://github.com/the-events-calendar/events-virtual/pull/281


src/admin-views/components/slide-toggle.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+ * View: Slide Toggle
+ *
+ * Override this template in your own theme by creating a file at:
+ * [your-theme]/tribe/admin-views/components/switch.php
```
paulmskim, line 6: override path

src/resources/js/slide-toggle.js
```diff
@@ -0,0 +1,95 @@
...
+ * Initializes in a Strict env the code that manages the Slide Toggle.
+ *
+ * @since TBD
+ *
+ * @param  {PlainObject} $   jQuery
+ * @param  {PlainObject} obj tribe.events.slideToggle
+ *
+ * @return {void}
+ */
+(function ( $, obj ) {
```
paulmskim, line 30: I don't have the full context here, but a couple thoughts:
1. We have an `accordion.js` in TEC, I wonder if we can copy over to common and use it there rather than creating yet another JS file. It's a generic accordion functionality that sets up an accordion based on the aria attributes. We can think about a replacement plan for the TEC accordion JS file later, but that functionality should have belonged to common anyway.
2. Curious on the name slide toggle, rather than accordion or something else, was there a specific reason for this name?

src/Tribe/Field.php
```diff
@@ -231,7 +231,7 @@ public function do_field() {
 				} else {
 
 					// fail, log the error
-					Tribe__Main::debug( esc_html__( 'Invalid field type specified', 'tribe-common' ), $this->type, 'notice' );
+					Tribe__Debug::debug( esc_html__( 'Invalid field type specified', 'tribe-common' ), $this->type, 'notice' );
```
jesseeproductions, line 234: I ran into a fatal with this, the method does not live in main.

src/resources/js/slide-toggle.js
jesseeproductions: Hi @paulmskim 
Thanks, I missed that script. So I am changing to use the accordion.js in tec. We had already used it in VE, but then that part had been removed. 

I named it after the jquery function, but it is gone now and only the switch will go to common.

src/admin-views/components/slide-toggle.php
```diff
@@ -0,0 +1,70 @@
...
+?>
+<div <?php tribe_classes( $toggle_wrap_classes ); ?>>
+	<button
+		id="<?php echo esc_attr( $label_id ); ?>"
+		<?php tribe_classes( $toggle_button_classes ); ?>
+		aria-expanded="<?php echo $expanded ? 'true' : 'false'; ?>"
+		aria-controls="<?php echo esc_attr( $panel_id ); ?>"
+	>
+		<?php
+		echo esc_html( $label );
```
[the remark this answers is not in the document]
jesseeproductions: I did add this, but in VE.

paulmskim (approved):

PR #1611, opened 2021-06-15
Ensure we pass an array.

Camwyn:
In php 8, when using `call_user_func_array` and `array_merge` we need to explicitly pass an array.
Casting does not work.

Tester: http://sandbox.onlinephpfunctions.com/code/074efaf5f90fe7d4b2d750329890530b449ca9d0 (allows switching php versions)

To be honest, I am not sure of the true (code) object here - if we're trying to "flatten the array" it might be better to do so in the `foreach` above the changed code?

[TEC-3945]

[TEC-3945]: https://theeventscalendar.atlassian.net/browse/TEC-3945


src/Tribe/Cost_Utils.php
```diff
@@ -346,7 +346,7 @@ public function parse_cost_range( $costs, $max_decimals = null, $sort = true ) {
 		}
 
 		$output_costs = [];
-		$costs        = call_user_func_array( 'array_merge', $costs );
+		$costs        = call_user_func_array( 'array_merge', [ $costs ] );
```
borkweb, line 349: If `$costs` is already an array here, this would cause a problem. I'd suggest the following:

```suggestion
		if ( ! empty( $costs ) && ! is_array( $costs ) ) {
			$costs        = call_user_func_array( 'array_merge', [ $costs ] );
		}
```

borkweb (changes requested):

Camwyn:
Updated tester link in description to more accurately reflect what we're doing in the plugin (and use newest version of this change)

src/Tribe/Cost_Utils.php
Camwyn: I dug a bit more - the real issue turns out to be strings for keys.
`array_values` fixes that - and we don't use the keys anyway?

Paraphrased:

As a clarification for why this is failing, it's due to the new named argument assignment feature in PHP 8. In prior versions of PHP the keys were ignored and each item in your array passed to call_user_func_array was assigned in-order to the arguments of the function you were attempting to call. Basically: array_merge($array['one'], $array['two'])

As of PHP 8.0, it's trying to match the keys of the array to the function argument name, you'll note that if you change one of the indexes to "array" and remove the other, this works on the latest PHP 8 RC, as the first argument in array_merge is labelled as '$array' and thus the array index of your array gets mapped to that. Basically: array_merge(one: $array['one'], two: $array['two'])

src/Tribe/Cost_Utils.php
```diff
@@ -346,7 +346,7 @@ public function parse_cost_range( $costs, $max_decimals = null, $sort = true ) {
 		}
 
 		$output_costs = [];
-		$costs        = call_user_func_array( 'array_merge', $costs );
+		$costs        = call_user_func_array( 'array_merge', array_values( $costs ) );
```
borkweb, line 349: `array_values()` accepts an array too, so this will likely have a similar issue. It _is_ ok to add a wrapping if statement.

borkweb: I just saw your reply, though. That makes sense as long as `$costs` is an array :aw_yeah:

borkweb (approved):

PR #1619, opened 2021-07-11
Allow external sources to change the ID of the post

mitogh:
This would help in redirecting subsequent cals from this function into
the "Real" post ID, specially in the iCal context.

Where the ID of the posts are not actual post IDs.


lucatume (commented):
While there is nothing wrong with this solution, the `post_id_helper` method is not the only way we fetch posts. My guess is we use `get_post`, probably, as much. And third-party code will use `get_post` way more often. To have the maximum level of compatibility, I think a solution filtering `get_post` would be more efficient, if a bit more complex, and would grant us more compatibility. Athough: I might not have all the context of this PR.

mitogh:
Right the main idea with this one is not to replace `get_post` but use it as an extension of that, and as a mechanism to patch our own code along the way as this would allow to redirect to the real post ID from a provisional ID in all the calls that request the ID of an event using this method, so the idea with this one is just take advantage of the cache populated by the filter on `get_post` and redirect to an instance that exists on the meta table or has relationship with taxonomies and such. @lucatume

lucatume (dismissed):

lucatume:
This PR is part of the work for RBE, **but is not required for the Alpha version of it to work correctly**. As such, it can wait the next release to be merged in the codebase.

src/Tribe/Main.php
```diff
@@ -557,7 +557,20 @@ public static function array_insert_before_key( $key, $source_array, $insert_arr
...
+		/**
+		 * Allow for a modify of the post ID in order to allow redirection of values before any other additional
+		 * wordpress action is called from this result.
+		 *
+		 * @since TBD
+		 *
+		 * @param int|bool         $post_id   The ID of the post if the $candidate value is a valid WP_Post Object, `false` otherwise.
+		 * @param null|int|WP_Post $candidate Post ID or object, `null` to get the ID of the global post object.
+		 *
+		 * @return  int|bool The ID of the post/
```
lucatume, line 571:
```suggestion
		 * @return  int|bool The ID of the post.
```

src/Tribe/Main.php
```diff
@@ -557,7 +557,20 @@ public static function array_insert_before_key( $key, $source_array, $insert_arr
 	public static function post_id_helper( $candidate = null ) {
 		$candidate_post = get_post( $candidate );
 
-		return $candidate_post instanceof WP_Post ? $candidate_post->ID : false;
+		$post_id = $candidate_post instanceof WP_Post ? $candidate_post->ID : false;
+
+		/**
+		 * Allow for a modify of the post ID in order to allow redirection of values before any other additional
```
lucatume, line 563:
```suggestion
		 * Allows modifying the post ID in order to allow redirection of values before any other additional
```

src/Tribe/Main.php
```diff
@@ -557,7 +557,20 @@ public static function array_insert_before_key( $key, $source_array, $insert_arr
...
+		 * wordpress action is called from this result.
```
lucatume, line 564:
```suggestion
		 * WordPress action is called from on result.
```

lucatume (approved):

borkweb (dismissed):
:+1:

borkweb (approved):

PR #1620, opened 2021-07-13
[CE-97] Fix ampersands displaying encoded

redscar:
[CE-97]

Ampersands were displaying encoded. This change displays them properly for all drop-downs that gather the data via Ajax. The change was made because of a bug reported in CE. If you had any categories or tags with an ampersand they would display escaped.

image

image

[CE-97]: https://theeventscalendar.atlassian.net/browse/CE-97

### DIFF
--- a/src/Tribe/Ajax/Dropdown.php
+++ b/src/Tribe/Ajax/Dropdown.php
@@ -88,6 +88,10 @@ class Tribe__Ajax__Dropdown {
 			}
 		}
 
+		foreach ( $results as $result ) {
+			$result->text = htmlspecialchars_decode( $result->text );
+		}
+
 		$data['results']    = $results;
 		$data['taxonomies'] = get_taxonomies();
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Ajax/Dropdown.php
```diff
@@ -88,6 +88,10 @@ public function search_terms( $search, $page, $args, $source ) {
 			}
 		}
 
+		foreach ( $results as $result ) {
+			$result->text = htmlspecialchars_decode( $result->text );
```
borkweb, line 92: Let's make sure we escape scary stuff:

```suggestion
			$result->text = wp_kses_post( htmlspecialchars_decode( $result->text ) );
```

borkweb (changes requested):
One tiny change!

bordoni (dismissed):

bordoni (approved):

PR #1623, opened 2021-07-28
Don't tell wp_query it's a search if the search term(s) are empty.

Camwyn:
The issue:
Wild snippets out there try to exclude pages from post searches and in the process remove all post types _except_ `post`.

This breaks our views - as the repository adds a blank search to every event query ( `'s' = ''` ) triggering `$wp_query->is_search` to be true.

This just tests for a search term before adding the search param to the query.

[TEC-4012]

[TEC-4012]: https://theeventscalendar.atlassian.net/browse/TEC-4012


src/Tribe/Repository.php
```diff
@@ -1794,7 +1794,9 @@ protected function apply_default_modifier( $key, $value ) {
 				$args = [ 'p' => $value ];
 				break;
 			case 'search':
-				$args = [ 's' => $value ];
+				if ( ! empty( $value ) ) {
```
lucatume, line 1797: I would make the comparison stricter. String zero, (`"0"`) will evaluate to empty while still being a legit, albeit strange, search term.
```suggestion
				if ( '' !== $value ) ) { 
```

lucatume (changes requested):

bordoni (approved):

bordoni (approved):

PR #1628, opened 2021-08-04
Fix QA fail issues for the In-App help Page

EugeneKyale:
@paulmskim I've made some updates to address some issues that were raised by QA. Here's a reference : https://theeventscalendar.atlassian.net/browse/TEC-3741?focusedCommentId=98939

@paulmskim please see these videos describing what the PR fixes : https://www.loom.com/share/4fb4cd274055408497de99675a30ed06 and https://www.loom.com/share/6b3c7012b5074876801c6e92cdd8388f

### DIFF
--- a/src/admin-views/troubleshooting/ea-status.php
+++ b/src/admin-views/troubleshooting/ea-status.php
@@ -22,22 +22,22 @@ $show_third_party_accounts = ! is_network_admin();
 		<?php
 			// license key
 			include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/troubleshooting/ea-status/license-key.php';
-				// if EA is not active, bail out of the rest of this
-				if ( $ea_active ) {
-					// current usage
-					include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/troubleshooting/ea-status/current-usage.php';
-					// server connection
-					include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/troubleshooting/ea-status/server-connection.php';
-					// scheduler status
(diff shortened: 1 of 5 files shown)


paulmskim:
@EugeneKyale please add artifacts of the changes that were made.

src/admin-views/troubleshooting/ea-status.php
```diff
@@ -17,28 +17,30 @@
...
+<div class="tribe-events-admin__ea-status-table-wrapper">
+	<table class="tribe-events-admin__ea-status-table">
+		<?php
+			// license key
+			include_once Tribe__Main::instance()->plugin_path . 'src/admin-views/troubleshooting/ea-status/license-key.php';
+				// if EA is not active, bail out of the rest of this
+				if ( ! $ea_active ) {
+					echo '</table></div>';
+					return;
+				} else {
```
paulmskim, line 29: spacing. also, this case, it would be better if we just checked if ea is active, rather than echoing the closing elements if not active.

```
if ( $ea_active ) {
	// do all includes
}
```

Having 2 places for closing tags makes this brittle and requires us to change multiple places rather than one place

paulmskim (changes requested):

src/admin-views/troubleshooting/ea-status.php
EugeneKyale: @paulmskim yeah that makes sense, I've updated the conditional statement. Working on the artifacts now.

paulmskim (approved):

PR #1642, opened 2021-08-18
Add a new warning dialog

borkweb:
This PR does two things of note:

1. It adjusts the Confirm dialog to use properly styled buttons. I'm not worried about making this change as we do not use the confirm dialog anywhere in our plugins at this time.
2. It adds a new Warning dialog where the Cancel button is the button of emphasis rather than the OK button.

🎫 [ECP-901]
🎥 
Capture

Related: https://github.com/the-events-calendar/events-pro/pull/1811

[ECP-901]: https://theeventscalendar.atlassian.net/browse/ECP-901

### DIFF
--- a/src/views/dialog/confirm.php
+++ b/src/views/dialog/confirm.php
@@ -25,8 +25,8 @@ $vars        = get_defined_vars();
 
 		<?php echo $content; ?>
 		<div class="tribe-dialog__button_wrap">
-			<button class="tribe-dialog__button tribe-dialog__button-cancel tribe-common-c-btn-border tribe-common-c-btn-border--alt"><?php echo esc_html( $cancel_button_text ); ?></button>
-			<button class="tribe-dialog__button tribe-dialog__button-confirm tribe-common-c-btn tribe-common-c-btn-border"><?php echo esc_html( $continue_button_text ); ?></button>
+			<button <?php tribe_classes( $cancel_button_classes ); ?>><?php echo esc_html( $cancel_button_text ); ?></button>
+			<button <?php tribe_classes( $continue_button_classes ); ?>><?php echo esc_html( $continue_button_text ); ?></button>
 		</div>
 	</div>
 </script>
--- a/src/views/dialog/warning.php
+++ b/src/views/dialog/warning.php
@@ -25,8 +25,8 @@ $vars        = get_defined_vars();
 
 		<?php echo $content; ?>
 		<div class="tribe-dialog__button_wrap">
-			<button class="tribe-dialog__button tribe-dialog__button-confirm tribe-common-c-btn-border tribe-common-c-btn-border--alt"><?php echo esc_html( $continue_button_text ); ?></button>
-			<button class="tribe-dialog__button tribe-dialog__button-cancel tribe-common-c-btn tribe-common-c-btn-border"><?php echo esc_html( $cancel_button_text ); ?></button>
+			<button <?php tribe_classes( $continue_button_classes ); ?>><?php echo esc_html( $continue_button_text ); ?></button>
+			<button <?php tribe_classes( $cancel_button_classes ); ?>><?php echo esc_html( $cancel_button_text ); ?></button>
 		</div>
 	</div>
 </script>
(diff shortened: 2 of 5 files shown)


Camwyn (approved):
Looks good!

src/views/dialog/confirm.php
```diff
@@ -25,8 +25,8 @@
 
 		<?php echo $content; ?>
 		<div class="tribe-dialog__button_wrap">
-			<button class="tribe-button tribe-confirm__cancel"><?php echo esc_html( $cancel_button_text ); ?></button>
-			<button class="tribe-button tribe-confirm__continue"><?php echo esc_html( $continue_button_text ); ?></button>
+			<button class="tribe-dialog__button tribe-dialog__button-cancel tribe-common-c-btn-border tribe-common-c-btn-border--alt"><?php echo esc_html( $cancel_button_text ); ?></button>
```
juanfra, line 28: Would it make sense to have these classes coming from the template vars and use `tribe_classes`? I can see some scenarios where we may want the button styles to be inverse of how they are right now. (i.e.: We may want the OK/Continue to have the accent color as bg color).

src/views/dialog/warning.php
```diff
@@ -0,0 +1,32 @@
...
+<?php $dialog_view->template( 'button', $vars, true ); ?>
+<script data-js="<?php echo esc_attr( 'dialog-content-' . $id ); ?>" type="text/template" >
+	<div <?php tribe_classes( $content_classes  ) ?>>
+		<?php if ( ! empty( $title ) ) : ?>
+			<h2 <?php tribe_classes( $title_classes ) ?>><?php echo esc_html( $title ); ?></h2>
+		<?php endif; ?>
+
+		<?php echo $content; ?>
+		<div class="tribe-dialog__button_wrap">
+			<button class="tribe-dialog__button tribe-dialog__button-confirm tribe-common-c-btn-border tribe-common-c-btn-border--alt"><?php echo esc_html( $continue_button_text ); ?></button>
```
juanfra, line 28: The same comment for these ones, regarding the classes coming from vars.

src/views/dialog/warning.php
```diff
@@ -0,0 +1,32 @@
...
+$vars        = get_defined_vars();
+?>
+<?php $dialog_view->template( 'button', $vars, true ); ?>
+<script data-js="<?php echo esc_attr( 'dialog-content-' . $id ); ?>" type="text/template" >
+	<div <?php tribe_classes( $content_classes  ) ?>>
+		<?php if ( ! empty( $title ) ) : ?>
+			<h2 <?php tribe_classes( $title_classes ) ?>><?php echo esc_html( $title ); ?></h2>
+		<?php endif; ?>
+
+		<?php echo $content; ?>
```
juanfra, line 26: Do we want to escape this?

juanfra (commented):
Looking good! A couple of questions

src/views/dialog/warning.php
borkweb: That content will often have HTML

juanfra: Or wp_kses

juanfra: ^ Oops haha, this last comment was left there before I saw yours. Yes, I realized that this could have HTML when I was checking the other PR. Maybe wp_kses could make sense here then?

src/views/dialog/confirm.php
borkweb: Excellent idea. Addressed!

bordoni (approved):

src/views/dialog/warning.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * Warn Dialog View Template
+ * The confirmation template for tribe-dialog.
+ *
+ * Override this template in your own theme by creating a file at [your-theme]/tribe/dialogs/warn.php
```
paulmskim, line 6: this should probably be `/tribe/dialog/warning.php`

paulmskim (commented):
just one quick comment

PR #1650, opened 2021-09-08
Tweak some of the Troubleshooting page optin code.

Camwyn:
~Prior to this, all calls to set the optin key would fail.~

~The old code made calls to https://theeventscalendar.com/wp-json/tribe_system/v2/customer-info/XXX~

~It should be going to https://support-api.tri.be/sysinfo/optin/XXX~

~There is a redirect for that in the code base for tec.com - but it is not working apparently.~

~tec.com refs:~
~https://github.com/the-events-calendar/theeventscalendar.com/blob/2f1dc6e9d4d186148eaf9d74302ec580003a64a0/systems/nginx/default#L104~
~https://github.com/the-events-calendar/theeventscalendar.com/blob/2f1dc6e9d4d186148eaf9d74302ec580003a64a0/ansible/group_vars/all/nginx.yml#L56~
~https://github.com/the-events-calendar/theeventscalendar.com/blob/2f1dc6e9d4d186148eaf9d74302ec580003a64a0/content/plugins/pue-glue/src/Admin/Legacy_Rewrite.php#L25~

📷 https://d.pr/i/5EjitG

Note: the URL "fix" has been ported to tec.com - so all that remains here is some tweaked handling of the response, and the addition of the missing DOM element.

[TEC-4052]

[TEC-4052]: https://theeventscalendar.atlassian.net/browse/TEC-4052

### DIFF
--- a/src/Tribe/Support.php
+++ b/src/Tribe/Support.php
@@ -417,13 +417,13 @@ if ( ! class_exists( 'Tribe__Support' ) ) {
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
 			$url = $url ? $url : urlencode( str_replace( [ 'http://', 'https://' ], '', get_site_url() ) );
 
-			$teccom_url = 'https://theeventscalendar.com/';
+			$teccom_url = 'https://support-api.tri.be';
 
 			if ( defined( 'TEC_URL' ) ) {
 				$teccom_url = trailingslashit( TEC_URL );
 			}
 
-			$query = $teccom_url . 'wp-json/tribe_system/v2/customer-info/' . $optin_key . '/' . $url;
+			$query = $teccom_url . 'sysinfo/optin/' . $optin_key . '/' . $url;
 
 			if ( $remove ) {
 				$query .= '?status=remove';
--- a/src/Tribe/Support.php
+++ b/src/Tribe/Support.php
@@ -415,15 +415,14 @@ if ( ! class_exists( 'Tribe__Support' ) ) {
 		 * @param null $pueadd boolean to disable messaging when coming from pue script
 		 */
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
-			$url = $url ? $url : urlencode( str_replace( [ 'http://', 'https://' ], '', get_site_url() ) );
-
-			$teccom_url = 'https://support-api.tri.be';
+			$url        = $url ? $url : urlencode( str_replace( [ 'http://', 'https://' ], '', get_site_url() ) );
+			$teccom_url = 'https://theeventscalendar.com';
 
 			if ( defined( 'TEC_URL' ) ) {
-				$teccom_url = trailingslashit( TEC_URL );
+				$teccom_url = TEC_URL;
 			}
 
-			$query = $teccom_url . 'sysinfo/optin/' . $optin_key . '/' . $url;
+			$query = trailingslashit( $teccom_url ) . 'wp-json/tribe_system/v2/customer-info/' . $optin_key . '/' . $url;
 
 			if ( $remove ) {
 				$query .= '?status=remove';
--- a/src/Tribe/Support.php
+++ b/src/Tribe/Support.php
@@ -416,13 +416,8 @@ if ( ! class_exists( 'Tribe__Support' ) ) {
 		 */
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
 			$url        = $url ? $url : urlencode( str_replace( [ 'http://', 'https://' ], '', get_site_url() ) );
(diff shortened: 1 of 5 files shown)


src/resources/js/admin/help-page.js
```diff
@@ -84,10 +84,30 @@ tribe.helpPage = tribe.helpPage || {};
 			ajaxurl,
 			request,
 			function ( results ) {
+
 				if ( results.success ) {
 					obj.$system_info_opt_in_msg.html( "<p class='optin-success'>" + results.data + "</p>" );
 				} else {
-					obj.$system_info_opt_in_msg.html( "<p class='optin-fail'>" + results.data.message + " Code:" + results.data.code + " Status:" + results.data.data.status + "</p>" ); // eslint-disable-line max-len
+					var html = "<p class='optin-fail'>Something has gone wrong!</p>";
```
juanfra, line 91: Do we want to move out from var now that we allow es6?

src/resources/js/admin/help-page.js
```diff
@@ -84,10 +84,30 @@ tribe.helpPage = tribe.helpPage || {};
...
+
+					if ( results.data ) {
+						if ( results.data.message ) {
+							html += ' ' + results.data.message;
```
juanfra, line 95: Is there any reason why we're not wrapping the return in `<p>` here as we do with the rest?

src/resources/js/admin/help-page.js
```diff
@@ -84,10 +84,30 @@ tribe.helpPage = tribe.helpPage || {};
...
+						} else if (  results.message) {
```
juanfra, line 96:
```suggestion
						} else if ( results.message ) {
```

src/resources/js/admin/help-page.js
```diff
@@ -84,10 +84,30 @@ tribe.helpPage = tribe.helpPage || {};
...
+							html += '<p>Code:' + results.data.code + '</p>';
+						}
+
+						if ( results.data.status ) {
+							html += '<p>Status:' + results.data.status + '</p>';
+						}
+
+					}
+
+					obj.$system_info_opt_in_msg.html(  html ); // eslint-disable-line max-len
```
juanfra, line 110:
```suggestion
					obj.$system_info_opt_in_msg.html( html ); // eslint-disable-line max-len
```

juanfra (commented):
A few comments on the code. I'm not super familiar with the URLs where we use to send the data. Would it be maybe a possibility to have an artifact of this working? Finally, this would go with a changelog entry, right?

src/Tribe/Support.php
```diff
@@ -417,13 +417,13 @@ public static function ajax_sysinfo_optin() {
 		public static function send_sysinfo_key( $optin_key = null, $url = null, $remove = null, $pueadd = false ) {
 			$url = $url ? $url : urlencode( str_replace( [ 'http://', 'https://' ], '', get_site_url() ) );
 
-			$teccom_url = 'https://theeventscalendar.com/';
+			$teccom_url = 'https://support-api.tri.be';
```
borkweb, line 420: I would rather not change this in the plugin as the URL isn't going to be tri.be for ever and we do not have control of rewrite rules for tri.be domains. I'll work on getting the theeventscalendar.com url to redirect.

src/resources/js/admin/help-page.js
Camwyn: nope - I missed that one (added the wrappers as an afterthought) thanks!

src/resources/js/admin/help-page.js
Camwyn: Probably. My linter was pestering me about it so I just changed it.

src/Tribe/Support.php
Camwyn: @borkweb - I totally understand that logic, should I close this PR then or just revert the URL change?

borkweb:
I have a Pull Request out for review to update the rewrite rules for TEC.com such that the rule in question functions as expected: https://github.com/the-events-calendar/theeventscalendar.com/pull/2373

src/Tribe/Support.php
Camwyn: I've reverted the URL change

src/resources/js/admin/help-page.js
Camwyn: I'm going to leave this one for now - as it follows the convention in the rest of the file.

EugeneKyale (dismissed):

EugeneKyale:
Looks good to me but might need a different set of eyes to make sure I'm not missing anything.

src/resources/js/admin/help-page.js
borkweb: `Something has gone wrong!` needs to be translatable. I would suggest registering that string and the strings `Code:` and `Status:` on lines 101 and 105 get translated and made available in a JS object so this object can use those translated values.

bordoni (approved):

PR #1660, opened 2021-09-27
[TEC-4064] Ensure all SVG elements have unique IDs.

EugeneKyale:
This PR ensures SVG elements have unique IDs that are randomly generated. 

Ticket : https://theeventscalendar.atlassian.net/browse/TEC-4064
Artifact : https://www.loom.com/share/0a10361662d64361ba37b5cd0d18c3e6

Edit:
Implements a cached incremental value for ids via a new `tribe_get_next_cached_increment()` function in utils.

I considered at this point just removing all the class logic, and relying on `tribe_get_next_cached_increment()`.
But I decided it was better that if we did pass a class-generated ID we had a known target for CSS/JS.

Props to @borkweb for the fix.


src/views/v2/components/icons/list.php
```diff
@@ -15,9 +15,11 @@
  *
  */
 $svg_classes = [ 'tribe-common-c-svgicon', 'tribe-common-c-svgicon--list' ];
+$clip0 = rand();
+$clip1 = rand();
```
paulmskim, line 19: We do not want to randomly set an id. We want to set an id but make sure it's not going to clash with other icons. Let's add a sensible id, maybe something related to the class (`tribe-common-c-svgicon__list-clip-path-0`) or something similar.

paulmskim (dismissed):
I don't know if I fully understand what the issue is here. We have 2 clip paths, both with different ids. Is the problem that they clash with another icon's clip path ids? Or is it something else?

src/views/v2/components/icons/list.php
EugeneKyale: What's happening here is that the icon appears twice in the list view hence why we're getting that replication. We need the IDs for both instances to be unique in this scenario.

EugeneKyale: Is there a better way to ensure the IDs are unique without generating them randomly?

Camwyn: So...
Part of the issue is that this exact svg get duplicated on the page - once for mobile and once for desktop.
Any hard-coded string is going to get duplicated. So that won't work.

...however...

We do pass a unique class to this icon here: https://github.com/the-events-calendar/the-events-calendar/blob/master/src/views/v2/components/events-bar/views.php#L37
and here: https://github.com/the-events-calendar/the-events-calendar/blob/master/src/views/v2/components/events-bar/views/list/item.php#L31

Perhaps we can "get sneaky" and use that class - and if we need to we can use a hardcoded pair for a fallback?

Something like 
```php
if ( ! empty( $classes ) ) {
	$svg_classes = array_merge( $svg_classes, $classes ); // this is already in place
	// add more here - grab the first passed class and use that as our ID prefix.
	$id_prefix = array_shift( $classes );
} else {
	$id_prefix = 'tribe-common-c-svgicon__list-clip-path-';
}
```

Then use that prefix to generate two IDs

EugeneKyale: @Camwyn  Great! this approach works well. Here's an updated screenshot of the section 
image

EugeneKyale: Also, I've checked and I don't see anywhere we're currently targeting the IDs in CSS and JS.

src/views/v2/components/icons/list.php
```diff
@@ -18,6 +18,12 @@
 
 if ( ! empty( $classes ) ) {
 	$svg_classes = array_merge( $svg_classes, $classes );
+	$id_prefix   = array_shift( $classes );
+	$clip0       = $id_prefix . '-0';
+	$clip1       = $id_prefix . '-1';
+} else {
+	$clip0       = 'tribe-common-c-svgicon__list-clip-path-' . rand();
+	$clip1       = 'tribe-common-c-svgicon__list-clip-path-' . rand();
 }
```
Camwyn, line 27: More like this:
```suggestion
	$svg_classes = array_merge( $svg_classes, $classes );
	$id_prefix   = array_shift( $classes );
} else {
	$id_prefix = 'tribe-common-c-svgicon__list-clip-path-' . rand();
}

$clip0 = $id_prefix . '-0';
$clip1 = $id_prefix . '-1';
```

Camwyn (dismissed):

Camwyn (dismissed):

src/views/v2/components/icons/list.php
```diff
@@ -18,6 +18,12 @@
 
 if ( ! empty( $classes ) ) {
 	$svg_classes = array_merge( $svg_classes, $classes );
+	$id_prefix   = array_shift( $classes );
+} else {
+	$id_prefix = 'tribe-common-c-svgicon__list-clip-path-' . rand();
 }
+
+$clip0 = $id_prefix . '-0';
+$clip1 = $id_prefix . '-1';
```
borkweb, line 27:
```suggestion
if ( ! empty( $classes ) ) {
	$svg_classes = array_merge( $svg_classes, $classes );
	$id_prefix   = array_shift( $classes );
} else {

	$id_prefix = 'tribe-common-c-svgicon__list-clip-path-';
	$id_prefix .= tribe_get_next_cached_increment( $id_prefix );
}

$clip0 = $id_prefix . '-0';
$clip1 = $id_prefix . '-1';
```


Place this function in tribe common in `src/functions/utils.php`:

```
/**
 * Get the next increment of a cached incrementable value.
 *
 * @since TBD
 *
 * @param string $key Cache key for the incrementor.
 * @param string $expiration_trigger The trigger that causes the cache key to expire.
 * @param int $default The default value of the incrementor.
 *
 * @return int
 **/
function tribe_get_next_cached_increment( $key, $expiration_trigger = '', $default = 0 ) {
	$cache = tribe( 'cache' );

	$value = (int) $cache->get( $key, $expiration_trigger, $default );
	$value++;
	$cache->set( $key, $value, \Tribe__Cache::NON_PERSISTENT, $expiration_trigger );

	return $value;
}
```

borkweb (approved):

PR #1680, opened 2021-11-29
New Price Classes Structure

moraleida:


src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param mixed $amount the value to set initially
+	 */
+	public function __construct( $amount = 0 ) {
+		$this->set_initial_representation( $amount );
+		$this->set_normalized_amount( $amount );
+		$this->update();
+		$this->class_name = trim( get_class( $this ), '\\' );
```
bordoni, line 61: I would avoid this usage of class completely, it has too many potential problems, I would much rather take a descriptive approach from the class extending this Abstract.

Like give the value a slug of sorts, so that it can be identified as the `type` of value we are talking about.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+		 * Filter the value returned for get_float() when implemented in a specific class name
+		 *
+		 * @since TBD
+		 *
+		 * @param float $value the float representation of the value
+		 * @param Abstract_Value the object instance
+		 *
+		 * @return float
+		 */
+		$value = apply_filters( "tec_common_value_{$this->class_name}_get_integer", $this->float, $this );
```
bordoni, line 131: Here use the `type` variable mentioned.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+		 * Filter the value returned for get_precision() when implemented in a specific class name
+		 *
+		 * @since TBD
+		 *
+		 * @param int $value the precision to which values will be calculated
+		 * @param Abstract_Value the object instance
+		 *
+		 * @return int
+		 */
+		$value = apply_filters( "tec_common_value_{$this->class_name}_get_precision", $this->precision, $this );
```
bordoni, line 164: Here use the `type` variable mentioned.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+			 * Filter the value returned for normalize() when implemented in a specific class name
+			 *
+			 * @since TBD
+			 *
+			 * @param float $value the normalized value
+			 * @param Abstract_Value the object instance
+			 *
+			 * @return float
+			 */
+			$value = (float) apply_filters( "tec_common_value_{$this->class_name}_normalized", (float) $value, $this );
```
bordoni, line 223: Here use the `type` variable mentioned.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	private function get_setters() {
+		$methods = $this->get_object_method_names( $this, \ReflectionProperty::IS_PRIVATE );
+
+		return array_filter( $methods, function ( $item ) {
+			return $this->is_valid_setter( $item );
+		} );
```
bordoni, line 419: I would love to avoid the Reflection and use a more descriptive approach.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+		}
+
+		$name = str_replace( [ 'set_', '_value' ], '', $name );
+
+		if ( in_array( $name, $vars, true ) ) {
+			return true;
+		}
+
+		return false;
+	}
```
bordoni, line 454: Lets take a descriptive approach to avoid reflection.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+	 */
+	private function update() {
+		$reflection = $this->get_reflection_class( $this );
+
+		foreach ( $this->get_setters() as $setter ) {
+			$method = $reflection->getMethod( $setter );
+			$method->setAccessible( true );
+			$method->invoke( $this );
+		}
+	}
```
bordoni, line 314: Same concept, take the setters from the perspective of a list of callable, and if they are not accessible it just doesn't work. No need to force the accessibility. I know it changes how certain methods are used, but it prevents the reflection, which can add complexity to the code flow for less experienced developers.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+
+		// If the initial amount did not have decimals specified, $pieces will be an array of a single
+		// numeric value, so we just return it as a float.
+		if ( 1 === count( $pieces ) && is_numeric( reset( $pieces ) ) ) {
+			return (float) reset( $pieces );
+		}
+
+		$decimal = array_pop( $pieces );
+
+		return (float) implode( '', array_merge( $pieces, [ '.', $decimal ] ) );
```
bordoni, line 296: I think this whole piece here deserves it's own method, since it has a ton of small complexities that we might want to call separately.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,455 @@
...
+			 *
+			 * @since TBD
+			 *
+			 * @param float $value the normalized value
+			 * @param Abstract_Value the object instance
+			 *
+			 * @return float
+			 */
+			return (float) apply_filters( 'tec_common_value_normalized', $value, $this );
+		}
```
bordoni, line 248: These two filters need to be applied for all values on this method.

src/Tribe/Values/ValueCalculation.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Tribe\Values;
+
+trait ValueCalculation {
```
bordoni, line 5: `Camel_Snake_Case` 
Also docblocks for all methods.

src/Tribe/Values/Price.php
```diff
@@ -0,0 +1,18 @@
...
+class Price extends Abstract_Currency {
+
+	public function set_up_currency_details() {
+		$this->use_wp_currency_locale   = tribe_get_option( 'tickets-commerce-use-currency-locale', false );
+		$this->currency_code            = tribe_get_option( 'tickets-commerce-currency-code', $this->get_currency_code() );
+		$this->currency_symbol          = tribe_get_option( 'tickets-commerce-currency-symbol', $this->get_currency_symbol() );
+		$this->currency_symbol_position = tribe_get_option( 'tickets-commerce-currency-symbol-position', $this->get_currency_symbol_position() );
+	}
+
+}
```
bordoni, line 18: This is entirely code that should live in Tickets Commerce.

src/Tribe/Values/ValueCalculation.php
```diff
@@ -0,0 +1,36 @@
...
+	public function sub_total( $multiplier ) {
+		$this->set_value( $this->multiply( $multiplier ) );
+
+		return $this;
+	}
+
+	public function total( $values ) {
+		$num = array_map( function ( $obj ) {
+			return $obj->get_float();
+		}, $values );
```
bordoni, line 16:
```suggestion
		$num = array_map( static function ( Value_Interface $obj ) {
			return $obj->get_float();
		}, $values );
```

src/Tribe/Values/ValueFormatting.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Tribe\Values;
+
+trait ValueFormatting {
```
bordoni, line 5: `Camel_Snake_Case` 
Also docblocks for all methods.

src/Tribe/Values/ValueFormatting.php
```diff
@@ -0,0 +1,87 @@
...
+	 *
+	 * @param float $value the normalized value to transform
+	 *
+	 * @param string the value rounded to the specified precision and formatted with proper separators.
+	 */
+	private function to_string( $value ) {
+
+		if ( ! $this->is_value_normalized( $value, __METHOD__ ) ) {
+			return $value;
+		}
```
bordoni, line 21: Avoid the requirement of passing the method.

src/Tribe/Values/Value_Interface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tribe\Values;
+
+interface Value_Interface {
```
bordoni, line 5: `Camel_Snake_Case` 
Also docblocks for all methods. Set the Docblocks here, and `@inheritDoc` on all others

src/Tribe/Values/ValueFormatting.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private function is_value_normalized( $value, $method ) {
+
+		if ( is_float( $value ) ) {
+			return true;
+		}
+
+		$type = gettype( $value );
+		_doing_it_wrong( esc_html_e( "$method expects a float value, $type found in $value", 'tribe-common' ) );
```
bordoni, line 82: Avoid this, it can lead to weird scenarios. I would just return a `WP_Error` that describes the problem.

src/Tribe/Values/Abstract_Currency.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Tribe\Values;
+
+abstract class Abstract_Currency extends Abstract_Value implements Currency_Interface {
```
bordoni, line 5: Docblocks for all methods. Set the Docblocks here, and `@inheritDoc` on all others

src/Tribe/Traits/Reflection_Tools.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Tribe\Traits;
+
+trait Reflection_Tools {
```
bordoni, line 5: I feel like we shouldn't need this, I would love to see us take a more descriptive approach, like mentioned on some of the other comments.

bordoni (changes requested):

src/Tribe/Values/Price.php
moraleida: This file will be removed entirely, it was just an implementation test. It will be replaced by this implementations [like this](https://github.com/the-events-calendar/event-tickets/pull/2219/files#diff-212eb06d08e7d134bea26fa4ba9dcb11e8db51c2d403d744805fc2040b44ee4c)

src/Tribe/Values/Abstract_Value.php
moraleida: @bordoni not sure I understand. Do you want me to break down `normalize` into a sequence of smaller methods? I can do that.

moraleida: Or do you want to make it a separate class?

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,423 @@
...
+	private $precision = 2;
+
+	/**
+	 * The class name representation to use when firing scoped filters
+	 *
+	 * @var string
+	 */
+	public $class_name;
+
+	public $setters;
```
codingmusician, line 53: Needs docblock.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,423 @@
...
+		 * Filter the value returned for get_float() when implemented in a specific class name
+		 *
+		 * @since TBD
+		 *
+		 * @param float $value the float representation of the value
+		 * @param Abstract_Value the object instance
+		 *
+		 * @return float
+		 */
+		$value = apply_filters( "tec_common_value_{$this->get_class_name()}_get_integer", $this->float, $this );
```
codingmusician, line 121:
```suggestion
		$value = apply_filters( "tec_common_value_{$this->get_class_name()}_get_float", $this->float, $this );

File: src/Tribe/Values/Abstract_Value.php [hunk 21]
```diff
@@ -0,0 +1,423 @@
...
+		 * Filter the value to be set as $normalized_amount for all implementations.
+		 *
+		 * @since TBD
+		 *
+		 * @param float $normalized_value the normalized value
+		 * @param Abstract_Value the object instance
+		 *
+		 * @return float
+		 */
+		$normalized_value = (float) apply_filters( "tec_common_value_normalized", $normalized_value, $this );
```
codingmusician, line 343: You might have a reason for doing it this way, but it would seem more consistent to put these within the `get_normalized_value()` method.

moraleida: There is a reason for this. In short: every single operation that takes place using any class that extends `Abstract_Value` relies on the response to `get_normalized_value()` being uniform. This makes it risky to add filters to it, because you can have filters return different values for the same object, and that has the potential of screwing up every calculation and possibly every formatting function too. Adding the filters before setting the parameter value is how we go around that problem. You can still change what's going to be in the parameter, but we make sure the value you choose to return will be the same for every calculation.

src/Tribe/Values/Value_Calculation.php
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	public function multiply( $multiplier ) {
+		return $this->get_float() * $multiplier;
+	}
+
+	public function to_integer( $value ) {
+		return (int) ( round( $value, $this->precision ) * pow( 10, $this->precision ) );
+	}
+}
```
bordoni, line 36: Doc blocks on all of these.

src/Tribe/Values/Abstract_Value.php
```diff
@@ -0,0 +1,435 @@
...
+	 * @var int
+	 */
+	private $precision = 2;
+
+	/**
+	 * The class type representation to use when firing scoped filters
+	 *
+	 * @var string
+	 */
+	public $value_type;
```
bordoni, line 51: All params should have a `@since`

src/Tribe/Values/Abstract_Currency.php
```diff
@@ -0,0 +1,337 @@
...
+
+	/**
+	 * @var string
+	 */
+	protected $currency_symbol = '$';
+
+	/**
+	 * @var string
+	 */
+	protected $currency_symbol_position = 'prefix';
```
bordoni, line 48: Alll params should have `@since`

bordoni (changes requested):

bordoni:
Closed in favor of #1686

PR #1685, opened 2021-12-09
Load `Event Countdown` widget assets on post type pages. [ECP-1016]

EugeneKyale:
This PR ensures the `Event Countdown` widget's assets are loaded onto post type pages to enable the new `Event Countdown` block to function correctly.

Related to : https://github.com/the-events-calendar/events-pro/pull/1864
Ticket : https://theeventscalendar.atlassian.net/browse/ECP-1016
Artifact : https://www.loom.com/share/f195e3c996834ce5b76bf7a943345df4


src/Tribe/Main.php
```diff
@@ -430,7 +430,7 @@ public function should_load_common_admin_css() {
 		$helper = Tribe__Admin__Helpers::instance();
 
 		// Are we on a post type screen?
-		$is_post_type = $helper->is_post_type_screen();
+		$is_post_type = $helper->is_post_type_screen() || 'post' === get_current_screen()->base;
```
bordoni, line 433: This feels like it will include a lot of other scripts in all pages. Can we track down which particular Assets we are trying to insert? I would like to see the change happening closer to the actual asset we need to include.

bordoni (changes requested):

src/Tribe/Main.php
EugeneKyale: I've identified the 2 files that we need to load and modified the approach slightly.

bordoni (approved):

PR #1693, opened 2022-01-03
dev: fix `$default` param types for `tribe_get_option()`

justlevine:
This PR changes the `$default` param in `tribe_get_option()` to `string|bool`, considering that's how TEC core uses the function.

Fixes:
image

Justification:
image


bordoni:
Closing this in favor of a different change, we will use `mixed` since it can actually be anything.

PR #1716, opened 2022-02-17
Ensure that tec-variables-full requires tec-variables-skeleton.

Camwyn:
This is due to the fact that we have (admin) styles that require full, but aren't requiring skeleton.

Can affect blue/green styles, so would like both teams to voice any objections before we add this to a release.

[TEC-4308]

[TEC-4308]: https://theeventscalendar.atlassian.net/browse/TEC-4308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juanfra (commented):
Looking good! I'd try to get a ticket from your PO so that we can track these things.

EugeneKyale (approved):
Looks good.

juanfra (approved):
🚀

PR #1724, opened 2022-03-08
[ET-1447] Modify logic of `filter_modify_to_module`.

juanfra:
[ET-1447]

Modify logic of `filter_modify_to_module` so that we can safely set as module those assets that are loaded in themes without support for `html5`, `scripts`. (WordPress is adding type in core, and we weren't accounting for that: https://github.com/WordPress/wordpress-develop/blob/8cc9e53847f8a46258d9e42c29f8d524ec0af592/src/wp-includes/script-loader.php#L2013)

https://user-images.githubusercontent.com/252415/157252395-562c6a39-7d22-4b8b-9df7-f3b4b60e5c31.mov

[ET-1447]: https://theeventscalendar.atlassian.net/browse/ET-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -181,7 +181,10 @@ class Tribe__Assets {
 		}
 
 		// Do not add Module to a Script that already has a type.
-		if ( false !== strpos( $tag, ' type=' ) ) {
+		if (
+			false !== strpos( $tag, ' type=' )
+			&& current_theme_supports( 'html5', 'script' ) // These themes already have the `type="text/javascript"` added by WordPress core.
+		) {
 			return $tag;
 		}
 
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -180,12 +180,11 @@ class Tribe__Assets {
 			return $tag;
 		}
 
-		// Do not add Module to a Script that already has a type.
-		if (
-			false !== strpos( $tag, ' type=' )
-			&& current_theme_supports( 'html5', 'script' ) // These themes already have the `type="text/javascript"` added by WordPress core.
-		) {
-			return $tag;
+		// These themes already have the `type='text/javascript'` added by WordPress core.
+		if ( ! current_theme_supports( 'html5', 'script' ) ) {
+			$replacement = 'type="module"';
+
+			return str_replace( "type='text/javascript'", $replacement, $tag );
 		}
 
 		$replacement = '<script type="module" ';


src/Tribe/Assets.php
```diff
@@ -181,7 +181,10 @@ public function filter_modify_to_module( $tag, $handle ) {
 		}
 
 		// Do not add Module to a Script that already has a type.
-		if ( false !== strpos( $tag, ' type=' ) ) {
+		if (
+			false !== strpos( $tag, ' type=' )
+			&& current_theme_supports( 'html5', 'script' ) // These themes already have the `type="text/javascript"` added by WordPress core.
```
Camwyn, line 186: Should this be `! current_theme_supports()` ?

moraleida: I don't think either of these will solve the issue alone.

At this point we already know that this is a module and that the tag has an attribute, so we need to replace that attribute to `type=module` instead of just returning it.

src/Tribe/Assets.php
```diff
@@ -181,7 +181,10 @@ public function filter_modify_to_module( $tag, $handle ) {
...
+		) {
 			return $tag;
```
moraleida, line 188:
```suggestion
               $replacement = 'type="module" ';
		return str_replace( 'type="text/javascript" ', $replacement, $tag );
```

maybe this?

moraleida (changes requested):

src/Tribe/Assets.php
moraleida: tbh, I'm not even sure we need the check for `current_theme_supports`.  if we do this `str_replace` only modules with text/javascript types will be replaced.

Modules without any type will still fall to the next return. I think this covers all of our bases

src/Tribe/Assets.php
juanfra: I was just keeping the if (checking if the type is defined), in case it touches some other thing and potentially breaks something.

At this point, we know that is a tribe asset of type module, but WP may have changed the HTML already making this of type="text/javascript" (what was happening, actually).

I'm alright with removing this `if` completely, but I was being cautious of not removing something that could potentially break other stuff.

src/Tribe/Assets.php
juanfra: [Because WP can be wiping out the type](https://github.com/WordPress/wordpress-develop/blob/8cc9e53847f8a46258d9e42c29f8d524ec0af592/src/wp-includes/script-loader.php#L2013), if the `current_theme_supports( 'html5', 'script' )`, and as a result if we get there the str_replace won't do anything because it won't find the `type="text/javascript"`

codingmusician (approved):

moraleida (approved):

PR #1727, opened 2022-03-11
[ET-1351] Create Upsell Notice Element

codingmusician:
### 🎫 Ticket

[ET-1351] 

### 🗒️ Description

New Upsell element that can be customized (see video below for more info).

### 🎥 Artifacts 

https://youtu.be/AhpjD-vycNc

### ✔️ Checklist
- [x] I've included a changelog entry in the readme.txt file. 
- [x] My code is tested. 
- [x] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).

[ET-1351]: https://theeventscalendar.atlassian.net/browse/ET-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Admin/Upsell_Notice.php
```diff
@@ -0,0 +1,100 @@
...
+	  * @since TBD
+	  *
+	  * @param array  $args Array of arguments that will ultimately be sent to the template.
+	  * @param bool   $echo Whether or not to echo the HTML. Defaults to true.
+	  *
+	  * @return string HTML of upsell notice.
+	  */
+	public function render( $args, $echo = true ) {
+		// Check if upsell should be rendered.
+		if( ! $this->should_render() ) {
```
juanfra, line 71:
```suggestion
		if ( ! $this->should_render() ) {
```

src/Tribe/Admin/Upsell_Notice.php
```diff
@@ -0,0 +1,100 @@
...
+		$args['link'] = wp_parse_args( $args['link'], [
+			'classes' => [],
+			'text'    => '',
+			'url'     => '',
+			'target'  => '_blank',
+			'rel'     => 'noopener noreferrer',
+		] );
+
+		$template = $this->get_template();
+		return $template->template( 'default', $args, $echo );
```
juanfra, line 98: Given that we're passing arguments with classes, maybe we can call it `main`? Just a suggestion, there's nothing wrong with `default`. But as a dev, I would think of `default` as something that's already default and that I should use something different if I want to customize, or something.

src/admin-views/upsell/default.php
```diff
@@ -0,0 +1,54 @@
...
+}
+
+?>
+<div <?php tribe_classes( $upsell_classes ); ?>>
+	<div class="tec-admin__upsell-content">
+		<div class="tec-admin__upsell-icon">
+			<img
+				class="tec-admin__upsell-icon-image"
+				src="<?php echo esc_url( $icon_url ); ?>"
+				alt="<?php esc_attr_e( 'The Events Calendar important notice icon', 'tribe-common' ); ?>"
```
juanfra, line 37: Would be nice if we can send this to the template.

src/admin-views/upsell/default.php
```diff
@@ -0,0 +1,54 @@
...
+			/>
+		</div>
+		<div class="tec-admin__upsell-text-wrap">
+			<span class="tec-admin__upsell-text">
+				<?php echo esc_html( $text ); ?>
```
juanfra, line 42: Maybe we will want to pass some HTML here. Probably we can go with `wp_kses_post` here to make it more flexible, in case we want to use `strong` or links within the text.

src/admin-views/upsell/default.php
```diff
@@ -0,0 +1,54 @@
...
+	$upsell_classes = array_merge( $upsell_classes, $classes );
+}
+
+$link_classes = [ 'tec-admin__upsell-link '];
+if ( ! empty( $link ) && ! empty( $link['classes'] ) ) {
+	$link_classes = array_merge( $link_classes, $link['classes'] );
+}
+
+?>
+<div <?php tribe_classes( $upsell_classes ); ?>>
```
juanfra, line 31: Probably we'll want to have `tribe-common` as a wrapper (maybe a level up) so that we have the resets, and that we can use the font that we want there.

src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -3905,3 +3905,51 @@ body.tribe-welcome #fs_connect {
 		width: 300px;
 	}
 }
+
+/* Upsell Styling */
+.tec-admin__upsell {
```
juanfra, line 3910: If we go with the `tribe-common` wrapper, these classes should be also wrapped.

src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -3905,3 +3905,51 @@ body.tribe-welcome #fs_connect {
...
+	display: inline-block;
+}
```
juanfra, line 3912: We need a blank line between these two guys.

src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -3905,3 +3905,51 @@ body.tribe-welcome #fs_connect {
 		width: 300px;
 	}
 }
+
+/* Upsell Styling */
```
juanfra, line 3909: It'll be also a good idea to include these classes in a new `.pcss` to start organizing things here.

src/admin-views/notices/upsell/icon.php
```diff
@@ -0,0 +1,20 @@
...
+ * Upsell Template
+ * The base template for TEC Upsell notices.
+ *
+ * Override this template in your own theme by creating a file at [your-theme]/tribe/upsell/upsell.php
+ *
+ * @since TBD
+ *
+ * @var string        $icon_url  URL to icon.
+ * 
+ * @version TBD
```
juanfra, line 12: Let's update the description. The admin views can't be overridden. Let's try to group the Since and version and then leave the var afterward. (the same for the other template.)

src/admin-views/notices/upsell/main.php
```diff
@@ -28,27 +28,25 @@
...
-				href="<?php echo esc_url( $link['url'] ); ?>" 
-				rel="<?php echo esc_attr( $link['rel'] ); ?>"
-				target="<?php echo esc_attr( $link['target'] ); ?>"
-			>
-				<?php echo esc_html( $link['text'] ); ?>
-			</a>
+<div class="tribe-common">
+	<div <?php tribe_classes( $upsell_classes ); ?>>
+		<div class="tec-admin__upsell-content">
+			<div class="tec-admin__upsell-icon">
```
juanfra, line 34: Should we move the wrapper to the template?

src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -3905,51 +3905,3 @@ body.tribe-welcome #fs_connect {
 		width: 300px;
 	}
 }
-
```
juanfra: This is great! what I meant was not to include it as a different asset, but leave it as tribe-common-admin and include it via `@import`. (for example, what we're doing [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/postcss/tickets.pcss) or [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/postcss/tickets-commerce.pcss))

juanfra (commented):
🥇

readme.txt
```diff
@@ -1,5 +1,9 @@
 == Changelog ==
 
+= [TBD] TBD =
+
+* Feature - New customizable upsell element to offer upgrades, additions and services that are available.
```
juanfra, line 5: Please let's include the ticket reference here.

src/resources/postcss/tribe-common-admin.pcss
```diff
@@ -3905,3 +3905,6 @@ body.tribe-welcome #fs_connect {
 		width: 300px;
 	}
 }
+
+/* Upsell CSS */
+@import "upsell.pcss";
```
juanfra, line 3910: I think that if we could put this inside a folder (can be `tribe-common-admin` or `tribe-admin`) as we do in other places it'd be perfect, so that we can start moving things there and keep it more organized. 

For example, how we do it for tickets commerce. Where we [have a main file](https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/postcss/tickets-commerce.pcss) (in this case this would be the main file) and it has calls to an [_all.pcss](https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/postcss/tickets-commerce/_all.pcss) file inside of [the folder](https://github.com/the-events-calendar/event-tickets/tree/master/src/resources/postcss/tickets-commerce). And that `_all.pcss` has calls to the different sections. 

That way we can start organizing the admin files and grow from here. This file is already around 4K lines and I'd love to see us splitting this in the future. Having this new structure can be a great beginning.

juanfra (dismissed):
Two minor things. Looking great other than that!

src/resources/postcss/tribe-common-admin.pcss
codingmusician: Gotcha. I notice that imports have to be done at the beginning of a pcss file, before any other statements, so I created a `tribe-common-admin` folder, then moved all the CSS from the `tribe-common-admin.pcss` file into `tribe-common-admin/_main.pcss` so I could import them in order within the `_all.pcss` file.

juanfra (approved):
🙌

PR #1736, opened 2022-04-11
Display alert on EA status if EA imports are disabled under Settings

jesseeproductions:
Artifacts
Enabled: https://cloudup.com/iRUr0DY0EC5
Disabled: https://cloudup.com/iYzQR2P8LX2

Ticket
_Ref:_ [TCMN-134](https://theeventscalendar.atlassian.net/browse/TCMN-134)


lucatume (approved):

src/admin-views/troubleshooting/ea-status/current-status.php
```diff
@@ -0,0 +1,41 @@
...
+$message  = esc_html_x( 'Imports Enabled in Settings', '', 'tribe-common' );
+$disabled = tribe_get_option( 'tribe_aggregator_disable', false );
+
+if ( $disabled ) {
+	$icon         = 'error';
+	$message      = _x( 'Imports disabled in Settings', '', 'tribe-common' );
+	$settings_url = Tribe__Settings::instance()->get_url( array( 'tab' => 'imports' ) );
+	$notes 		  = sprintf(
+		'<a href="%1$s">%2$s</a>',
+		$settings_url,
```
Camwyn, line 24: Might want to escape
```suggestion
		esc_url( $settings_url ),
```

Camwyn (approved):
I checked our code - the output of `get_url()` isn't escaped. **Approving** as we never escape it in practice (but _thinking_ maybe we should?)

PR #1739, opened 2022-04-13
Ensure the classic "forget" param overrides all else.

Camwyn:
This allows use of hte link in the editor to switch on the fly.

Bad: https://lw.slack.com/files/U01H2U7LZ0W/F03B1TA7CM6/screen_recording_2022-04-06_at_15.53.41.mov
Good: https://d.pr/v/o8262J

[TEC-4287]

[TEC-4287]: https://theeventscalendar.atlassian.net/browse/TEC-4287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -3,7 +3,7 @@
 = [TBD] TBD =
 
 * Tweak - Added EA status row showing if it is enabled or disabled in the Event Aggregator system status [TCMN-134]
-* Fix - Ensure the classic "forget" param overrides all else when loading the editor w/Classic Editor active. [TEC-4287]
+* Fix - Ensure the Classic Editor "forget" parameter overrides all else when loading the editor w/Classic Editor active. [TEC-4287]
 * Fix - Do not autoload options used to save batched data. [EA-427]
 * Fix - Update bootstrap logic to make sure Common will correctly load completely in the context of plugin activations requests. [TEC-4338]
 
--- a/readme.txt
+++ b/readme.txt
@@ -3,6 +3,7 @@
 = [TBD] TBD =
 
 * Tweak - Added EA status row showing if it is enabled or disabled in the Event Aggregator system status [TCMN-134]
+* Fix - Ensure the classic "forget" param overrides all else when loading the editor w/Classic Editor active. [TEC-4287]
 * Fix - Do not autoload options used to save batched data. [EA-427]
 * Fix - Update bootstrap logic to make sure Common will correctly load completely in the context of plugin activations requests. [TEC-4338]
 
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -3,6 +3,7 @@
 = [TBD] TBD =
 
 * Tweak - Added EA status row showing if it is enabled or disabled in the Event Aggregator system status [TCMN-134]
+* Fix - Ensure the classic "forget" param overrides all else when loading the editor w/Classic Editor active. [TEC-4287]
```
lucatume, line 6: Should `classic` be `Classic Editor`?

lucatume, line 6: `param` => `parameter`.

src/Tribe/Editor/Compatibility/Classic_Editor.php
```diff
@@ -236,7 +236,7 @@ public function filter_tribe_editor_should_load_blocks( bool $should_load_blocks
 		}
 
 		if ( static::get_classic_override() ) {
-			$should_load_blocks = true;
+			$should_load_blocks = ! $should_load_blocks;
```
lucatume, line 239: The previous version of the code would set `$should_load_blocks` to `true`. This one to the opposite of the previous value. The logic went from constant to contextual, is this correct? It might be, I'm asking as I would not know.

src/Tribe/Editor/Compatibility/Classic_Editor.php
```diff
@@ -264,6 +264,11 @@ public function filter_tribe_editor_should_load_blocks( bool $should_load_blocks
 			$should_load_blocks = false;
 		}
 
+		// The "forget" param still overrides all...
```
lucatume, line 267: Either full-stops, exclamation or interrogation marks to close a comment line. This comment does not really explain what is happening here and the `forget` param is mentioned but not found near this piece of code. Am I missing something?

lucatume (changes requested):

readme.txt
Camwyn: doh -yes

src/Tribe/Editor/Compatibility/Classic_Editor.php
Camwyn: Correct - the override "reverses" whatever is the current setup - so classic to block and block to classic.
ie: assuming we're overriding,  if `classic_editor` is set, then we get blocks, otherwise we get calssic

jesseeproductions (approved):

PR #1758, opened 2022-05-20
Avoid loops caused by meta update tracker.

Camwyn:
We run `filter_watch_updated_meta` on the `update_post_metadata` hook, and then we run `update_post_meta` inside it.

Let's prevent looping.

[ECP-1203]

[ECP-1203]: https://theeventscalendar.atlassian.net/browse/ECP-1203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Tracker.php
+++ b/src/Tribe/Tracker.php
@@ -282,8 +282,12 @@ class Tribe__Tracker {
 		// If we got here we will update the Modified Meta
 		$modified[ $meta_key ] = $now;
 
+		// Avoid loops!
+		remove_filter( 'update_post_metadata', [ $this, 'filter_watch_updated_meta' ], PHP_INT_MAX - 1 );
 		// Actually do the Update
 		update_post_meta( $post->ID, self::$field_key, $modified );
+		// Safe to filter again.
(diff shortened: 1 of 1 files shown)


src/Tribe/Tracker.php
```diff
@@ -282,8 +282,12 @@ public function filter_watch_updated_meta( $check, $post_id, $meta_key, $meta_va
 		// If we got here we will update the Modified Meta
 		$modified[ $meta_key ] = $now;
 
+		// Avoid loops!
+		remove_filter( 'update_post_metadata', [ $this, 'filter_watch_updated_meta' ], PHP_INT_MAX - 1 );
```
bordoni, line 286: No need to use `PHP_INT_MAX` give it a normal priority that is not 10, and we are all good.

bordoni (changes requested):

lucatume (dismissed):

bordoni (approved):

PR #1767, opened 2022-06-20
Use shortened link

zbtirrell:
We should always use shortened links and never direct ones


Camwyn:
Adding this to the B22.strudel release - low risk.

bordoni (approved):

PR #1769, opened 2022-06-22
Add some utility functions to Tribe__Admin__Helpers 

Camwyn:
...and use them in the Notice tests for admin pages.

[TEC-4422]

[TEC-4422]: https://theeventscalendar.atlassian.net/browse/TEC-4422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Admin/Helpers.php
+++ b/src/Tribe/Admin/Helpers.php
@@ -190,4 +190,81 @@ class Tribe__Admin__Helpers {
 
 		return false;
 	}
+
+	/**
+	 * Get a list of the admin screens for showing notices, etc.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string> $additional A list of additional screens to add to the list.
+	 *
+	 * @return array<string> $screens The list of screen IDs.
+	 */
+	public static function get_admin_screens_list( $additional = [] ) {
+		// @TODO: some of these are redundant or old -this list should be audited.
+		$default_screens = array_merge(
+			self::get_settings_screens_list(),
+			self::get_app_shop_screens_list()
+		);
+
+		if ( empty( $additional ) ) {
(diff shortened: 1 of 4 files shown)


src/Tribe/Admin/Helpers.php
```diff
@@ -190,4 +190,81 @@ public function is_action( $action = null ) {
...
+	 */
+	public static function get_admin_screens_list( $additional = [] ) {
+		// @TODO: some of these are redundant or old -this list should be audited.
+		$default_screens = array_merge(
+			self::get_settings_screens_list(),
+			self::get_app_shop_screens_list()
+		);
+
+		if ( empty( $additional ) ) {
+			return $default_screens;
```
juanfra, line 211: Bailing here means this value will never be filtered with `tec_admin_screens_list`. Maybe we can do a `wp_parse_args` without bailing and then returning the filtered value.

src/Tribe/Admin/Helpers.php
```diff
@@ -190,4 +190,81 @@ public function is_action( $action = null ) {
...
+	/**
+	 * Get a list of the admin settings screens for showing notices, etc.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<string> $screens The list of screen IDs.
+	 */
+	public static function get_settings_screens_list() {
+		$screens = [
+			'tribe_events_page_tec-events-settings',
```
juanfra, line 233: I feel that this should come via filter from TEC.

Camwyn: Oh, I agree - I was iterating changes 😉

PR #1771, opened 2022-06-28
Started working on static analysis

szepeviktor:
Hello! I've started analyzing `tribe-common` package.

It turns out it calls **downstream** functions. (from `the-events-calendar`)

e.g. `tribe_get_default_value`, `Widget_Abstract::format_dependency`

Could you help me resolve that? It prevents any further work.

PR #1772, opened 2022-06-28
Fix first half of PHPStan Level 0 problems

szepeviktor:
From #1771

### DIFF
--- a/src/Tribe/Repository/Decorator.php
+++ b/src/Tribe/Repository/Decorator.php
@@ -268,7 +268,7 @@ abstract class Tribe__Repository__Decorator implements Tribe__Repository__Interf
 	 * {@inheritdoc}
 	 */
 	public function save( $return_promise = true ) {
-		$this->decorated->save( $return_promise );
+		return $this->decorated->save( $return_promise );
 	}
 
 	/**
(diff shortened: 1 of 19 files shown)


lucatume:
Some more cleanup, rebased on `tweak/phpstan-setup-lvl-0`

lucatume:
Re: #1771 and downstream issues: Common is never shipped alone and is always bundled in either The Events Calendar or Event Tickets. That partially explains the downstream dependencies. It's pulled in both as a sub-module.

Sky is the limit when it comes to refactoring but, to address the current downstream dependency issue, I've [updated the `phpstan.neon.dist` file](https://github.com/the-events-calendar/tribe-common/pull/1772/commits/9cc647e4601812a05b957689d1475dd02b4e7953) to scan The Events Calendar **one level up** from Common.

Why?

If you clone [The Events Calendar](https://github.com/the-events-calendar/the-events-calendar) and build it using this command:
```bash
git clone git@github.com:the-events-calendar/the-events-calendar.git
cd the-events-calendar
git checkout -b szepeviktor-phpstan tweak/phpstan-setup-lvl-0
git submodule update --init --recursive
composer install --no-dev
cd common
composer install
composer run stan
```
The directory structure should be the following from the director that contains `the-events-calendar`:
```
.
├── the-events-calendar
│   ├── common
```
This is the directory structure the TEC team always works with: Common inside The Events Calendar. As such: it makes sense to leverage that convention for PHPStan testing.

szepeviktor:
All right. I would use a package manager (composer, npm) instead of git submodules.
But this is not my project! :)

borkweb:
We definitely considered including common as a composer dependency in the past, but the development flow for our business works more cleanly with git submodules. We have not experienced any significant issues or complications of note with submodules in recent years with the tooling that we have in place!

PR #1795, opened 2022-08-09
Fix parameters order in `implode()`

theskinnyghost:
Hello, passing the `$glue` and `$pieces` parameters in reverse order to `implode()` has been deprecated since PHP 7.4 ([see docs](https://www.php.net/manual/en/function.implode.php#refsect1-function.implode-changelog)) so this will throw a Fatal Error on PHP8+

This PR fixes the order so that it will continue to work correctly on PHP8.


Camwyn:
Thanks @theskinnyghost ! 

I was just working on this too - I'm going to merge it into my working branch and it will go out in the next maintenance release.

PR #1807, opened 2022-09-06
Prevent E_ERROR on the context usage of is_main_query() too early

bordoni:
[WP.org Report](https://wordpress.org/support/topic/site-crash-28)
---

This problem is due to some really early usage of the `tribe_context()->is( 'is_main_query' )`, which leads into having problems with some Third-party plugins that are making use of a WP_Query before the global `WP_Query` is defined.

📷 http://i.bordoni.me/UCzroOEIM0mLrYoWBPpX

### DIFF
--- a/tests/wpunit/Tribe/Context_LocationsTest.php
+++ b/tests/wpunit/Tribe/Context_LocationsTest.php
@@ -13,33 +13,49 @@ class Context_LocationsTest extends \Codeception\TestCase\WPTestCase {
 	/**
 	 * @test
 	 */
-	public function should_not_fail_on_wp_query_null() {
-		global $wp_query;
+	public function is_main_query_should_not_error_on_wp_query_null() {
+		$GLOBALS['wp_query'] = null;
 
-		$wp_query = null;
+		$context = (new \Tribe__Context());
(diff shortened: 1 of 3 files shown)


Camwyn (dismissed):

Camwyn (approved):

tests/wpunit/Tribe/Context_LocationsTest.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @since   TBD
+ *
+ * @package Tribe
+ */
+class Context_LocationsTest extends \Codeception\TestCase\WPTestCase {
+	/**
+	 * @test
+	 */
+	public function is_main_query_should_not_error_on_wp_query_null() {
```
borkweb, line 16: Do you think we need to do a `::tearDown()` where we `unset()` `$GLOBALS['wp_query']` and `$GLOBALS['wp_the_query']` if they are set?

borkweb:
You'll get errors with this latest change...

You need to do parent::setUp() and parent::tearDown()

borkweb (approved):
:+1:

PR #1812, opened 2022-09-30
Notice docblock corrections.

Camwyn:
Doing some tidying up:

1. We say it's an int but it's a Date object.
2. Some place where we say Black Friday - but we're dealing with a different sale.


skyshab (approved):
Looks good!

PR #1822, opened 2022-10-26
WP Engine System mu plugin + Divi/Avada timezone conflict. [TEC-4387]

EugeneKyale:
Ensure the TEC timezone settings are applied correctly when using a combination of the WP Engine System MU plugin and Divi or Avada Themes.

https://theeventscalendar.atlassian.net/browse/TEC-4387

https://www.loom.com/share/47416f97696f421a956bcd580faf904a


stratease (changes requested):
Looks like something that could use a test?

PR #1824, opened 2022-11-01
Handle user and site locale correctly in rewrite rule generation context

lucatume:
Ticket: [TEC-3373](https://theeventscalendar.atlassian.net/browse/TEC-3733)

Artifacts:
* tests that are already covering the issue
* dedicated tests in TEC, ECP and VE
* [Screencast of TEC, ECP and VE in action](https://share.cleanshot.com/yMMucx)

Related PRs:
* [TEC](https://github.com/the-events-calendar/the-events-calendar/pull/4031) 
* [ECP](https://github.com/the-events-calendar/events-pro/pull/2127) 
* [VE](https://github.com/the-events-calendar/events-virtual/pull/443) 

This PR, and the ones in TEC, ECP, and VE associated with this, put in place the correct handling of localized slug translations in the rewrite context by making sure the locale used to generate rewrite rules' bases is the site one, not the current user one.

Before delving into the technical side of this, a short recap of expected behaviors (I'm using Italian as the "other" language):

| Site locale | User Locale | Expected Rewrite slug locale |
|---|---|---|
|en_US|en_US|en_US|
|en_US|it_IT|en_US|
|it_IT|it_IT|it_IT|
|it_IT|en_US|it_IT|

In short: the language used to generate rewrite rules' bases should be the site one, never the user one.

### When are rewrite rules generated?
To clear the field for the following discussion, it's important to know:
1. Rewrite rules are regenerated when visiting the `/wp-admin/options-general.php` page. Just **visiting** it will trigger the rewrite rules regeneration, **no need to press Save Changes**.
2. Rewrite rules can be regenerated in the context of a REST API request.
3. Rewrite rules can be generated using wp-cli.

> Why not just redirect the user visiting `/wp-admin/options-general.php` to either `/wp-admin/options-general.php?_locale=<locale>` or `/wp-admin/options-general.php?wp_lang=<locale>`?

It would not cover all the possible venues by which rewrite rules could be generated; the request could come from wp-cli or could be programmatical, done by other code.

> Why not use an AJAX request to the site with a doctored, correct locale to do that? 

It would not work when the site cannot reach itself or there is no web-server running (i.e. in wp-cli).

### The problem
Common, TEC, ECP and VE all localize their strings, including the slugs, at `init` time. At `init` time the user locale will **override** the site locale: the loaded translations are the ones **for the user locale**, not the site locale.
The previous rewrite rules generation system would use those strings, localized at `init` time to the user language, to generate rewrite rules: this would lead to the rewrite rules being generated for the user language, not the site one.

### The solution
Before the generation of rewrite rules, if the user locale is not the same as the site locale, switch language to the site one.
This requires the plugins to use live translations to filter the rewrite bases, and not the ones created at `init` time.
Restore the current locale after the rules are generated.


bordoni:
I've moved this PR from `release/B22.tiramisu` to `master`, please review and point it to the correct branch. The next sprint will live in `release/B22.wagashi`.

borkweb:
I like the approach, @lucatume. I say: run with it!

src/Tribe/Rewrite.php
```diff
@@ -108,8 +117,9 @@ public static function instance() {
 	 *
 	 * @param WP_Rewrite|null $wp_rewrite
 	 */
-	public function __construct( WP_Rewrite $wp_rewrite = null ) {
+	public function __construct( WP_Rewrite $wp_rewrite = null, Translations_Loader $locale_switcher = null ) {
 		$this->rewrite = $wp_rewrite;
+		$this->translations_loader = tribe( Translations_Loader::class );
```
stratease, line 122:
```suggestion
		$this->translations_loader = $locale_switcher;
```
Was this the intended logic here...?

stratease (changes requested):

src/Tribe/Rewrite.php
lucatume: The idea was to fallback on the instance provided by the Service Locator, but I forgot to take into account the input parameter; good catch, thanks.

stratease (dismissed):

Camwyn:
I've updated the common PR to B23.alexa as well as this one.

PR #1823, opened 2022-10-27
[Fix] Ensure string resolution of Lazy_String

amiut:
`Lazy_String::__toString()` does not resolve to string in some cases which causes fatal error where string is expected.
In our case problem occurs in **community add event** page where `Event` object's properties like `title`, `permalink`, `excerpt`, etc. are empty and the problem happens on `wp_set_cache` level.
For example look at [this line](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/Tribe/Models/Post_Types/Event.php#L210), the result of [tribe_events_get_the_excerpt](https://github.com/the-events-calendar/the-events-calendar/blob/74a292df5169b185a7a202b2277a82d138846f70/src/functions/template-tags/general.php#L1528) function is possible to be null which makes `Lazy_String::__toString()` return null as well.

PR #1828, opened 2022-11-08
Prevent Warnings from `Lazy_String` on PHP 8.1

bordoni:
Resolves #1823

### DIFF
--- a/src/Tribe/Utils/Lazy_String.php
+++ b/src/Tribe/Utils/Lazy_String.php
@@ -71,11 +71,16 @@ class Lazy_String implements \Serializable, \JsonSerializable {
 	 */
 	public function __toString() {
 		if ( null === $this->string ) {
-			$this->string = (string) call_user_func( $this->value_callback );
+			$value =  call_user_func( $this->value_callback );
+			if ( ! $value instanceof \Generator ) {
+				$value = (string) $value;
+			}
+
+			$this->string = $value;
 			$this->resolved();
 		}
 
-		return $this->string;
+		return (string) $this->string;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Utils/Lazy_String.php
```diff
@@ -71,7 +71,12 @@ public function __construct( callable $callback, $escape_callback = 'esc_html' )
 	 */
 	public function __toString() {
```
Camwyn, line 72:
```suggestion
	public function __toString(): string {
```

src/Tribe/Utils/Lazy_String.php
```diff
@@ -71,7 +71,12 @@ public function __construct( callable $callback, $escape_callback = 'esc_html' )
...
 		if ( null === $this->string ) {
-			$this->string = call_user_func( $this->value_callback );
+			$value =  call_user_func( $this->value_callback );
+			if ( ! $value instanceof \Generator ) {
```
Camwyn, line 75: Why `Generator`?

Camwyn (changes requested):
We should enforce via the return cast.

src/Tribe/Utils/Lazy_String.php
bordoni: Because if we use `yield` it will return that.

Camwyn (approved):

tyrann0us:
Hi @bordoni, we tried to update the Events Calendar to `v6` where this fix is included. However, multiple issues made us wait for a more stable 6.* version.

Is there any chance to backport this fix to 5.*? Thanks!

PR #1837, opened 2022-11-18
Modify the way we handle Async Testing to use a newly created Timed Options

bordoni:
Using WordPress transients can pose weird problems around performance, especially because when someone does not have proper object caching, the Transients are stored using two separate options that WordPress does not autoload.

The most important part to pay attention to is the autoloading for transients using the Options table because they are not autoloaded, nor can they be easily primed all at once. This leads to having two more queries to the DB for every flag and small transients we add.

This PR addresses the first case where this is problematic, which is the feature support for ASYNC requests and adding the actual Timed_Options, which is the solution to the problem described above.

Important to remember that Timed_Options are not a replacement for transient data, as they don't have a good connection to cache expiration, which can lead to weird debugging scenarios.

image


src/Common/Storage/Timed_Option.php
```diff
@@ -0,0 +1,351 @@
...
+
+		if ( null !== $pre ) {
+			return $pre;
+		}
+
+		$time = time();
+
+		// If we have a stored value that is not expired, use it.
+		if (
+			! $force &&
```
jesseeproductions, line 157: check on the placement of the && in the conditions here and other places. In some cases it is at the end of the line and some at the start.

src/Tribe/Feature_Detection.php
```diff
@@ -71,13 +71,15 @@ public function supports_async_process( $force = false ) {
 			return (bool) $supports_async_process;
 		}
 
-		$cached = get_transient( self::$transient );
-
 		$this->lock_option_name = 'tribe_feature_support_check_lock';
+		$transient_name         = Tribe__Process__Tester::TRANSIENT_NAME;
+		$timed_option_exists = tec_timed_option()->exists( $transient_name );
```
jesseeproductions, line 76: Alignment is off.

stratease (approved):

PR #1838, opened 2022-11-21
Avoid fatal on invalid terms

lucatume:
Ticket: n/a

Artifacts: tests

This fixes an issue where the function would assume all the terms
returned from the `get_terms` function would be valid `WP_Term`
instances; due to the filtering that happens in the `get_terms`
filter, that might not be true.
The fix consists in dropping invalid term results from the prime
operation.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 
 = [5.0.4] TBD =
 
-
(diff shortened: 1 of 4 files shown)


lucatume:
Merging this to fix an issue reported on wp.org support forums.

PR #1840, opened 2022-11-22
Fix select2 `clear selection` functionality. [TEC-4393]

EugeneKyale:
Ticket: [TEC-4393](https://theeventscalendar.atlassian.net/browse/TEC-4393)

This PR fixes the issue with select2 where `clear selection` functionality doesn't work when the `data-placeholder` attribute doesn't contain a value. 

Artifact:
https://www.loom.com/share/852beb40d4664a11a978afa30cc01f7a

### DIFF
--- a/vendor/tribe-selectWoo/dist/js/selectWoo.full.js
+++ b/vendor/tribe-selectWoo/dist/js/selectWoo.full.js
@@ -1875,7 +1875,12 @@ S2.define('select2/selection/allowClear',[
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    // Allow clearing when the data-placeholder attribute isn't set. 
+    if ( typeof this.placeholder !== 'undefined' ) {
+      this.$element.val( this.placeholder.id ).trigger( 'change' );
+    } else {
+      this.$element.val( '' ).trigger( 'change' );
+    }
 
     this.trigger('toggle', {});
   };


Camwyn (approved):

vendor/tribe-selectWoo/dist/js/selectWoo.full.js
```diff
@@ -1875,7 +1875,12 @@ S2.define('select2/selection/allowClear',[
       }
     }
 
-    this.$element.val(this.placeholder.id).trigger('change');
+    // Allow clearing when the data-placeholder attribute isn't set. 
+    if ( typeof this.placeholder !== 'undefined' ) {
+      this.$element.val( this.placeholder.id ).trigger( 'change' );
+    } else {
+      this.$element.val( '' ).trigger( 'change' );
+    }
```
Camwyn, line 1883: Don't need to change it but think about if doing it this way makes sense:
```suggestion
    if ( typeof this.placeholder !== 'undefined' ) {
      this.$element.val( this.placeholder.id );
    } else {
      this.$element.val( '' );
    }
    
    this.$element.trigger( 'change' );
```

It reduces code duplication and makes it clear that the `if()` is just changing the value. The trigger happens either way.

EugeneKyale: @Camwyn ooh yeah, this seems more elegant.

PR #1843, opened 2022-11-30
Fix bad pathing creating double-slashes.

Camwyn:
To do so we run `array_filter()` (without a second param - so it just removes empty items) on $path before we run `implode()`. 

When we implode the $path arrays, ensure we first remove all empty items or we wind up with things like `/site/tec/dev/themes/twentytwenty//tribe/events-pro//custom-tables-v1` instead of
`/site/tec/dev/themes/twentytwenty/tribe/events-pro/custom-tables-v1`

Part of [ECP-1400]

:movie_camera: https://d.pr/v/ihcOLc

Another PR coming on the ECP side as well

### DIFF
--- a/src/Tribe/Template.php
+++ b/src/Tribe/Template.php
@@ -416,7 +416,7 @@ class Tribe__Template {
 		$path = array_merge( (array) $this->template_base_path, $this->folder );
 
 		// Implode to avoid Window Problems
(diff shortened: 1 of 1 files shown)


lucatume (changes requested):
Add some test coverage.

bordoni (approved):

PR #1853, opened 2022-12-15
Adding more uopz support

stratease:
[TEC-4548](https://theeventscalendar.atlassian.net/browse/TEC-4548)
[Deadlock PR](https://github.com/the-events-calendar/the-events-calendar/pull/4081) relies on this.

Adds more support for uopz functionality in our test trait.


tests/_support/Traits/With_Uopz.php
```diff
@@ -104,4 +104,22 @@ private function set_class_fn_return( $class, $method, $value, $execute = false
 		uopz_set_return( $class, $method, $value, $execute );
 		$this->uopz_set_returns[] = [ $class, $method ];
 	}
+
+	private function set_class_property( $object, $field, $value ) {
+		if ( ! function_exists( 'uopz_set_property' ) ) {
+			$this->markTestSkipped( 'uopz extension is not installed' );
+		}
+		uopz_set_property( $object, $field, $value );
```
lucatume, line 112: The trait should define a method to run `@after` (see examples in this trait and others) to unset any still applicable property.

tests/_support/Traits/With_Uopz.php
```diff
@@ -104,4 +104,22 @@ private function set_class_fn_return( $class, $method, $value, $execute = false
...
+	}
+
+	private function add_class_fn( $class, $function, $handler ) {
+		if ( ! function_exists( 'uopz_add_function' ) ) {
+			$this->markTestSkipped( 'uopz extension is not installed' );
+		}
+		uopz_add_function(
```
lucatume, line 119: As the comment above, the trait should define an `@after` method to unset at method set for the purpose of testing.

lucatume (changes requested):
Add cleanup.

The trait was originally replicated in TEC to avoid having to depend on Common, and its changes, in tests.

That dependency is already embedded in the plugin logic, so it makes sense to have TEC tests depend on a Common testing facility. Could you look into what effort it would be to remove the `uopz` trait from TEC to use this one?

tests/_support/Traits/With_Uopz.php
stratease: Added some unset/cleanup. Not all API's have a uopz unset, so I got creative.

lucatume (approved):

PR #1855, opened 2022-12-22
[ET-1551] Add contrast methods to Color Utility class

codingmusician:
### Ticket
[ET-1551]

### Description
For Tickets Emails, we wanted a function to determine whether or not to use light text or dark text on top of a certain color of background. We decided to add a couple of `contrast` methods that will help us with this.

[ET-1551]: https://theeventscalendar.atlassian.net/browse/ET-1551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Utils/Color.php
```diff
@@ -513,4 +513,62 @@ private static function _checkHex( $hex ) {
...
+		$dark_ratio = self::get_contrast_ratio( $hex_color, $dark_color );
+		$lite_ratio = self::get_contrast_ratio( $hex_color, $lite_color );
+
+		// Return the highest ratio.
+		if ( $dark_ratio > $lite_ratio ) {
+			return $dark_color;
+		} else {
+			return $lite_color;
+		}
+	}
```
rafsuntaskin, line 539:
```suggestion
		// Return the highest ratio.
		if ( $dark_ratio > $lite_ratio ) {
			return $dark_color;
		}
		return $lite_color;
	}
```

rafsuntaskin:
Also, can we have some tests added?

codingmusician:
> 

Good call, @rafsuntaskin . I'll see if I can create a couple of tests.

src/Tribe/Utils/Color.php
```diff
@@ -513,4 +513,61 @@ private static function _checkHex( $hex ) {
...
 	}
+
+	/**
+	 * Get color that contrasts the most.
+	 * 
+	 * @since TBD
+	 *
+	 * @param string $hex_color  Hex value of color to compare.
+	 * @param string $lite_color Hex value of dark contrast color.
+	 * @param string $dark_color Hex value of light contrast color.
```
juanfra, line 524: Are these right? I'm a bit confused.

src/Tribe/Utils/Color.php
```diff
@@ -513,4 +513,61 @@ private static function _checkHex( $hex ) {
...
+	 * 
+	 * @return string
+	 */
+	public static function get_contrast_color( $hex_color, $lite_color = '#ffffff', $dark_color = '#000000' ) {
```
juanfra, line 528: Maybe we could use `$light_color`? Lite would be the informal way to call it, right?

juanfra (commented):
💪 Looking good. Left a couple of small comments.

src/Tribe/Utils/Color.php
codingmusician: Yes. You pass in two other colors (light and dark) to compare the first color to. It returns the one that contrasts the most.

src/Tribe/Utils/Color.php
codingmusician: I think their interchangeable, but you're right. Light is a little more formal than lite. I used 'lite' for alignment purposes since it was the same number of characters as 'dark'. I'll make the change.

src/Tribe/Utils/Color.php
juanfra: @codingmusician but then `$lite_color` has the description `Hex value of dark contrast color` and `$dark_color` the other way around. Just double checking.

codingmusician: Oh! Whoops. LOL

codingmusician:
@rafsuntaskin I'm going to create the test in a separate PR. I'm having a bit of trouble with it.

bordoni (approved):

PR #1856, opened 2023-01-03
Add notice for PHP 7.4 update requirement.

bordoni:
image


readme.txt
```diff
@@ -1,5 +1,9 @@
 == Changelog ==
 
+= [5.0.7] 2023-01-TBD =
+
+* Tweak - Adding a notice to users using version prior to PHP 7.4 about the incompatibility after February of 2023
```
borkweb, line 5:
```suggestion
* Tweak - Added a dashboard notice for sites running PHP versions lower than 7.4 to alert them that the minimum version of PHP is changing to 7.4 in Feburary 2023.
```

borkweb (dismissed):
Looks great! Just a wording tweak for the changelog.

src/Tribe/Admin/Notice/Php_Version.php
```diff
@@ -92,18 +92,11 @@ public function should_display() {
...
-
-		// for PHP version above 5.4 and up to 5.6
-		if ( version_compare( phpversion(), '5.4.0' ) >= 0 ) {
-			$text = __( 'Starting March 2019, %1$s will no longer support versions prior to PHP 5.6. Your site is currently using PHP version %2$s which will no longer be supported by %1$s. For best results, we recommend using PHP 5.6 or above.', 'tribe-common' );
-		} else {
-			// for PHPversions below 5.4
-			$text = __( 'Starting March 2019, %1$s will no longer work with versions prior to PHP 5.4. Currently your site is using PHP version %2$s. For best results, we recommend using PHP 5.6 or above.', 'tribe-common' );
-		}
+		// PHP versions older than 7.4
+		$text = __( 'Starting February 2023, %1$s will no longer work with versions prior to PHP 7.4. Currently your site is using PHP version %2$s. For best results, we recommend using PHP 7.4 or above.', 'tribe-common' );
```
Camwyn, line 96: Translation comment for placeholders?

Camwyn: Something like 
`/* Translators: %1$s list of plugins, %2$s current PHP version. */`

src/Tribe/Admin/Notice/Php_Version.php
```diff
@@ -8,7 +8,7 @@ public function hook() {
 
 		// display the PHP version notice
 		tribe_notice(
-			'php-deprecated',
+			'php-deprecated-74',
```
Camwyn, line 11: This doesn't need to change, but I'm not sure changing this each time is wise 🤔 
(i.e. are we going to change this to `php-deprecated-82` at some point?)

Camwyn: Also thinking out loud - does it make sense to make the required version a (string) param, then it only needs to get changed in one place 🤔

src/Tribe/Admin/Notice/Php_Version.php
```diff
@@ -92,18 +92,14 @@ public function should_display() {
...
 
-		// for PHP version above 5.4 and up to 5.6
-		if ( version_compare( phpversion(), '5.4.0' ) >= 0 ) {
-			$text = __( 'Starting March 2019, %1$s will no longer support versions prior to PHP 5.6. Your site is currently using PHP version %2$s which will no longer be supported by %1$s. For best results, we recommend using PHP 5.6 or above.', 'tribe-common' );
-		} else {
-			// for PHPversions below 5.4
-			$text = __( 'Starting March 2019, %1$s will no longer work with versions prior to PHP 5.4. Currently your site is using PHP version %2$s. For best results, we recommend using PHP 5.6 or above.', 'tribe-common' );
-		}
+		/* Translators: %1$s read more label */
+		$link = sprintf( '<a href="https://wordpress.org/support/update-php/" target="_blank">%1$s</a>', __( 'Read more.', 'tribe-common' ) );
```
Camwyn, line 99: This could be part of the above (`$text`) string - why separate them?

Camwyn (approved):
Left some thoughts/comments but this is GTG

src/Tribe/Admin/Notice/Php_Version.php
bordoni: This way, we avoid the Problem of people messing with the HTML that we want.

borkweb (approved):

PR #1857, opened 2023-01-09
Allow disabling the logger

lucatume:
Ticket: n/a, emerged as part of Automator work

This updates the Logger logic to allow disabling it by setting the `TEC_DISABLE_LOGGING` constant or environment variable to a truthy value.
In the context of tests, use the environment variable to have it carried in the sub-processes.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,6 +1,6 @@
 == Changelog ==
 
-= [5.0.8] 2023-01-18 =
+= [5.0.8] 2023-01-19 =
 
 * Fix - Correct handling of translated slugs in rewrite context. [TEC-3733]
 * Fix - Handle the case where rewrite rules map to arrays avoiding fatal errors. [TEC-4567]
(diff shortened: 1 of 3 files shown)


jesseeproductions (approved):
This looks good. Gustavo mentioned the same thing to me about not being able to disable the logger cleanly so I think this is a great addition.

Camwyn:
This looks good to me as well - although I wonder if we want to think about: 
a) adding a filter for folks that freak out about logging and/or 
b) making it disabled by default and use the constant to _enable_ it?

lucatume:
I've opened a new PR for this based on the correct branch in #1867

lucatume:
> This looks good to me as well - although I wonder if we want to think about:
> a) adding a filter for folks that freak out about logging and/or
> b) making it disabled by default and use the constant to _enable_ it?

a. good idea, added.
b. I do not think disabling logging should be the default. By the time one needs it, the customer has reached out to support for an issue support would not be able to debug due to missing information.

PR #1864, opened 2023-01-22
Remove safe_mode ini check which is no longer available in PHP 5.4+

sc0ttkclark:
PHP compatibility checks show this message in regards to the safe_mode usage:

> INI directive 'safe_mode' is deprecated since PHP 5.3 and removed since PHP 5.4

This may not be the correct fix, we might want to blow that conditional up in general but this is a good starting point to ensure the right fix is made.


Camwyn:
I created a new PR here: https://github.com/the-events-calendar/tribe-common/pull/1998 with a readme entry

PR #1867, opened 2023-01-24
tweak/allow disabling logger

lucatume:
Reproposal of #1857 on the correct base, see the original PR for comments and review.

This adds a filter-based disable feature of the logger using the `tec_disable_logging` filter.


lucatume:
All tests are passing and I'm merging it after approval on previous incarnation (#1857)

PR #1875, opened 2023-02-02
Fixing cache check logic.

stratease:
This is a [companion to this PR](https://github.com/the-events-calendar/events-pro/pull/2166). In ECP the `ct1_integration` test was failing with an unprimed cache do to a [false positive occurring here](https://github.com/the-events-calendar/events-pro/blob/master/src/Events_Pro/Custom_Tables/V1/Models/Provisional_Post.php#L226-L228).

If doing `isset($cache[$key])` this would ignore cache expiration and would report a valid cache, but then return false when accessing the value from `wp_cache_get`. Now this is expiration sensitive.

Artifact:
Existing test that was failing in ECP (PR above) and new tests in common.


bordoni (approved):

src/Tribe/Cache.php
```diff
@@ -373,16 +373,23 @@ public function make_key( $components, $prefix = '', $sort = true ) {
...
 	 */
 	#[\ReturnTypeWillChange]
 	public function offsetExists( $offset ) {
-		return isset( $this->non_persistent_keys[ $offset ] );
+		if ( ! isset( $this->non_persistent_keys[ $offset ] ) ) {
+			return false;
+		}
+		$value = $this->get( $offset );
+
+		return $value !== false && $value !== null;
```
stratease, line 392: I think we should definitely consider null as not set, e.g.

```php
$cache['some_id'] = null;
// Null values will no longer pass this check.
if(isset($cache['some_id'])) {
```

This update allows for that, and will also consider expiration time (previously ignored).

stratease: Regarding expiration time, before this check a situation like this could and was happening...
```php
$cache->set('id', 123, 1); // Set `id` with `123` for 1 second.

// One second later, in same request....

// This would evaluate to true even though `id` expired.
if(isset($cache['id'])) {
  // The expectation would be that the value is `123`, but it would be `false`.
  // False is the default from wp_cache_get() if the value expired.
  $id = $cache['id']; 
}
```

tests/wpunit/Tribe/CacheTest.php
```diff
@@ -81,27 +75,65 @@ public function it_should_allow_setting_many_different_values_using_array_access
 			'foo8'  => [],
 			'foo9'  => 'false',
 			'foo10' => 'null',
-			'foo11' => false,
-			'foo12' => null,
```
stratease, line 85: `null` and `false` are no longer considered valid values (`false` because of wp_cache_get() and it's definition of false being a not cached return value).

This test was written with the expectation that even though you set `null` and `false` they would be considered valid and set. I think this is problematic for the issues listed above.

This change will adjust our expected behavior for the `Cache` object.

tests/wpunit/Tribe/CacheTest.php
```diff
@@ -81,27 +75,65 @@ public function it_should_allow_setting_many_different_values_using_array_access
...
+	 * @dataProvider should_treat_some_values_as_not_set_data_provider
+	 * @test
+	 */
+	public function should_treat_some_values_as_not_set( string $key, $value ) {
+		$cache = $this->make_instance();
+
+		$cache[ $key ] = $value;
+
+		// Invalid values will not be considered set.
+		$this->assertFalse( isset( $cache[ $key ] ) );
```
stratease, line 130: Now we are validating "invalid" values as not being set, regardless of expiration.

Camwyn:
When this gets merged we need to update G23.beans as well - the two branches need to stay in sync

src/Tribe/Cache.php
Camwyn:
```suggestion
		return ! empty( $value );
```
Covers null, false, and empty strings (which should probably also be invalid?)

stratease: I did consider that, but I wouldn't go that loose. I don't even like considering `false` invalid, I'm only doing so because that's how WP treats their cache getter. I'd lean toward only considering `null` and `false` invalid, because we may legitimately cache various results that are legit but `empty()`. It is very dependent on the specific case.

stratease:
> When this gets merged we need to update G23.beans as well - the two branches need to stay in sync

Merge bb8 into beans?

src/Tribe/Cache.php
stratease: To go on a rant on this... if time/agreement permitted, this is how I'd lean.

I'd consider `false` a legit cache value.
```php
$cache['is_turned_on'] = false; // A bool flag that was set to "off" - legit use case.

if(isset($cache['is_turned_on'])) { // Would be true, because false is a legit value.
```
And internally I'd do some serialization to work around Wordpresses `false` interpretation as an indicator for "cache expired/doesn't exist" from the `wp_cache_get()` func we use. 

e.g. a simplified `Cache` set method to handle bool `false`.
```php
// Cache class
public function set($id, $value) {
  if($value === false) {
    $value = "some_serialized_id";
  }
  wp_cache_set($id, $value);
}
```
Then if we fetch it...
```php
// Cache class
public function get($id) {
  $val = wp_cache_get($id, $value);

  if($val === 'some_serialized_id') {
    return false; // This was the serialized value.
  }

  if($val === false) {
    return null; // This is what should be returned if no cache value exists - null is a standard for "something doesn't exist".
  }

  // All other legit values returned.
  return $val;
}
```

Then `null` would be the only not "legit" value, and I think most people would safely assume that.

bordoni: For me, the biggest question is if we should keep this for the next release so we can run a bunch of QA on that since Cache is used all over the codebase.

lucatume: The `wp_cache_get` function takes a fourth argument `$found` that is set by reference that allows decoupling what is returned from the cache from its found state. Why not leverage that? Any object cache implementation will set and implement that.

lucatume (changes requested):
I think leveraging the `$found` argument of the `wp_cache_get` function could provide a less opinionated take on what is considered a "set" and "unset" cached value.

src/Tribe/Cache.php
stratease: I like this - didn't see that. Thanks @lucatume .

stratease: Updated with this line of reasoning - now `false` is a valid value. Do we want to do this? This will require fixing any number of places we looked for `false`.

Also switched the `$default` to `null` in the getter, if no value specified the default will not return a `false` bool. This is to be consistent with `null` being the check for a cache value (where `false` is valid now and would persist).

tests/wpunit/Tribe/CacheTest.php
stratease: The `DateUtils_Test` is failing because it was checking for `false` in the cache. Just one example that would need tweaking if we go this route...

.travis.yml
```diff
@@ -42,9 +42,6 @@ env:
     - WP_ADMIN_PASSWORD="password"
 
 before_install:
-  # Rollback Composer to version 1.
```
lucatume, line 45: Pretty sure this needs update from the base branch.

src/Tribe/Cache.php
lucatume: > This will require fixing any number of places we looked for `false`.

I would not change this behaviour of the `Tribe__Cache` API now: `false` might be a valid `wp_cache_get` value, but the `Tribe__Cache` class will treat it as non valid.

tests/wpunit/Tribe/CacheTest.php
lucatume: See my reply above, I would keep this limit of the `Tribe__Cache` implementation for back-compatibility purposes.

lucatume (changes requested):
The logic is fine, the branch will need update from the base branch.

src/Tribe/Cache.php
stratease: Then using `$found` is pointless, since the only point was to differentiate `false` stored value vs `false` no stored value.

I'll go back to what I had.

lucatume: The idea was to keep a smart logic that could be dumbed down and "hidden" in the cache API. Now the change is reverted it's fine to leave it like that.

lucatume (approved):

Camwyn (approved):

PR #1877, opened 2023-02-07
Safeguard around rewrite_filter_array hook params.

stratease:
[TEC-4679](https://theeventscalendar.atlassian.net/browse/TEC-4679)

Adds a safeguard around the `rewrite_filter_array` hook callback, which expects an array. Now will gracefully return non-array values and avoids fatal errors from improper use of the filter.

Artifacts:
Screenshots in ticket.
```
PHP Fatal error: Uncaught TypeError: Argument 1 passed to Tribe__Rewrite::remove_percent_placeholders() must be of the type array, bool given, called in /home/regeneravida/public_html/wp-includes/class-wp-hook.php on line 308 and defined in /home/regeneravida/public_html/wp-content/plugins/the-events-calendar/common/src/Tribe/Rewrite.php:177
```

[TEC-4679]: https://theeventscalendar.atlassian.net/browse/TEC-4679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Rewrite.php
+++ b/src/Tribe/Rewrite.php
@@ -193,18 +193,35 @@ class Tribe__Rewrite {
 		return $rules;
 	}
 
+	/**
+	 * Filter for the `rewrite_rules_array` hook.
+	 *
+	 * @since TBD
+	 *
+	 * @param array|mixed $rules The rules to be filtered.
+	 *
+	 * @return array|mixed Rules after filtering.
+	 */
+	public function filter_rewrite_rules_array( $rules ) {
+		if ( ! is_array( $rules ) ) {
+			return $rules;
+		}
+
+		return $this->remove_percent_placeholders( $rules );
+	}
+
 	protected function add_hooks() {
 		add_filter( 'generate_rewrite_rules', [ $this, 'filter_generate' ] );
 
 		// Remove percent Placeholders on all items
-		add_filter( 'rewrite_rules_array', [ $this, 'remove_percent_placeholders' ], 25 );
+		add_filter( 'rewrite_rules_array', [ $this, 'filter_rewrite_rules_array' ], 25 );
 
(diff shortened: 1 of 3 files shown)


bordoni (approved):

src/Tribe/Rewrite.php
```diff
@@ -193,18 +193,35 @@ public function remove_percent_placeholders( array $rules ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param array|mixed $rules The rules to be filtered.
+	 *
+	 * @return array|mixed Rules after filtering.
+	 */
+	public function filter_rewrite_rules_array( $rules ) {
+		if ( ! is_array( $rules ) ) {
+			return $rules;
```
Camwyn, line 207: Is there a good reason to not simply cast it to an array?

Camwyn: It's supposed to be one: https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-includes/class-wp-rewrite.php#L190

stratease: Even if we cast it, it will be malformed. I don't see a scenario where casting it will solve anything other than pushing the problem farther downstream?

i.e. 
```
$rules = 'bob';
return (array)$rules; // ['bob'] is not a legit structure (right?)
```

jesseeproductions (approved):

src/Tribe/Rewrite.php
Camwyn: Fair enough.

stratease: Haha, ja man!

PR #1878, opened 2023-02-09
feat(Context) add the is_editing_posts_list method

lucatume:
Ticket: [TEC-4542](https://theeventscalendar.atlassian.net/browse/TEC-4542)

Add the `Tribe__Context::is_editing_posts_list( $post_type): bool`
method to the context to allow unified detection of an admin posts list
edit screen for specified post types.

Artifacts: tests

[TEC-4542]: https://theeventscalendar.atlassian.net/browse/TEC-4542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Context.php
+++ b/src/Tribe/Context.php
@@ -1656,4 +1656,30 @@ class Tribe__Context {
 			$this->request_cache[ $key ] = $val;
 		}
 	}
+
+	/**
+	 * Whether the current request is one to edit a list of the specified post types or not.
+	 *
+	 * The admin edit screen for a post type is the one that lists all the posts of that typ,
+	 * it has the URL `/wp-admin/edit.php?post_type=<post_type>`.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|array<string> $post_type The post type or post types to check.
+	 *
+	 * @return bool Whether the current request is one to edit a list of the specified post types or not.
+	 */
+	public function is_editing_posts_list( $post_type ): bool {
+		// Quick check: are we on the `/wp-admin/edit.php` page?
+		global $pagenow;
+
+		if ( $pagenow !== 'edit.php' ) {
+			return false;
+		}
+
+		// Run some more thorough checks for the post type(s).
+		$post_types = array_filter( (array) $post_type );
+
+		return $this->is_editing_post( $post_types );
+	}
 }
--- a/readme.txt
+++ b/readme.txt
@@ -1,5 +1,9 @@
 == Changelog ==
 
+= [TBD] TBD =
+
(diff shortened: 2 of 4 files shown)


src/Tribe/Context.php
```diff
@@ -1656,4 +1656,30 @@ public function safe_set( $values_or_key, $value = null ) {
...
+	 * The admin edit screen for a post type is the one that lists all the posts of that typ,
+	 * it has the URL `/wp-admin/edit.php?post_type=<post_type>`.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|array<string> $post_type The post type or post types to check.
+	 *
+	 * @return bool Whether the current request is one to edit a list of the specified post types or not.
+	 */
+	public function is_editing_posts_list( $post_type ): bool {
```
stratease, line 1672: In my mind this is more of a generic utility - `Context` is starting to feel like an unconstrained bucket of utility/state management. Am I wrong on this? 

Does it make sense to move this utility into a more singly focused utility, like a `Page_Util` (don't like it but there's an example).

stratease (changes requested):

src/Tribe/Context.php
lucatume: I'm answering in particular for this method first: the `is_editing_post` method is already part of the Context class, it seemed to me the `is_editing_posts_list` method would fit here as well.

In general: the idea behind context is to provide an API over checks that would be done against globals and HTTP super-globals. In that regard, this method follows the path by keeping in line with an information you would pull "from the context".

With a relatively small lift, the Context class could be refactored to a façade/proxy for other, more specialized classes. The pre-Context state of the code showed that developers, both in the TEC team and third-party, would not find those specialized classes and would re-implement yet another version of that "look what the context is" logic.

So: a large Context class has the advantage of being the one, easy-to-find and guess, reference for context logic; comes with the disadvantage of being large.
A Context class broken into smaller classes is cleaner, but comes with a discoverability cost.

A façade seems to be a good middle-ground.

stratease: Gotcha - all of that makes sense.

I do like the idea of a facade at the very least. 

Although not sure you were recommending this but as long as there is some cut-off of things inside a particular facade. As you mentioned the intent for Context is inferring details from globals/http requests (although it does seem to do more).

Might be something worth adding to our "dev todo" project list @bordoni

stratease (approved):

bordoni (approved):

PR #1883, opened 2023-02-18
Site Health Info framework

bordoni:


src/Common/Site_Health/Info_Section_Abstract.php
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	/**
+	 * Sorts the fields stored on this section, will retain keys.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function sort_fields(): void {
```
Camwyn, line 220: Be more explicit - sort by what? Where?
```suggestion
	/**
	 * Sorts the fields stored on this section in-place by priority. Will retain keys.
	 *
	 * @since TBD
	 *
	 * @return void
	 */
	protected function sort_fields_by_priority(): void {
```

Camwyn:
I think this is redundant/outdated now. correct me if I'm wrong @bordoni

PR #1898, opened 2023-03-17
Replace FILTER_SANITIZE_STRING

EugeneKyale:


lucatume (changes requested):
Rename the function to `tec_sanitize_string`.
Add tests dedicated to the `tec_sanitize_string` function directly, to ensure there is direct coverage on it, not just indirect one.

lucatume (approved):

PR #1902, opened 2023-03-28
Feature - Recurring Events + Tickets

lucatume:


src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register() {
+		/*
+		 * Look up and set the value in the container request cache to allow building the same Controller
+		 * with a **different** container. (e.g. in tests).
+		 */
+		if ( $this->container->getVar( static::class . '_registered' ) ) {
```
Camwyn, line 38: Why can't we just use `isBound()` or `offsetExists()` here?

Camwyn (changes requested):
Small tweak and a question

Camwyn:
@lucatume I updated the PR to be against master - as daikon has been merged.
Haven't dealt with the conflicts yet.

PR #1908, opened 2023-04-09
Update utils.php

oheinrich:
In line 1130 the function tribe_sanitize_deep( &$value ) is using the constant FILTER_SANITIZE_STRING.

This constant FILTER_SANITIZE_STRING is deprecated since php 8.1: https://stackoverflow.com/questions/69207368/constant-filter-sanitize-string-is-deprecated

I changed it to this:
$value = htmlspecialchars( $value, ENT_QUOTES, 'UTF-8' );


Camwyn:
The release coming out today will address all of this (there are a few more as well), but thank you for the effort!

PR #1909, opened 2023-04-10
Fix edge case bug, with multiple actions being cached and not expirin…

stratease:
[ECP-1475](https://theeventscalendar.atlassian.net/browse/ECP-1475)

Fixes a cache bug that was causing the post object to be cached and carry stale data, and any following updates to be made with no longer valid start dates etc. This was due to our memoized action timestamps, which have now been removed in lieu of Wordpress' `get_option` cache.

Artifacts:
First failure in video demonstrates the bug (second failure is existing bug for separate ticket)
https://drive.google.com/file/d/1EP9k51dU1EuVLQ6MCBGWMENiNPQGjUVo/view

[ECP-1475]: https://theeventscalendar.atlassian.net/browse/ECP-1475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Cache.php
+++ b/src/Tribe/Cache.php
@@ -289,6 +289,7 @@ class Tribe__Cache implements ArrayAccess {
 	 * Returns the time of an action last occurrence.
 	 *
 	 * @since 4.9.14 Changed the return value type from `int` to `float`.
+	 * @since TBD No longer memoizes the first triggered timestamp.
 	 *
 	 * @param string $action The action to return the time for.
 	 *
--- a/src/Tribe/Cache.php
+++ b/src/Tribe/Cache.php
@@ -295,14 +295,6 @@ class Tribe__Cache implements ArrayAccess {
 	 * @return float The time (microtime) an action last occurred, or the current microtime if it never occurred.
 	 */
 	public function get_last_occurrence( $action ) {
-		static $cache_var_name = __METHOD__;
-
-		$cache_last_actions = tribe_get_var( $cache_var_name, [] );
-
-		if ( isset( $cache_last_actions[ $action ] ) ) {
-			return $cache_last_actions[ $action ];
-		}
-
 		$last_action = (float) get_option( 'tribe_last_' . $action, null );
 
 		if ( ! $last_action ) {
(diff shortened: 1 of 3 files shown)


src/Tribe/Cache.php
```diff
@@ -289,32 +289,21 @@ public function get_id( $key, $expiration_trigger = '' ) {
...
 	public function get_last_occurrence( $action ) {
-		static $cache_var_name = __METHOD__;
-
-		$cache_last_actions = tribe_get_var( $cache_var_name, [] );
-
-		if ( isset( $cache_last_actions[ $action ] ) ) {
-			return $cache_last_actions[ $action ];
-		}
-
 		$last_action = (float) get_option( 'tribe_last_' . $action, null );
```
bordoni, line 299: We need to make sure if the result of this, when `get_option( KEY, '__UNIQUE_DEFAULT_KEY__' ) === '__UNIQUE_DEFAULT_KEY__'` we do a `update_option( KEY, null, true );` so that the value is auto-loaded next time.

bordoni (changes requested):

src/Tribe/Cache.php
stratease: So we want it to save/return `null` if not saved yet?

stratease: To clarify on my question above - right now it returns the `microtime()` value. If I understand you correctly, you are asking to change it to return `null` if it's not set instead of the current `microtime()` ? That is a change in logic (which is why I left things as they are).

And regarding the auto-load - the logic as it is should already be doing that.

bordoni: No no. What i am trying to suggest is that when you have no value in the DB, WordPress wont preload any value thus creating a database request to look for the value, what I am asking you to make sure that once we verify that the value is not existent just fill with something that will be preloaded next time, as a way to avoid this query being present every time in case the option doesn't exist.

Basically preventing a bunch of calls to `tribe_last_{$action}` to exist more than in one request. Afterwards there will always be a preloaded value even if `null`.

stratease: Isn't that what this is doing already?

https://github.com/the-events-calendar/tribe-common/pull/1909/files#diff-03cdddb06b33fcad3e0a4ae404379bf6fcdb2eb19c5cfbc1444c55001d2ddec5L308-L310

bordoni (approved):

PR #1911, opened 2023-04-19
feat(src/Provider) add the base Controller class and test case

lucatume:
Ticket, for context: [ECP-1465](https://theeventscalendar.atlassian.net/browse/ECP-1465).

Add the base Controller class and the dedicated test case.
See [ECP PR](https://github.com/the-events-calendar/events-pro/pull/2247) for an example use of both.

Note that this code does not include the modifications done to the `tribe_container` in the `feature/recurring-events-plus-tickets` branch but lays the ground to use it and leverages what can be leveraged immediately.

[ECP-1465]: https://theeventscalendar.atlassian.net/browse/ECP-1465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,135 @@
+<?php
+/**
+ * The base clas all Controllers should extend.
```
Camwyn, line 3:
```suggestion
 * The base class all Controllers should extend.
```

src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,135 @@
...
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register() {
+		/*
+		 * Look up and set the value in the container request cache to allow building the same Controller
+		 * with a **different** container. (e.g. in tests).
+		 */
+		if ( $this->container->getVar( static::class . '_registered' ) ) {
```
Camwyn, line 38: Should we set '_registered' as a prop? Maybe one that is not editable for sanity's sake?

src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,135 @@
...
+namespace TEC\Common\Provider;
+
+use tad_DI52_ServiceProvider as Service_Provider;
+use TEC\Common\StellarWP\ContainerContract\ContainerInterface;
+use Tribe__Log as Log;
+
+/**
+ * Class Controller.
+ *
+ * @since   TBD
```
Camwyn, line 19:
```suggestion
 * @since TBD
```

tests/_support/Provider/Controller_Test_Case.php
```diff
@@ -0,0 +1,212 @@
...
+
+use Codeception\TestCase\WPTestCase;
+use TEC\Common\Provider\Controller;
+use Tribe\Tests\Traits\With_Uopz;
+use Tribe__Container as Container;
+
+/**
+ * Class Controller_Test_Case.
+ *
+ * @since   TBD
```
Camwyn, line 16:
```suggestion
 * @since TBD
```

src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,135 @@
+<?php
+/**
+ * The base clas all Controllers should extend.
+ *
+ * @since   TBD
```
Camwyn, line 5:
```suggestion
 * @since TBD
```

Camwyn: extra spaces make these harder to search/replace 😉

Camwyn (changes requested):

src/Common/Provider/Controller.php
stratease: Having it in our global vars can be useful to check outside, as we sometimes do. And I'd have one source of truth.

src/Common/Provider/Controller.php
```diff
@@ -0,0 +1,135 @@
...
+	 * Whether the controller is active or not.
+	 *
+	 * Controllers will be active by default, if that is not the case, the controller should override this method.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether the controller is active or not.
+	 */
+	public function is_active(): bool {
+		return true;
```
stratease, line 85: Is this a different use case than `$this->container->getVar( static::class . '_registered' )` ? Seems like this would be checking if it is registered or not...?

src/Common/Provider/Controller.php
stratease: Although I prefer something like `tribe(Some_Controller::class)->is_active()` which references that var...

src/Common/Provider/Controller.php
Camwyn: This is a placeholder for children to define what constitutes "active"

src/Common/Provider/Controller.php
lucatume: > Although I prefer something like tribe(Some_Controller::class)->is_active()
That is already part of the API, but it's about the Controller being "active" as in "relevant for the current request"; e.g. a wp-cli controller is not active when the request context is not the a wp-cli one. I guess you mean a method like `Controller::is_registered`, which is currently lacking from the API. I've added it. 
Note: I've updated the `Controller_Test_Case` to replace the global service locator (`tribe()`) with the test one when using `Controller_Test_Case::make_controller` to make sure the `Controller::is_registered` method is correct spoofed.

src/Common/Provider/Controller.php
lucatume: See my answer to the other questions and the difference between `is_active` and `is_registered`.

lucatume:
Updated the code to add the `Controller::is_registered` method, applied review suggestions.

stratease (approved):

PR #1912, opened 2023-04-20
Fix for fatal.

stratease:
This adds a Strauss namespaced definition for Firebase/JWT.

### DIFF
--- a/src/Tribe/Promoter/Connector.php
+++ b/src/Tribe/Promoter/Connector.php
@@ -54,7 +54,7 @@ class Tribe__Promoter__Connector {
 			'userId'       => $user_id,
 		];
 
-		$token = \Firebase\JWT\JWT::encode( $payload, $promoter_key, 'HS256' );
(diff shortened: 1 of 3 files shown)


src/Tribe/Promoter/Connector.php
```diff
@@ -54,7 +54,7 @@ public function authorize_with_connector( $user_id, $secret_key, $promoter_key,
 			'userId'       => $user_id,
 		];
 
-		$token = \Firebase\JWT\JWT::encode( $payload, $promoter_key, 'HS256' );
+		$token = \TEC\Common\Firebase\JWT\JWT::encode( $payload, $promoter_key, 'HS256' );
```
jesseeproductions, line 57: Looks good, a suggestion to do use TEC\Common\Firebase\JWT\JWT as TEC_JWT and then use that here and below.

bordoni (approved):

jesseeproductions (approved):

PR #1915, opened 2023-04-27
Fallback for JWT classes.

stratease:
Fallback for Strauss renaming of the JWT.

ET and EVA both depend on JWT from common.


Camwyn (approved):

bordoni:
Good stuff.

PR #1916, opened 2023-04-28
Passing null to `strpos()` is deprecated in PHP 8.1

Camwyn:
See: https://lw.slack.com/archives/C01STUW82P3/p1682112698672179

### DIFF
--- a/src/Tribe/Repository.php
+++ b/src/Tribe/Repository.php
@@ -2126,7 +2126,7 @@ abstract class Tribe__Repository
 				$args['meta_query'][ $array_key ]['value'] = $meta_value;
 			}
 
-			if ( 0 === strpos( $type_or_format, '%' ) ) {
+			if ( ! empty( $type_or_format ) && 0 === strpos( $type_or_format, '%' ) ) {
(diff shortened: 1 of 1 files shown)


bordoni (dismissed):

src/Tribe/Repository.php
```diff
@@ -2126,7 +2126,7 @@ protected function build_meta_query( $meta_key, $meta_value = 'value', $compare
 				$args['meta_query'][ $array_key ]['value'] = $meta_value;
 			}
 
-			if ( 0 === strpos( $type_or_format, '%' ) ) {
+			if ( ! empty( $type_or_format ) && 0 === strpos( $type_or_format, '%' ) ) {
```
stratease, line 2129:
```suggestion
			if ( is_string( $type_or_format ) && 0 === strpos( $type_or_format, '%' ) ) {
```
Probably better to check for string instead of the various types that would pass the empty check.

stratease (changes requested):

src/Tribe/Repository.php
Camwyn: Good point. I was thinking it would not pass `0` or an empty string to the next check (strpos) since that will fail the check anyway?

PR #1923, opened 2023-05-25
Fix/ecp 1505 speed memoize

stratease:
[ECP-1505](https://theeventscalendar.atlassian.net/browse/ECP-1505)

Adding a [system / environment configuration utility](https://github.com/the-events-calendar/tribe-common/pull/1923/files#diff-cf03646ad083f81f8ec80bbdd775d8ac45c75c7bc1bf302f6fb06dfa34a1dc64).

Being used in https://github.com/the-events-calendar/the-events-calendar/pull/4275

[ECP-1505]: https://theeventscalendar.atlassian.net/browse/ECP-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Repository/Query_Filters.php
```diff
@@ -616,16 +616,6 @@ protected function where_field_is( $where, WP_Query $query, $field, $prepare = '
 		return $where;
 	}
 
-	/**
```
stratease, line 619: Just a note - this is being removed because during shutdown it slows down due to garbage collection(?), and it is not likely necessary (don't think destruct would run til shutdown anyways, there are object references in the filter list...).

readme.txt
```diff
@@ -3,6 +3,7 @@
 = [TBD] TBD =
 
 * Release - B23.Goldorak
+* Fix - Adding Configuration feature, to enable simple feature flag and other checks, with less boilerplate. See [readme](https://github.com/the-events-calendar/tribe-common/pull/1923/files#diff-cf03646ad083f81f8ec80bbdd775d8ac45c75c7bc1bf302f6fb06dfa34a1dc64) for more details. [ECP-1505]
```
bordoni, line 6: Lets split this into two Changelogs.

bordoni (changes requested):
I like this, can you present this on an architecture call? Just to make sure more people are aware. But I feel like this is a good move. Given the changelog change we should be able to merge.

readme.txt
stratease: @bordoni not sure I follow - what do you mean two changelogs? Can you give me an example of what you mean?

PR #1925, opened 2023-05-26
use an array for paths so we can be more deterministic

Camwyn:
To solve the issue of ET always wanting to go first, and filters not being choosy


src/Tribe/Main.php
```diff
@@ -783,7 +783,17 @@ public function get_parent_plugin_file_path(): string {
 		 * @var string $parent_plugin_file The current path to the parent plugin file.
 		 *
 		 */
-		return apply_filters( 'tec_common_parent_plugin_file', $this->parent_plugin_file );
+		$paths = apply_filters( 'tec_common_parent_plugin_file', [] );
```
stratease, line 786: Any test coverage for this filter/method?

stratease (changes requested):

bordoni (approved):

PR #1946, opened 2023-06-28
Resolve telemetry path issue on Windows

borkweb:
When we look at paths during the bootstrapping of Telemetry, we need to be sure to normalize paths so that the directory separator matches what WordPress expects.

:ticket: [TEC-4842]

Artifact - saving on Windows fatals every time until this fix is in place, then we get:
Screenshot 2023-06-28 171300

Props to @Camwyn for the debug session.

[TEC-4842]: https://theeventscalendar.atlassian.net/browse/TEC-4842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 
 = [5.1.2] TBD =
 
-* Fix - Properly handle plugin paths on windows during telemetry booting. [TEC-4842]
+* Fix - Properly handle plugin paths on Windows during telemetry booting. [TEC-4842]
 
 = [5.1.2.2] 2023-06-23 =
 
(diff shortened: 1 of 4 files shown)


readme.txt
```diff
@@ -2,7 +2,7 @@
 
 = [5.1.2] TBD =
 
-* Release - B23.Goldorak
+* Fix - Properly handle plugin paths on windows during telemetry booting. [TEC-4842]
```
Camwyn, line 5:
```suggestion
* Fix - Properly handle plugin paths on Windows during telemetry booting. [TEC-4842]
```

Camwyn (approved):

PR #1950, opened 2023-07-03
2023 Stellar Sale Banner

EugeneKyale:
Ticket : [TEC-4841](https://theeventscalendar.atlassian.net/browse/TEC-4841)

Adds the 2023 Stellar Sale Banner.

**Screenshot 📸**

image

[TEC-4841]: https://theeventscalendar.atlassian.net/browse/TEC-4841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Dependency.php
```diff
@@ -548,6 +548,28 @@ protected function check_addon_dependencies( $main_class ) {
...
+
+		/**
+		 * Return if we have a premium plugin active.
+		 * Results may be unreliable if called before the tribe_plugins_loaded hook.
+		 *
+		 * @since TBD
+		 *
+		 * @return bool
+		 */
+		public function has_active_premium_plugin(): bool {
```
stratease, line 560:
```suggestion
		public function has_active_not_premium_plugin(): bool {
```

Not premium?

stratease (changes requested):

src/Tribe/Dependency.php
stratease: The naming is hard here - it checks both ET and TEC... `base` plugins?

EugeneKyale: I think we should retain the name since the method only returns true in the event that there is at least one active plugin that isn't TEC or ET.

stratease (approved):

PR #1951, opened 2023-07-05
feat(Container) add register_on_action method

lucatume:
Add a `Container::register_on_action` method to allow this kind of code
in a Service Provider:

```php
$this->container->register_on_action( 'some_action', Some_Provider::class);
```

The `Some_Provider` service provider will be registered:
* ... if the `some_action` action was already dispatched
* ... if, and when, the `some_action` action is dispatched

Furthermore, the `TEC\Common\Contracts\Container::register` method will
fire a generic action when a Service Provider is registered and a custom
one each Service Provider can define with the `static string
$registration_action` property.

Put together this allows chaining the activation of Service Providers to
one another without relying on `class_exists` and other heuristic
lookups:

```php

class Base_Provider extends Service_Provider {
  protected static string $registration_action = 'base_provider_registered';

  public function register() {
    $this->container->setVar( 'secret', '12345' );
  }
}

class Extension_Provider extends Service_Provider {
  public function register() {
    $this->container->singleton( Api::class, fn() => new Api( $this->container->getVar( 'secret' ) );
  }
}

// Register the extension provider only if the base provider was registered.

$this->container->register_on_action( 'base_provider_registered', Extension_Provider::class );
```


src/Common/Contracts/Container.php
```diff
@@ -27,4 +27,79 @@ public function get( $id ) {
...
+	 * @param string $class The service provider class name.
+	 * @param string ...$alias Optional. The alias(es) to register the service provider with.
+	 *
+	 * @return void The Service Provider is registered immediately if the action has already fired, or when
+	 *              the action fires.
+	 *
+	 * @throws \TEC\Common\lucatume\DI52\ContainerException If the provider class is marked as deferred but
+	 *                                                      does not provide a set of deferred registrations.
+	 */
+	public function register_on_action( string $action, string $class, string ...$alias ): void {
```
stratease, line 89: Nice!

stratease (dismissed):

src/Common/Contracts/Container.php
```diff
@@ -27,4 +27,79 @@ public function get( $id ) {
...
+	 * Registers a service provider on a given action is dispatched.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $action The action to register the provider on.
+	 * @param string $class The service provider class name.
+	 * @param string ...$alias Optional. The alias(es) to register the service provider with.
+	 *
+	 * @return void The Service Provider is registered immediately if the action has already fired, or when
+	 *              the action fires.
```
Camwyn, line 84:
```suggestion
	 * @return void The Service Provider is registered when the action fires,
	 *               or immediately if the action has already fired.
```

src/Common/Contracts/Container.php
```diff
@@ -27,4 +27,79 @@ public function get( $id ) {
...
+		parent::register( $serviceProviderClass, ...$alias );
+
+		/**
+		 * Fires when a service provider is registered by the container.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $serviceProviderClass The service provider class name.
+		 */
+		do_action( 'tec_container_registered_provider', $serviceProviderClass );
```
Camwyn, line 55: Can we add a
`do_action( 'tec_container_registered_provider_' . $serviceProviderClass );`

To allow for specific targeting w/o having to check every one of these?

Maybe as part of the failed check for `$serviceProviderClass::$registration_action`?

Camwyn: PS: do we want to pass/hook alias(es) as well? 🤔

Camwyn (commented):
Immediate thoughts and comments

src/Common/Contracts/Container.php
lucatume: Good idea!
I've added a class-specific action, always dispatched whether the custom action is dispatched or not, and included the aliases in the dispatched actions.

src/Common/Contracts/Container.php
```diff
@@ -27,4 +27,89 @@ public function get( $id ) {
...
+		) {
+			return;
+		}
+
+		/**
+		 * Fires a custom action defined by the Service Provider when it's registered.
+		 *
+		 * @since TBD
+		 */
+		do_action( $serviceProviderClass::$registration_action, $serviceProviderClass, $alias );
```
codingmusician, line 81: Quick question: Will there be any situation where `$serviceProviderClass::$registration_action` might not exist or do all service providers have it defined?

Camwyn (approved):
Excited to get to use this 😄

PR #1960, opened 2023-07-24
[ET-1820] Use `wp_date` for date formats to avoid translation issues with Month names.

rafsuntaskin:
[ET-1820]

**Before:**
Screenshot 2023-08-02 at 4 18 28 PM

**After:**
Screenshot 2023-08-02 at 4 18 44 PM

[ET-1820]: https://theeventscalendar.atlassian.net/browse/ET-1820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,6 +2,7 @@
 
 = [5.1.4] TBD =
 
+* Fix  - Make use of `wp_date` to format dates and avoid translation issues with translating month names in other languages. [ET-1820]
 * Feat - Fire an action on Service Provider registration; register Service Providers on action with `Container::register_on_action`.
 
 = [5.1.3] 2023-07-13 =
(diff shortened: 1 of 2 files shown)


codingmusician (approved):

bordoni:
We need to revert this before the release. This method `reformat` has some usages in our codebase where the actual reformatted date is saved in the Database it's not just for display purposes.

That particular fix is good if applied only for visual reformats, but currently that method is too low level. Lets revert this tomorrow.

The actual fix should be on ET or ET+ where this is used for the display. `Tribe__Date_Utils::build_date_object( $previous_date )->format_i18n( FORMAT_HERE );`

PR #1970, opened 2023-08-29
Declare all class properties for PHP 8.1 compatibility.

Camwyn:
Also fixes some incorrect docblocks - specifically around parameter and return types.

Possibly too late for Jenkins.

No ticket, no QA needed. (activation tests should cover anything wrong here)


stratease (approved):
Awesome!

PR #1981, opened 2023-09-21
Do not opt in the subscriber repeatedly

Camwyn:
It results in spamming the server with requests.

[TEC-4920]

[TEC-4920]: https://theeventscalendar.atlassian.net/browse/TEC-4920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Telemetry/Telemetry.php
+++ b/src/Common/Telemetry/Telemetry.php
@@ -493,19 +493,15 @@ final class Telemetry {
 		}
 
 		// Check for cached slugs.
-		$cached_slugs = tribe( 'cache' )->get( 'tec_telemetry_slugs' );
+		$cached_slugs = tribe( 'cache' )['tec_telemetry_slugs'] ?? null;
 
-		// We have already run and the slug list hasn';t changed since then.
+		// We have already run and the slug list hasn't changed since then.
 		if ( is_null( $opted ) && ! empty( $cached_slugs ) && $cached_slugs == $tec_slugs  ) {
 			return;
 		}
 
 		// No cached slugs, or the list has changed, or we're running manually - so (re)set the cached value.
-		tribe( 'cache' )->set(
-			'tec_telemetry_slugs',
-			$tec_slugs,
-			Tribe__Cache::NON_PERSISTENT
-		);
+		tribe( 'cache' )['tec_telemetry_slugs'] = $tec_slugs;
 
 		// In case we're not specifically passed a status...
 		$new_opted = $this->calculate_optin_status( $opted );
--- a/src/Common/Telemetry/Telemetry.php
+++ b/src/Common/Telemetry/Telemetry.php
@@ -13,7 +13,6 @@ use TEC\Common\StellarWP\Telemetry\Config;
 use TEC\Common\StellarWP\Telemetry\Opt_In\Status;
 use Tribe__Container as Container;
 use TEC\Common\StellarWP\Telemetry\Opt_In\Opt_In_Template;
-use Tribe__Cache;
 
 /**
  * Class Telemetry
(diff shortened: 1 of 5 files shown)


src/Common/Telemetry/Telemetry.php
```diff
@@ -123,7 +130,7 @@ public function boot(): void {
 			return;
 		}
 
-		$telemetry_server   = ! defined('TELEMETRY_SERVER') ? 'https://telemetry.stellarwp.com/api/v1': TELEMETRY_SERVER;
+		$telemetry_server   = ! defined('STELLARWP_TELEMETRY_SERVER') ? 'https://telemetry.stellarwp.com/api/v1': STELLARWP_TELEMETRY_SERVER;
```
Camwyn, line 133: This is the cross-product constant we agreed upon. Was going to update it in k9 anyway.

src/Common/Telemetry/Telemetry.php
```diff
@@ -462,13 +467,12 @@ public static function get_tec_telemetry_slugs() {
 	 * @return void
 	 */
 	public function register_tec_telemetry_plugins( $opted = NULL ) {
-		$new_opted = $opted;
 		// Let's reduce the amount this triggers.
-		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
```
Camwyn, line 467: This got moved to `boot()`

borkweb: Because the add-ons call `register_tec_telemetry_plugins()` (for now), I think we should keep this here.

src/Common/Telemetry/Telemetry.php
```diff
@@ -462,13 +467,12 @@ public static function get_tec_telemetry_slugs() {
 	 * @return void
 	 */
 	public function register_tec_telemetry_plugins( $opted = NULL ) {
```
borkweb, line 469: We need the addition that I mentioned in my message in Slack: 

> register_tec_telemetry_plugins() is being called a bunch. It doesn't need to be. Let's make sure that is called only once.

Let's add a `Tribe__Cache::get()` and `Tribe__Cache::set()` (non-persistent) so we ensure this method is called only once. We could ignore the prevention if manually setting an opt-in.

borkweb: We additionally talked about ensuring that `::init()` in this method is only called during `admin_init`

borkweb (commented):
Another item is bullet 2 in my suggestion in Slack for the filtering of `http_request_args` to pass along the common version to the opt-in request.

src/Common/Telemetry/Telemetry.php
Camwyn: `boot()` runs on `tribe_plugins_loaded` and fires `tec_common_telemetry_preload`

`init()` runs on `admin_init`:
When `tec_common_telemetry_preload` fires -> `Provider::hook_telemetry_init()` is called -> it hooks `Provider::initialize_telemetry()` on `admin_init` -> then on `admin_init`, `initialize_telemetry` calls `Telemetry::init()`

src/Common/Telemetry/Telemetry.php
```diff
@@ -133,8 +141,6 @@ public function boot(): void {
 		// Set a unique plugin slug.
 		Config::set_stellar_slug( $stellar_slug );
 
-		self::$plugin_path  = \Tribe__Main::instance()->get_parent_plugin_file_path();
```
Camwyn, line 136: took this out because it was being run twice.

src/Common/Telemetry/Telemetry.php
borkweb, line 133:
```suggestion
		$telemetry_server   = ! defined( 'STELLARWP_TELEMETRY_SERVER' ) ? 'https://telemetry.stellarwp.com/api/v1': STELLARWP_TELEMETRY_SERVER;
```

src/Common/Telemetry/Telemetry.php
borkweb: Ah, I didn't realize `init()` was being called on admin_init.

borkweb (dismissed):

bordoni (dismissed):

PR #1991, opened 2023-10-05
From WordPress 6.3 leverage the Script API strategy feature to add defer/async.

adamsilverstein:
## Context
WordPress introduced a new strategy feature in the Script API in version 6.3. See [this post](https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/) for more details.

I am working on a [PR](https://github.com/the-events-calendar/the-events-calendar/pull/4392) that leverages this new feature to add `defer` and `async` (focusing on front end scripts for now). In order for this to actually work in The Events Calendar, some changes are also needed here in the common library.

## Details
* From WordPress 6.3 forward, no longer add the `script_loader_tag` hooked `filter_tag_async_defer` function which was used to add async/defer previously
* No longer cast `in_footer` argument to a boolean, from WP 6.3 this argument can be an array
* Add defer to the underscore before/after scripts to ensure their execution order remains intact.
* Ensure front end scripts are deferred including 'tribe-query-string' and 'tribe-common'

Note that using this core strategy feature means that if any other scripts depends on a script marked for defer is not itself deferrable, neither script will be deferred. This is to ensure script execution order remains intact.

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -39,7 +39,11 @@ class Tribe__Assets {
 		// Hook the actual registering of.
 		add_action( 'init', [ $this, 'register_in_wp' ], 1, 0 );
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_stellar_wp_fonts' ] );
-		add_filter( 'script_loader_tag', [ $this, 'filter_tag_async_defer' ], 50, 2 );
+
+		// From 6.3 leverage the build in defer/async strategy feature.
+		if ( version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.3', '<' ) ) {
+			add_filter( 'script_loader_tag', [ $this, 'filter_tag_async_defer' ], 50, 2 );
+		}
 		add_filter( 'script_loader_tag', [ $this, 'filter_modify_to_module' ], 50, 2 );
 		add_filter( 'script_loader_tag', [ $this, 'filter_print_before_after_script' ], 100, 2 );
 
@@ -705,7 +709,7 @@ class Tribe__Assets {
 
 		// Clean these
 		$asset->priority  = absint( $asset->priority );
-		$asset->in_footer = (bool) $asset->in_footer;
+		$asset->in_footer = $asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
 		$asset->media     = esc_attr( $asset->media );
 
 		// Ensures that we have a priority over 1.
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -637,6 +637,7 @@ class Tribe__Assets {
 			'print'         => false,
 
 			'async'         => false,
+			'defer'         => false,
 			'module'        => false,
 
 			// Print before and after.
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -563,6 +563,7 @@ class Tribe__Assets {
 	 * Register an Asset and attach a callback to the required action to display it correctly.
 	 *
 	 * @since 4.3
+	 * @since 5.2 The `$in_footer` parameter was expanded to accept a boolean or an array, mirroring the change in WordPress core.
 	 *
 	 * @param object            $origin    The main object for the plugin you are enqueueing the asset for.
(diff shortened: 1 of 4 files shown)


src/Tribe/Assets_Pipeline.php
```diff
@@ -35,9 +35,9 @@ public function prevent_underscore_conflict( $tag, $handle ) {
 
 		if ( 'underscore' === $handle ) {
 			$dir = Tribe__Main::instance()->plugin_url . 'src/resources/js';
-			$tag = "<script src='{$dir}/underscore-before.js'></script>\n"
+			$tag = "<script src='{$dir}/underscore-before.js' defer></script>\n"
 			       . $tag
-			       . "<script src='{$dir}/underscore-after.js'></script>\n";
+			       . "<script src='{$dir}/underscore-after.js' defer></script defer>\n";
```
sc0ttkclark, line 40: @adamsilverstein I wasn't expecting `</script defer>` is that intentional?

adamsilverstein: no that looks off, will fix - thanks.

adamsilverstein: fixed

sc0ttkclark (approved):

src/Tribe/Assets_Pipeline.php
```diff
@@ -35,9 +35,9 @@ public function prevent_underscore_conflict( $tag, $handle ) {
 
 		if ( 'underscore' === $handle ) {
 			$dir = Tribe__Main::instance()->plugin_url . 'src/resources/js';
-			$tag = "<script src='{$dir}/underscore-before.js'></script>\n"
+			$tag = "<script src='{$dir}/underscore-before.js' defer></script>\n"
```
stratease, line 38: Should we be hard code `defer`ing this script? What was the idea here? Want to make sure we considered unintended consequences.

src/Tribe/Assets.php
```diff
@@ -705,7 +710,7 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
 
 		// Clean these
 		$asset->priority  = absint( $asset->priority );
-		$asset->in_footer = (bool) $asset->in_footer;
+		$asset->in_footer = $asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
```
stratease, line 713:
```suggestion
		$asset->in_footer = is_array($asset->in_footer) ? $asset->in_footer : (bool)$asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
```
For better compatibility, we probably should check if array else cast to bool.

stratease (changes requested):
Small tweaks and a question. Nice work!

src/Tribe/Assets_Pipeline.php
adamsilverstein: Ah, good question!

My goal here is make sure all three of these scripts execute in the correct order AND are deferred. The before and after scripts can be deferred as long as underscore is also deferred, however it is possible it won't be - if it has a non deferrable dependency. 

I think this can be adjusted to take that into account, possibly directly here with a check, or possible all three scripts (underscore-before, underscore and underscore-after) should all use the `wp_enqueue_script` approach rather than directly outputting these wrapper scripts. 

I will review the scripts again and suggest an improved approach here.

src/Tribe/Assets.php
adamsilverstein: Makes sense, thanks!

src/Tribe/Assets_Pipeline.php
adamsilverstein: @stratease - I improved this in a0b38d52b56199f6cd93aa4bc1abd63719d4dc14 - now the defer attributes are only added when `underscrore` is already deferred. I also added an incline comment to explain why the defer attribute is added here (namely: to ensure execution order remains intact, which is true when all scripts are deferred).

Let me know if you have any questions!

stratease:
Picked this back up.

Forked to this PR https://github.com/the-events-calendar/tribe-common/pull/2051 for tracking and some final touches. Trying to get in an upcoming release...

Thanks for the work on this @adamsilverstein

PR #1992, opened 2023-10-06
Feature/new telemetry

Camwyn:


src/Common/Telemetry/Telemetry.php
```diff
@@ -639,4 +653,115 @@ public static function disable_modal( $slug, $enable = false ) {
...
+		}
+
+		// Grab all the options, cached.
+		$options = wp_load_alloptions();
+
+		// Filter out everything but our license keys.
+		$options = array_filter(
+			$options,
+			function( $key ) {
+				return str_starts_with( $key, "pue_install_key_");
```
Camwyn, line 732: Note this does not get licenses from ET+ and some other plugins as they store them differently. WIP.

src/Common/Telemetry/Telemetry.php
```diff
@@ -639,4 +653,115 @@ public static function disable_modal( $slug, $enable = false ) {
...
+		}
+
+		if ( ! empty( $telemetry['stellar_licenses'] ) ) {
+			// return changed data set and flag.
+			return [ true, $telemetry ];
+		}
+
+		// Fallback - no changes.
+		return [ false, $telemetry ];
+	}
```
borkweb, line 766: Consider returning just a boolean and passing `$telemetry` by reference. That way you don't need to do a fancy return type. (note: there are also a handful of standards changes below as well). I'd suggest not merging the below and, instead, copy/pasting the suggested changes into your editor to see what truly has changed.

```suggestion
	 * @param object $telemetry_by_reference The telemetry data object.
	 *
	 * @return bool If the data has changed.
	 */
	protected function add_licenses( &$telemetry_by_reference ): bool {
		// No data sent, bail safely.
		if ( empty( $telemetry_by_reference ) ) {
			return false;
		}
		
		$telemetry = $telemetry_by_reference;

		// We don't need the path info for this.
		$tec_slugs = array_keys( self::get_tec_telemetry_slugs() );
		// Remove parent slugs - they don't have licenses.
		$tec_slugs = array_diff( $tec_slugs, self::$base_parent_slugs );

		// Nothing to add, bail safely.
		if ( empty( $tec_slugs ) ) {
			return false;
		}

		if ( ! isset( $telemetry['stellar_licenses'] ) ) {
			// Set up if it doesn't exist.
			$telemetry['stellar_licenses'] = [];
		} elseif( ! is_array( $telemetry['stellar_licenses'] ) ) {
			// Make sure it's an array if it does exist.
			$telemetry['stellar_licenses'] = (array) $telemetry['stellar_licenses'];
		}

		// Grab all the options, cached.
		$options = wp_load_alloptions();

		// Filter out everything but our license keys.
		$options = array_filter(
			$options,
			static function( $key ) {
				return str_starts_with( $key, 'pue_install_key_' );
			},
			ARRAY_FILTER_USE_KEY
		);

		// No keys found.
		if ( empty( $options ) ) {
			return false;
		}

		foreach ( $options as $slug => $key ) {
			$slug = str_replace( 'pue_install_key_', '', $slug );
			$slug = str_replace( '-', '_', $slug );

			/**
			 * Filter the license slug.
			 * This allows for some plugins that use older nomenclature (like Filter Bar) to add a more "modern" slug.
			 *
			 * @since TBD
			 *
			 * @param string $slug The stored license slug.
			 */
			$slug = apply_filters( 'tec_telemetry_license_slug_' . $slug, $slug );

			$telemetry['stellar_licenses'][ $slug ] = $key;
		}

		if ( ! empty( $telemetry['stellar_licenses'] ) ) {
			// Update telemetry to hold the new modifications.
			$telemetry_by_reference = $telemetry;
			return true;
		}

		// Fallback - no changes.
		return false;
	}
```

src/Common/Telemetry/Telemetry.php
borkweb, line 732: `str_starts_with()` doesn't exist for PHP 7.4.

PR #1994, opened 2023-10-09
Fix AM/PM time format for French locale in the `tribe_format_date()` function.

EugeneKyale:
## Ticket
[TEC-4807](https://stellarwp.atlassian.net/browse/TEC-4807)

## 🛠️ Changes
This fixes an issue where the French locale was not respecting the `AM/PM` time format for the `g:i A` and `g:i a` time formats.

To fix this issue, the `tribe_format_date()` function now temporarily overrides the locale to `English (US)` when formatting dates and times with the French locale and either of the `g:i A` or `g:i a` time formats. This ensures that the `AM/PM time` format is respected.

This change is necessary because French does not have a dedicated `AM/PM` system. Instead, French speakers use 24-hour time.

### Before

image

### After

image

[TEC-4807]: https://stellarwp.atlassian.net/browse/TEC-4807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/functions/template-tags/date.php
+++ b/src/functions/template-tags/date.php
@@ -55,8 +55,7 @@ if ( ! function_exists( 'tribe_format_date' ) ) {
 		$original_locale = get_locale();
 
 		/**
-		 * Temporarily override locale for French AM/PM time display.
-		 * This addresses an issue where the French locale was not respecting the AM/PM time format.
+		 * Temporarily override locale to English (US) for French AM/PM time display, as French does not have a dedicated AM/PM system.
 		 */
 		if ( 'fr_FR' === $original_locale && ( 'g:i A' === $date_format || 'g:i a' === $date_format ) ) {
 			switch_to_locale( 'en_US' );
--- a/src/functions/template-tags/date.php
+++ b/src/functions/template-tags/date.php
@@ -54,10 +54,13 @@ if ( ! function_exists( 'tribe_format_date' ) ) {
 
 		$original_locale = get_locale();
 
+		// Check if the current locale is French and if the date format is using AM/PM.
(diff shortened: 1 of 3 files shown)


src/functions/template-tags/date.php
```diff
@@ -50,8 +52,24 @@ function tribe_format_date( $date, $display_time = true, $date_format = '' ) {
 			}
 		}
 
+		$original_locale = get_locale();
+
+		/**
+		 * Temporarily override locale to English (US) for French AM/PM time display, as French does not have a dedicated AM/PM system.
+		 */
+		if ( 'fr_FR' === $original_locale && ( 'g:i A' === $date_format || 'g:i a' === $date_format ) ) {
```
stratease, line 60:
```suggestion
		if ( 'fr_FR' === $original_locale && stripos( $date_format, 'a' ) !== false ) {
```
Don't have all the context, but should we be checking for the `a` character or just two very specific formats?

stratease: Also, maybe just set a flag and use the flag on the two if statements so it isn't duplicated.
```
$should_convert = 'fr_FR' === $original_locale && stripos( $date_format, 'a' ) !== false;

if($should_convert) { ...
```

EugeneKyale:
> 'g:i A'



> This function has no test coverage, perhaps we can start a `template_tags/DateTest.php` in `wpunit`?

Added recommended tests

src/functions/template-tags/date.php
EugeneKyale: It's affecting the 2 formats that use the `AM/PM` system i.e. `g:i A` and `g:i a`. 

I've updated to use a flag.

stratease (approved):

PR #1996, opened 2023-10-11
This checks notice transients for serialized classes 

Camwyn:
...and if it finds one that does not exist, it deletes the notice transient entirely.

Also fixes some logic behind transient pruning.

[TEC-4951]

[TEC-4951]: https://stellarwp.atlassian.net/browse/TEC-4951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,5 +1,9 @@
 == Changelog ==
 
+= [5.1.10.1] 2023-10-12 =
+
+* Fix - Correct a problem that can cause a fatal when plugins are deactivated in a certain order. [TEC-4951]
+
 = [5.1.10] 2023-10-11 =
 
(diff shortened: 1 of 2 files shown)


stratease (approved):
If we can add a test either in this PR or another, that would be good. Throw some "fake" serialized objects in the transients and attempt fetching from various API's.

Camwyn:
> If we can add a test either in this PR or another, that would be good. Throw some "fake" serialized objects in the transients and attempt fetching from various API's.

That's a good idea. I don't want to hold up the hotfix for it but we should def. do that in the next MR.

PR #2003, opened 2023-10-30
Defining dynamic props.

stratease:
[BTRIA-2088](https://stellarwp.atlassian.net/browse/BTRIA-2088)

```
Deprecated: Creation of dynamic property TEC\Event...
```

[BTRIA-2088]: https://stellarwp.atlassian.net/browse/BTRIA-2088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Field.php
+++ b/src/Tribe/Field.php
@@ -68,6 +68,95 @@ if ( ! class_exists( 'Tribe__Field' ) ) {
 		 */
 		public $settings;
 
+		/**
+		 * @var string
+		 */
+		public $type;
+
+		/**
+		 * @var string
+		 */
+		public $class;
+
+		/**
+		 * @var string
+		 */
+		public $label;
+
+		/**
+		 * @var array
+		 */
+		public $label_attributes;
+
+		/**
+		 * @var string
+		 */
+		public $error;
+
+		/**
+		 * @var string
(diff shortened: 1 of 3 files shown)


src/Tribe/Field.php
```diff
@@ -68,6 +68,95 @@ class Tribe__Field {
 		 */
 		public $settings;
 
+		/**
+		 * @var string
+		 */
+		public $type;
```
Camwyn, line 74: These should really all be:
```suggestion
		public string $type;
```

Camwyn (changes requested):

src/Tribe/Field.php
stratease: I don't think it's super safe - I intentionally avoided type casting the props just to not run into edge case fatals I may miss in this fast pass. I saw some that were null/string and others I wasn't sure about. I think we want safe first at this point.

Camwyn: Fair, I'll grudgingly allow it :wink:

Camwyn (approved):

PR #2024, opened 2023-11-28
Change ETWP Event Add-Ons image

redscar:
image


juanfra (changes requested):
It looks like the image has a different aspect ratio and it looks out of line with the other images. Can you please ask David for the image with the same aspect ratio so that the size of the images within the grid is respected?

redscar:
> It looks like the image has a different aspect ratio and it looks out of line with the other images. Can you please ask David for the image with the same aspect ratio so that the size of the images within the grid is respected?

I just updated the artifact with the new image David sent over.

juanfra (approved):
Thanks for working on this Brian!

PR #2026, opened 2023-12-06
Improve the Select Ajax Requests.

stratease:
[BTRIA-2148](https://stellarwp.atlassian.net/browse/BTRIA-2148)

Adding whitelist to input params.

image

[BTRIA-2148]: https://stellarwp.atlassian.net/browse/BTRIA-2148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Ajax/Dropdown.php
+++ b/src/Tribe/Ajax/Dropdown.php
@@ -281,7 +281,7 @@ class Tribe__Ajax__Dropdown {
 		$filter = sanitize_key( 'tribe_dropdown_' . $args->source );
 		if ( has_filter( $filter ) ) {
 			$data = apply_filters( $filter, [], $args->search, $args->page, $args->args, $args->source );
-		} elseif ( in_array( $args->source,['search_terms', 'search_posts'] ) ) {
+		} elseif ( in_array( $args->source, ['search_terms', 'search_posts'] ) ) {
 			$data = call_user_func_array( [ $this, $args->source ], array_values( (array) $args ) );
 		}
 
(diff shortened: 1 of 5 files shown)


src/Tribe/Ajax/Dropdown.php
```diff
@@ -235,20 +235,33 @@ public function convert_children_to_array( $results ) {
...
 	 */
 	public function parse_params( $params ) {
-		$defaults = [
+		$defaults       = [
 			'search' => null,
 			'page'   => 0,
-			'args'   => [],
 			'source' => null,
 		];
+		$allowed_params = [
```
Camwyn, line 250: Why not just do a list of allowed keys and `array_intersect_key`?

Camwyn (changes requested):

src/Tribe/Ajax/Dropdown.php
stratease: I think it would be more complicated/lines of code. As you mentioned this does the defaults too, with `array_intersect_key` we would not get defaults right?

```php
        $allowed_params = [
            'search' => $params['search'] ?? null,
            'page'   => $params['page'] ?? 0,
            'source' => $params['source'] ?? null,
            'args'   => [ 'post_status' => 'publish' ],
        ];
```

src/Tribe/Ajax/Dropdown.php
```diff
@@ -274,7 +281,7 @@ public function route() {
 		$filter = sanitize_key( 'tribe_dropdown_' . $args->source );
 		if ( has_filter( $filter ) ) {
 			$data = apply_filters( $filter, [], $args->search, $args->page, $args->args, $args->source );
-		} else {
+		} elseif ( in_array( $args->source,['search_terms', 'search_posts'] ) ) {
```
Camwyn, line 284:
```suggestion
		} elseif ( in_array( $args->source, ['search_terms', 'search_posts'] ) ) {
```

Camwyn (approved):

PR #2047, opened 2024-01-19
feat(modules/utils) add the noRsvpsOnRecurring utility function

lucatume:
Add the `noTicketsOnRsvp` function to the utilities to pick up the `tec-no-rsvps-on-recurring` body class.

The method is used by the Event Tickets plugin to discern whether RSVP ticktes should be added to recurrring Events or not.

### DIFF
--- a/src/modules/utils/recurrence.js
+++ b/src/modules/utils/recurrence.js
@@ -40,5 +40,5 @@ export const noTicketsOnRecurring = () => {
  * @returns {boolean} Whether RSVPs are allowed on Recurring events or not.
  */
 export const noRsvpsOnRecurring = () => {
-	return document.body.classList.contains('tec-no-rsvp-on-recurring')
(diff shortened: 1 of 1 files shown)


src/modules/utils/recurrence.js
```diff
@@ -25,8 +25,20 @@ export const hasRecurrenceRules = ( state ) => {
...
+
+/**
+ * Returns whether RSVPs are allowed on Recurring events or not.
+ *
+ * @since 5.8.0
+ *
+ * @returns {boolean} Whether RSVPs are allowed on Recurring events or not.
+ */
+export const noRsvpsOnRecurring = () => {
+	return document.body.classList.contains('tec-no-rsvp-on-recurring')
```
bordoni, line 43:
```suggestion
	return document.body.classList.contains( 'tec-no-rsvp-on-recurring' )
```

bordoni (changes requested):

bordoni (approved):

PR #2051, opened 2024-01-30
Adding support for defer/async from WP 6.3

stratease:
[TECENG-48](https://stellarwp.atlassian.net/browse/TECENG-48)

Utilize defer API available in WordPress 6.3.

See https://github.com/the-events-calendar/tribe-common/pull/1991

**Required by:** https://github.com/the-events-calendar/the-events-calendar/pull/4470

[TECENG-48]: https://stellarwp.atlassian.net/browse/TECENG-48?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Assets.php
```diff
@@ -39,7 +39,11 @@ public function __construct() {
 		// Hook the actual registering of.
 		add_action( 'init', [ $this, 'register_in_wp' ], 1, 0 );
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_stellar_wp_fonts' ] );
-		add_filter( 'script_loader_tag', [ $this, 'filter_tag_async_defer' ], 50, 2 );
+
+		// From WordPress 6.3 leverage the Script API strategy feature to add defer/async.
+		if ( version_compare( strtok( get_bloginfo( 'version' ), '-' ), '6.3', '<' ) ) {
```
bordoni, line 44: Avoid the DB lookup `global $wp_version;`

src/Tribe/Main.php
```diff
@@ -225,7 +225,7 @@ public function load_assets() {
 			$this,
 			[
 				[ 'tribe-accessibility-css', 'accessibility.css' ],
-				[ 'tribe-query-string', 'utils/query-string.js' ],
+				[ 'tribe-query-string', 'utils/query-string.js', null, null, [ 'in_footer' => [ 'strategy' => 'defer' ] ] ],
```
bordoni, line 228: Why are we changing this to defer? Is that fixing a particular bug?

src/Tribe/Assets_Pipeline.php
```diff
@@ -35,9 +35,9 @@ public function prevent_underscore_conflict( $tag, $handle ) {
 
 		if ( 'underscore' === $handle ) {
 			$dir = Tribe__Main::instance()->plugin_url . 'src/resources/js';
-			$tag = "<script src='{$dir}/underscore-before.js'></script>\n"
-			       . $tag
-			       . "<script src='{$dir}/underscore-after.js'></script>\n";
+			$tag = "<script src='{$dir}/underscore-before.js' defer></script>\n"
+				. $tag
+				. "<script src='{$dir}/underscore-after.js' defer></script>\n";
```
bordoni, line 40: Why defer to these?

bordoni (changes requested):

src/Tribe/Assets.php
```diff
@@ -705,7 +710,7 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
 
 		// Clean these
 		$asset->priority  = absint( $asset->priority );
-		$asset->in_footer = (bool) $asset->in_footer;
+		$asset->in_footer = is_array( $asset->in_footer ) ? $asset->in_footer : (bool) $asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
```
bordoni, line 713: Lets not change how `in_footer` works, when we see the new version of WordPress just modify how we use the values.

bordoni (changes requested):
I would rather we dont change the stored value for `in_footer` but how we pass the existing values to the `wp_register_script`.

src/Tribe/Main.php
stratease: I believe it was just to optimize the page load. @adamsilverstein ?

src/Tribe/Assets.php
stratease: @bordoni we have to change something here or the `array` will be cast to a `bool` everytime. Remove the casting altogether? Not sure whether this is here for a particular situation and if it will be safe to remove the casting.

Camwyn: https://github.com/the-events-calendar/tribe-common/pull/2054
Don't use `in_footer` - use existing `async`/`defer` params

bordoni: What I meant was, dont use the `in_footer` from Tribe Assets, use the already existing `async`/`defer` we have plus the `in_footer` value to compose the array for WordPress.

bordoni:
Have to close this PR since we will be using https://github.com/stellarwp/assets/ to handle assets, this change will need to be ported over there.

PR #2054, opened 2024-02-09
Modify the approach in #2051

Camwyn:
Alternate approach to backwards compat

See #2051

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -188,16 +188,19 @@ class Tribe__Assets {
 
 		$tag_has_async = false !== strpos( $tag, ' async ' );
 		$tag_has_defer = false !== strpos( $tag, ' defer ' );
-		$replacement = '<script ';
 
-		if ( $asset->async && ! $tag_has_async ) {
-			$replacement .= 'async ';
+		// Note we only support one of these tags at a time. So if one exists, we bail.
+		if ( $tag_has_async || $tag_has_defer ) {
+			return $tag;
 		}
 
-		if ( $asset->defer && ! $tag_has_defer ) {
-			$replacement .= 'defer ';
-		}
+		$replacement = '<script '; // Trailing space is intentional and important.
 
+		if ( $asset->async ) {
+			$replacement .= 'async '; // Again, trailing space is important.
+		} elseif ( $asset->defer  ) {
+			$replacement .= 'defer '; // Again, trailing space is important.
+		}
 
 		return str_replace( '<script ', $replacement, $tag );
 	}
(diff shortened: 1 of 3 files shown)


src/Tribe/Assets.php
```diff
@@ -710,9 +702,25 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
...
 		// Clean these
 		$asset->priority  = absint( $asset->priority );
-		$asset->in_footer = is_array( $asset->in_footer ) ? $asset->in_footer : (bool) $asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
 		$asset->media     = esc_attr( $asset->media );
 
+		// Since WordPress 6.3, the `in_footer` parameter accepts an array argument.
+		if ( version_compare( strtok( $wp_version, '-' ), '6.3', '>=' ) ) {
+			// if `in_footer` is set to boolean true, add it to the `in_footer` array. i.e. [ 'in_footer' => true ].
+			if ( (bool) $asset->in_footer ) {
+				$asset->in_footer['in_footer'] = (bool) $asset->in_footer;
```
stratease, line 711:
```
				$asset->in_footer['in_footer'] = (bool) $asset->in_footer;
```
This comes in as a `bool` and gets converted to an `array`?

stratease (changes requested):

src/Tribe/Assets.php
Camwyn: Yes,. instead of overloading the `in_footer` param where it's sent, we continue to send it as a bool, and if conditions are right (proper WP version and async of defer set) we convert it to an array that contains the original boolean value.

stratease: Won't it `PHP Warning:  Uncaught Error: Cannot use a scalar value as an array` ?

stratease (changes requested):

src/Tribe/Assets.php
```diff
@@ -710,9 +702,25 @@ public function register( $origin, $slug, $file, $deps = [], $action = null, $ar
 
 		// Clean these
 		$asset->priority  = absint( $asset->priority );
-		$asset->in_footer = is_array( $asset->in_footer ) ? $asset->in_footer : (bool) $asset->in_footer; // Since WordPress 6.3, this parameter accepts an array argument.
 		$asset->media     = esc_attr( $asset->media );
 
+		// Since WordPress 6.3, the `in_footer` parameter accepts an array argument.
+		if ( version_compare( strtok( $wp_version, '-' ), '6.3', '>=' ) ) {
+			// if `in_footer` is set to boolean true, add it to the `in_footer` array. i.e. [ 'in_footer' => true ].
+			if ( (bool) $asset->in_footer ) {
```
stratease, line 710:
```suggestion
			if ( is_bool( $asset->in_footer ) ) {
```
Is this what you meant?

src/Tribe/Assets.php
stratease, line 711:
```suggestion
				$asset->in_footer = [ 'in_footer' => $asset->in_footer ];
```
Avoid the error?

stratease (changes requested):

bordoni:
We will need to port this PR to https://github.com/stellarwp/assets/

PR #2062, opened 2024-03-18
refactor(Controller_Test_Case)

lucatume:
### 🎫 Ticket

No ticket, born in the context of wp-browser update.

### 🗒️ Description

Rename the `setUp` method to `setUpTestCase` and annotate it with `@before`.
Rename the `tearDown` method to `tearDownTestCase` and annotate it with `@after`.
Not overriding the base test case `setUp` and `tearDown` methods allows the `Controller_Test_Case` to be used across different PHPUnit versions.
Newer versions would require the `: void` return type that older versions will not support.

### Artifacts

There are no tests covering the `Controller_Test_Case` directly, but I've run Event Tickets and The Events Calendar tests relying on the `Controller_Test_Case` and did not run into issues.

### ✔️ Checklist
- [x] Changelog entry in the `readme.txt` file.


Camwyn (approved):
I only see (current coed) impacts in ECP and ET.
This won't require changing anything and feels like an improvement.

Let's merge it and then update any existing release branches.

PR #2064, opened 2024-03-29
Update wp-browser to v3.5

lucatume:
Part of Innovation Day work.

This PR updates wp-browser to version `3.5` and, along with it, Codeception to version `4` and PHPUnit to version `9.6`.

I've updated the test suite configurations (not really required, but it slims them down) and update the tests to use the correct PHPUnit assertions as required.


bordoni (approved):

bordoni:
@Camwyn can we make sure this gets merged into `release/T24.asag`?

PR #2076, opened 2024-04-25
Adding date fns and removing momentjs libs. Some proposed approaches.

stratease:
### 🎫 Ticket

[TEC-4242]

### 🗒️ Description

Updating our javascript modules to leverage `date-fns` and vanilla JS instead of `moment`. This update touches Block Editor and Classic Editor differently.

### 🎥 Artifacts 

image

### ✔️ Checklist

[TEC-4242]: https://stellarwp.atlassian.net/browse/TEC-4242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jest.setup.js
```diff
@@ -27,4 +27,3 @@ global.render = render;
 global.mount = mount;
 global.renderer = renderer;
 
-moment.tz.setDefault( 'UTC' );
```
stratease, line 30: Will we need this? Global settings?

src/modules/elements/time-picker/element.js
```diff
@@ -89,9 +87,9 @@ const TimePicker = ( {
 	);
 
 	const formatLabel = ( seconds ) => {
-		return momentUtil
+		return dateUtil
```
stratease, line 90:
```suggestion
		return dateUtil
```
Maybe we abstract out whatever we use inside our "util". This avoids being tied to a particular lib, centralizing a dependency?

stratease (commented):
Just a thought, but if we migrate directly "inline" it could be quite cumbersome to do. I don't recommend we clone `moments` api, but perhaps we abstract out a bit and hide the implementation details. 

For example: 

Currently it's API is like this:
```
var comparison_end_time = moment( obj.end_time, 'h:mm a' ).format( 'HH:mm' );
```
And if we directly move to `date-fns` it would be something like this:
```
var comparison_start_time = obj.start_time ? dateFns.format( dateFns.toDate( obj.start_date + ' ' + obj.start_time ), 'HH:mm' ) : null;
```
with potentially more refactoring to be done to handle a native `Date` vs `moment`. 

Perhaps we middle ground, and use `date-fns` and `vanilla` inside a util that will hide the implementation details.

I'm not sold on either way, both have pros and cons.

bordoni (changes requested):
We need to update the `package-safelist.json`

Camwyn:
We wound up going with Dayjs for now

PR #2079, opened 2024-05-01
Replace Assets with stellarwp/assets

lucatume:
This PR aims at replacing Common `Tribe__Assets` library with the one from `stellarwp/assets`.

Looking around, it seems the most widely used entry point for asset registration is the `tribe_asset` function.
For this reason I've picked that as the entrypoint and middleware between the `stellarwp/assets` library and the rest of the plugins.

Currently, the code in its current form loads correctly about 80% of the assets.

TODO:
1. Work out why _some_ assets are not being loaded correctly.
2. Since the `stellarwp/assets` library is configured in Common, but then used by all plugins, the only way to dynamically work out the `file` argument is to make it a URL. It works for most things, but not for some. It also runs an `is_file` check that, while cached, is not ideal.
3. The adapter code currently in the body of the `tribe_asset` function should likely be moved to a dedicated adapter class to expose the same API and keep the code out of the template tag.
4. Dedicated testing


lucatume:
[Opened PR on `stellarwp/assets` to add missing suports.](https://github.com/stellarwp/assets/pull/7)

PR #2081, opened 2024-05-03
SLR Support

lucatume:


bordoni (dismissed):
I wonder which branch we should use, I would urge this to be released with the plugin consolidation. So probably `bucket/plugin-consolidation`

src/Tribe/Assets.php
```diff
@@ -34,17 +25,54 @@ public static function instance() {
 	 * Register the Methods in the correct places.
 	 *
 	 * @since 4.3
+	 * @since TBD Emptied of all hooks in favor fo the stellarwp/assets library.
```
Camwyn, line 28:
```suggestion
	 * @since TBD Emptied of all hooks in favor of the stellarwp/assets library.
```

Camwyn (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tribe-common",
-  "version": "5.2.7",
+  "version": "5.4.0-dev",
```
stratease, line 3: We skipped 5.3?

src/functions/template-tags/general.php
```diff
@@ -559,14 +564,14 @@ function tribe_format_currency( $cost, $post_id = null, $currency_symbol = null,
 	 *
 	 * @category Events
 	 * @param  string $optionName Name of the option to retrieve.
-	 * @param string  $default    Value to return if no such option is found.
+	 * @param string $default    Value to return if no such option is found.
```
stratease, line 567: Can you fix the rest of the docblock spacing, since you started?

stratease (changes requested):

PR #2088, opened 2024-05-18
EVA-145 Deprecated Constants

pattihis:
### 🎫 Ticket

[EVA-145]

### 🗒️ Description

EVA-145 Deprecated Constants

### ✔️ Checklist

[EVA-145]: https://stellarwp.atlassian.net/browse/EVA-145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Event_Automator/Plugin_Register.php
```diff
@@ -173,4 +175,15 @@ public function add_et_and_tec_as_loose_dependency( $dependencies ) {
...
 	}
+
+	/**
+	 * Load the deprecated constants.
+	 *
+	 * @since TBD Migrated to Common from Event Automator
+	 */
+	protected function load_deprecated() {
+		if ( ! defined( 'EVENT_AUTOMATOR_FILE' ) ) {
+			define( 'EVENT_AUTOMATOR_FILE', __FILE__ );
```
bordoni, line 186: Should this point to the `Plugin_Register` file or to the main plugin file?

bordoni (changes requested):

src/Common/Event_Automator/Plugin_Register.php
pattihis: @bordoni I don't think that it matters anymore. In the original plugin it was pointing to the file at the root, that does not exist anymore, and was used to load files that we don't load anymore. Where should I point that constant to?

bordoni: Oh yeah for sure, but I would like to to be consistent when moving in case third-party developers were using in some weird way.

stratease: @Camwyn and @bordoni I think the only scenario that someone may be using this legitimately is if they were loading EVA `admin-views`, and those are in `common` now. So that file path should be local to those resources (and any others of similar vein which should be in `common` as well). This one is a bit quirky because we are splitting EVA among 3 locations and `common` is the "common denominator".

It will likely take a chunk of time to move and dissect the template files up to the parent plugins, probably bigger risk than it's worth. 

My main assumption here is a lot of what we are doing for EVA in `common` now will be deprecated and formally organized and moved into ECP and ET+ when we decide to tackle that.

I see this as the smallest lift and smallest risk for now - but please let us know if I'm missing something?

pattihis: Yes, and for the `src/admin-views/` templates to work as before, we would need to put that constant in here:

https://github.com/the-events-calendar/tribe-common/blob/master/tribe-common.php

That does not feel right to me. Please advice...

bordoni: Fair point. Proceed as is.

bordoni (approved):

PR #2093, opened 2024-05-24
Feature/stellarwp assets

Camwyn:
🎫 Ticket
[TCMN-172](https://stellarwp.atlassian.net/browse/TCMN-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

🗒️ Description
Convert to usage of [StellarWP Assets](https://github.com/stellarwp/assets)

[TCMN-172]: https://stellarwp.atlassian.net/browse/TCMN-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2096, opened 2024-06-05
fix(Assets) enqueue the minified version of a package if available

lucatume:
This PR fixes an issue where the `Tribe__Assets::register()` method would not correctly register the minified version of assets.

This is a follow-up fix to the work done in #2093.

The minified version of the files will be loaded if:
1. The non-minified version of the file does not exist; this is the case for vendor files whose non-minified version will not be included in the plugin package.
2. The `SCRIPT_DEBUG` constant is not set or set to `false` indicating that, where possible, the minified version of the files should be loaded.

### DIFF
--- a/src/Tribe/Assets.php
+++ b/src/Tribe/Assets.php
@@ -347,7 +347,7 @@ class Tribe__Assets {
 			return false;
 		}
 
-		$extension   = substr( $file, strrpos( $file, '.' ) + 1 );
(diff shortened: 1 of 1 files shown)


lucatume:
Closing this as a duplicate of this PR [on `stellarwp/assets`](https://github.com/stellarwp/assets/pull/10).

PR #2143, opened 2024-07-15
Uplink - Licensing Tests

redscar:
Tests for the Uplink Controller.

Tests are failing due to the live Uplink library not being updated with our new logic.

To test locally with dev Uplink branch `composer require stellarwp/uplink:dev-feat/fluent-form-builder --prefer-source`

### DIFF
--- a/tests/wpunit/Common/Libraries/Uplink/Controller_Test.php
+++ b/tests/wpunit/Common/Libraries/Uplink/Controller_Test.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace TEC\Common\Libraries\Uplink;
+
+use tad\Codeception\SnapshotAssertions\SnapshotAssertions;
+use TEC\Common\StellarWP\Uplink\Register;
+use Tribe__Main;
+
+class Controller_Test extends \Codeception\TestCase\WPTestCase {
+	use SnapshotAssertions;
+
+	/**
+	 * @before
+	 */
+	public function register_uplink() {
+		tribe( Controller::class )->register_uplink();
+		$this->register_plugin();
+	}
+
+	public function register_plugin() {
+		Register::plugin(
+			'common-test-slug',
+			'common-test',
+			'1.0.0',
+			dirname( __FILE__ ),
+			tribe( Tribe__Main::class )
+		);
+	}
+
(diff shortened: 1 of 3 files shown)


tests/wpunit/Common/Libraries/Uplink/Controller_Test.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @var Controller
+	 */
+	private $controller;
+
+	/**
+	 * @before
+	 */
+	public function setup_uplink() {
+		$this->controller = tribe( Controller::class );
+		$this->controller->register_uplink();
```
dpanta94, line 26: The provider is not being loaded by common's loading process ? I am not sure if it should cause the provider's load is bound in `plugins_loaded` ( @see `common/src/Tribe/Main.php` `tribe_register_provider( Libraries\Provider::class );` )

tests/wpunit/Common/Libraries/Uplink/Controller_Test.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @test
+	 */
+	public function it_should_remove_actions_on_unregister() {
+		$this->controller->do_register(); // First add them
+		$this->controller->unregister(); // Then remove them
+
+		// Assert actions were removed
+		$this->assertFalse( has_action( 'init', [ $this->controller, 'register_uplink' ] ) );
+		$this->assertFalse( has_filter( 'tribe_license_fields', [ $this->controller, 'register_license_fields' ] ) );
+	}
```
dpanta94, line 93: I'd suggest we combine those

tests/wpunit/Common/Libraries/Uplink/Controller_Test.php
redscar: If I remove `$this->controller->register_uplink();` each test fails because the plugin has already been registered.  Setting it up like this allows all of the tests to pass correctly. Otherwise, I get the error `[TEC\Common\StellarWP\Uplink\Exceptions\ResourceAlreadyRegisteredException] The resource "common-test-slug" is already registered.`

dpanta94 (approved):

PR #2146, opened 2024-07-15
Merge plugin consolidation

stratease:
### 🗒️ Description
Merge OM into bucket.

### ✔️ Checklist


tests/_bootstrap.php
```diff
@@ -16,7 +16,7 @@
 	 */
 
 	// TEC Test Support.
-	$tec_dir         = dirname( __DIR__, 2 );
+	$tec_dir         = dirname( __DIR__, 3 ) . '/the-events-calendar';
```
Camwyn, line 19: You shouldn't assume we are in the TEC directory!

Camwyn (approved):

Camwyn:
I've skipped all the EVA tests as I don't think they belong here - they were failing due to missing classes from parent plugins and such.

tests/_bootstrap.php
stratease: @Camwyn  It is some workaround for EVA tests. Not sure we should mess with this at this point. cc @pattihis

Camwyn: The test is skipped for now, so the heck with it. It should work the same.

Camwyn (approved):

PR #2167, opened 2024-08-08
Groundwork/cleanup

Camwyn:
### 🎫 Ticket

[TEC-5134]

### 🗒️ Description

Some cleanup and groundwork for the new nav structure to allow "nested tabs" 

### 🎥 Artifacts 
Passing tests are artifact - no new functionality here, just some renaming.

### DIFF
--- a/src/Tribe/Admin/Activation_Page.php
+++ b/src/Tribe/Admin/Activation_Page.php
@@ -187,7 +187,7 @@ class Tribe__Admin__Activation_Page {
 
 		delete_transient( $this->args['activation_transient'] );
 
-		if ( ! current_user_can( Tribe__Settings::instance()->requiredCap ) ) {
+		if ( ! current_user_can( tribe( 'settings' )->requiredCap ) ) {
 			return;
 		}
 
@@ -286,7 +286,7 @@ class Tribe__Admin__Activation_Page {
 	 * @return string
 	 */
 	protected function get_message_page_url( $slug ) {
-		$settings = Tribe__Settings::instance();
+		$settings = tribe( 'settings' );
 
 		$url = ! empty( $this->args['admin_url'] ) ? $this->args['admin_url'] : $settings->get_url();
 
@@ -329,7 +329,7 @@ class Tribe__Admin__Activation_Page {
 	 * in the Events > Settings slot instead, for this request only).
 	 */
 	protected function disable_default_settings_page() {
-		remove_action( $this->args['admin_page'], [ Tribe__Settings::instance(), 'generatePage' ] );
+		remove_action( $this->args['admin_page'], [ tribe( 'settings' ), 'generate_page' ] );
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


src/Tribe/Admin/Activation_Page.php
```diff
@@ -187,7 +187,7 @@ public function maybe_redirect() {
 
 		delete_transient( $this->args['activation_transient'] );
 
-		if ( ! current_user_can( Tribe__Settings::instance()->requiredCap ) ) {
+		if ( ! current_user_can( tribe( 'settings' )->requiredCap ) ) {
```
JPry, line 190: I like the shift to using `tribe()`, but there's one thing that I feel should be mitigated.

Technically, there is a chance that calling `tribe()` could result in an exception being thrown. I admit, this chance is very small. But since this will run in an unpredictable WordPress environment, I think we should be extra cautious.

This applies to all the places it is called.

src/Tribe/Settings.php
```diff
@@ -5,822 +5,911 @@
 	die( '-1' );
 }
 
-use Tribe\Admin\Pages as AdminPages;
+use Tribe\Admin\Pages as Admin_Pages;
 
-if ( ! class_exists( 'Tribe__Settings' ) ) {
+if ( did_action( 'tec_settings_init' ) ) {
+	// If We've already initialized, don't do it again - return the current instance.
+	return tribe( 'settings' );
```
JPry, line 12: Why the `return tribe( 'settings' );` here, instead of just `return;`?

src/Tribe/Settings.php
```diff
@@ -5,822 +5,911 @@
...
+		_deprecated_function( __METHOD__, 'TBD', 'display_success' );
+		$this->display_success();
+	}
 
-			return is_callable( $condition )
-				? call_user_func( $condition, $this->current_fields )
-				: true == $condition;
-		}
-	} // end class
-} // endif class_exists
```
JPry, line 826: While we're here, how about removing the `// end class` comment above?

JPry (changes requested):

src/Tribe/Settings.php
Camwyn: Actually a fair question - I was thinking at the time I needed to return the class, but I don't here. Need more coffee?

src/Tribe/Admin/Activation_Page.php
JPry: @Camwyn Any thoughts on this? I know it's not exactly unique to your PR but I thought it might be worth discussing or thinking about.

Camwyn: In the old code - when `Tribe__Settings::instance()` was called it returned `tribe( 'settings' )` so I think if that was going to happen it already would be happening.

I understand and appreciate the concern though.

If it does happen somewhere, we've either got a build or timing issue (that should be addressed anyway).

JPry (approved):

PR #2170, opened 2024-08-12
Structure and styling

Camwyn:
This sets up the nested tabs with a bit of CSS for the eventual desired styling.

[TEC-5134](https://theeventscalendar.atlassian.net/browse/TEC-5134)

[TEC-5134]: https://stellarwp.atlassian.net/browse/TEC-5134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Settings_Tab.php
```diff
@@ -87,6 +87,24 @@ class Tribe__Settings_Tab {
...
 	public $priority;
 
+	/**
+	 * Parent tab ID, if any.
+	 *
+	 * @since TBD
+	 *
+	 * @var string $parent
+	 */
+	public $parent = null;
```
JPry, line 97: These are new, so I think it would be better to make the `protected` (or even `private`), and provide `get_*()` methods and `has_*()` methods.

src/Tribe/Settings_Tab.php
```diff
@@ -270,6 +296,59 @@ public function do_content(): void {
...
 	}
 
+	/**
+	 * Adds a child tab to the current tab.
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Settings_Tab $tab The child tab to add.
+	 */
+	public function add_child( $tab ): void {
```
JPry, line 306: This should have a typed parameter, then we don't need to filter the children later.

```suggestion
	public function add_child( Tribe__Settings_Tab $tab ): void {
```

src/Tribe/Settings_Tab.php
```diff
@@ -270,6 +296,59 @@ public function do_content(): void {
...
+	/**
+	 * Checks if the current tab has children.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function has_children(): bool {
+		// Ensure all our children are tabs, as expected.
+		$this->children = array_filter(
```
JPry, line 320: This filtering should be unnecessary with the above comment in place, and we can simplify this method to the `return` statement on line 327.

src/Tribe/Settings_Tab.php
```diff
@@ -270,6 +296,59 @@ public function do_content(): void {
...
+	public function get_children(): array {
+		return $this->children;
+	}
+
+	/**
+	 * Gets the priority of the current tab.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The priority of the tab. This ill be a float stored as a string i.e. '5' or '5.5'.
```
JPry, line 346:
```suggestion
	 * @return string The priority of the tab. This will be a float stored as a string i.e. '5' or '5.5'.
```

src/functions/template-tags/html.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * HTML functions (template-tags) for use in WordPress templates.
+ * HTML functions ( template-tags ) for use in WordPress templates.
```
JPry, line 3: I think some find/replace process didn't distinguish between comments and code 😄

src/Tribe/Settings.php
```diff
@@ -328,30 +327,153 @@ public function init_tabs() {
...
+	 * @return Tribe__Settings_Tab|null
+	 */
+	public function get_tab( $id ): ?Tribe__Settings_Tab {
+		// Find tab if a parent.
+		$tab_object = $this->tabs[ $id ] ?? null;
+
+		// Find tab if a child tab.
+		if ( empty( $tab_object ) ) {
+			foreach( $this->tabs as $tab ) {
+				if ( $tab->has_children() && isset( $tab->children[ $id ] ) ) {
```
JPry, line 438: I don't love this... I feel like tabs should already have their children added by the time we're here.

src/Tribe/Settings.php
```diff
@@ -470,23 +650,144 @@ public function generate_page(): void {
...
+		ob_start();
+		?>
+		<li <?php tribe_classes( $class ); ?>>
+			<a
+				id="<?php echo esc_attr( $tab->id ); ?>"
+				class="tec-nav__link"
+				href="<?php echo esc_url( $url ); ?>"><?php echo esc_html( $tab->name ); ?></a>
+			<?php
+			if ( $tab->has_children() ) {
+				$this->generate_subnav( $tab, $tab->children );
```
JPry, line 727: Two things here:

1. This feels like something the parent tab should be responsible for (generating the subnav content)
2. I'd rather see the tab's children accessed via a method instead of a `public` property.

JPry (changes requested):
These are some initial thoughts based on what I'm seeing while working on my own PR

src/functions/template-tags/html.php
Camwyn: nope doh... 🤦🏼

src/Tribe/Settings.php
Camwyn: We're not adding a child, just making sure we find it

src/Tribe/Settings.php
Camwyn: 1. I disagree - we set up the tabs here, it seems natural we'd set up the subtabs as well. (renamed function)
2. Agreed - done.

PR #2171, opened 2024-08-12
Add settings sidebar classes

JPry:
### 🎫 Ticket

[TEC-5137]

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s).

[TEC-5137]: https://stellarwp.atlassian.net/browse/TEC-5137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Admin/Entities/Paragraph.php
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * Render the paragraph content.
+	 *
+	 * @return void
+	 */
+	public function render() {
+		printf(
+			'<p class="%s">%s</p>',
+			esc_attr( implode( ' ', $this->classes ) ),
```
Camwyn, line 52: tribe_classes() will render the attribute from an array (including the class="") and do your escaping: 
https://github.com/the-events-calendar/tribe-common/blob/master/src/functions/template-tags/html.php#L31

src/Common/Admin/Section.php
```diff
@@ -0,0 +1,116 @@
...
+	 * Set the classes for the title element.
+	 *
+	 * @param array $classes The classes for the title element.
+	 *
+	 * @return static
+	 */
+	public function set_title_classes( array $classes ) {
+		$classes = array_filter(
+			$classes,
+			function( $class ) {
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
			function ( $class ) {
```

src/Common/Admin/Section.php
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * Validate the level for the title element.
+	 *
+	 * @param int $level The level for the title element.
+	 *
+	 * @return void
+	 * @throws InvalidArgumentException If the level is not between 1 and 6.
+	 */
+	private function validate_title_level( int $level ) {
+		if ( $level < 1 || $level > 6 ) {
```
Camwyn, line 87: Can we make these filterable?
```php
$allowed_levels = apply_filters( 'tec_allowed_sidebar_levels', [] );
if ( in_array( $level, $allowed_levels) ) { ... }
```
That will make the exception text a bit more complicated...implode?

src/Common/Admin/Section.php
```diff
@@ -0,0 +1,116 @@
...
+	 * @return void
+	 */
+	protected function render_title() {
+		if ( empty( $this->title ) ) {
+			return;
+		}
+
+		printf(
+			'<h%1$d class="%2$s">%3$s</h%1$d>',
+			$this->title_level,
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
			tag_escape( $this->title_level ),
```

src/Common/Admin/Entities/Paragraph.php
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * @var string[]
+	 */
+	private array $classes;
+
+	/**
+	 * Paragraph constructor.
+	 *
+	 * @param string $content The content for the paragraph.
```
Camwyn, line 32:
```suggestion
	 * @param string        $content The content for the paragraph.
	 * @param array<string> $classes An array of classes for the paragraph.	 
```

src/Common/Admin/Entities/Paragraph.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	public function __construct( string $content, $classes = [] ) {
+		$this->content = $content;
+		$this->classes = array_filter(
+			$classes,
+			function( $class ) {
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
			function ( $class ) {
```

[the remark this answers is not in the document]
Camwyn: use `$single_class` or something to avoid the reserved word.

src/Common/Admin/Settings_Sidebar.php
```diff
@@ -0,0 +1,104 @@
...
+	 *
+	 * @return void
+	 */
+	private function render_header_image() {
+		if ( empty( $this->header_image_url ) ) {
+			return;
+		}
+
+		printf(
+			'<img src="%1$s" class="tribe-settings-sidebar__header-image" alt="%2$s">',
```
Camwyn, line 99: If the header image is purely decorative, it should have an empty alt attribute (and `role="presentation"`).

I'd _guess_ most of these will be decorative

Camwyn (changes requested):

src/Common/Admin/Section.php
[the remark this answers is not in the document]
JPry: @Camwyn For this one, I was leaning towards ignoring the error on this line. The `tag_escape()` function is for strings, and `$this->title_level` is validated to be an integer between 1 and 6.

src/Common/Admin/Section.php
JPry: > Can we make these filterable?

Well, this is specifically for use with heading tags, so our only options are 1 through 6.

src/Tribe/Settings_Tab.php
```diff
@@ -127,17 +151,26 @@ public function __construct( $id, $name, $args = [] ) {
 		// Parse args with defaults.
 		$this->args = wp_parse_args( $args, $this->defaults );
 
+		// Add ID and name to the args.
+		$this->args['id']   = $id;
+		$this->args['name'] = $name;
+
 		// Set each instance variable and filter.
-		$this->id   = apply_filters( 'tribe_settings_tab_id', $id );
-		$this->name = apply_filters( 'tribe_settings_tab_name', $name );
```
Camwyn, line 132: While I can't find a use of either of these filters, we can't just remove them. Incorporate them into the new code, or deprecate them.

i.e. 
`$this->args['id']   = apply_filters( 'tribe_settings_tab_id', $id );`

Oh, and they need docblocks - no rest for the weary, eh? 😉

Camwyn: Actually, not sure why we had these separate from the foreach below - all we need to do is set a default (`null`) id and name in `$defaults` - at least for name 🤔

src/Tribe/Settings_Tab.php
```diff
@@ -349,6 +366,147 @@ public function get_priority(): string {
...
+	 * Determine the field value.
+	 *
+	 * @since TBD
+	 *
+	 * @param array  $field The field array.
+	 * @param string $key   The field key.
+	 *
+	 * @return mixed The field value.
+	 */
+	protected function get_field_value( $field, $key ) {
```
Camwyn, line 379: I thought we talked about moving this into `Field.php`?

src/Tribe/Settings_Tab.php
```diff
@@ -349,6 +366,147 @@ public function get_priority(): string {
...
+			$parent_option = $field['parent_option'] ?? Tribe__Main::OPTIONNAME;
+		}
+
+		/**
+		 * Get the field's parent_option in order to later get the field's value.
+		 *
+		 * @param string $parent_option The parent option name.
+		 * @param string $key           The field key.
+		 */
+		$parent_option = apply_filters( 'tribe_settings_do_content_parent_option', $parent_option, $key );
```
Camwyn, line 395: Any _new_ filters should be namespaced properly: https://docs.theeventscalendar.com/developer/plugins/namespacing/#prefixes

So this should really be `tec_settings_do_content_parent_option`

src/Tribe/Settings_Tab.php
```diff
@@ -349,6 +366,147 @@ public function get_priority(): string {
...
+	 * @throws InvalidArgumentException If the sidebar is invalid.
+	 */
+	protected function validate_sidebar( $sidebar ) {
+		// If it's a callable or an instance of Settings_Sidebar, we're good.
+		if ( is_callable( $sidebar ) || $sidebar instanceof Settings_Sidebar ) {
+			return;
+		}
+
+		// Everything else is invalid.
+		throw new InvalidArgumentException(
```
Camwyn, line 505: Should this function return a boolean just in case?
i.e. `if ( validate_sidebar() ) {...}`

Also - don't depend on the exception throwing a fatal - a catch block might handle it silently (although this code will still stop at the point we `throw`)

src/Tribe/Traits/Array_Access.php
```diff
@@ -0,0 +1,72 @@
...
+declare( strict_types=1 );
+
+namespace Tribe\Traits;
+
+/**
+ * Trait Array_Access
+ *
+ * @since TBD
+ */
+trait Array_Access {
```
Camwyn, line 19: This is awfully similar to [src/Tribe/Data.php](https://github.com/the-events-calendar/tribe-common/blob/master/src/Tribe/Data.php)

Camwyn (changes requested):
Comments, ideas, fixes, and suggestions.

Camwyn:
Another thought:
We need to do a new upsell.

I am styling the exiting one to fit the new designs, but we need to add one. Wondering if that could fit this architecture - it's essentially a Container with a header, paragraph(s) a button and an image...

src/Common/Admin/Entities/Base_Entity.php
```diff
@@ -0,0 +1,64 @@
...
+		return ob_get_clean();
+	}
+
+	/**
+	 * Get the classes for the entity.
+	 *
+	 * @return string
+	 */
+	protected function get_classes(): string {
+		return $this->classes?->get_classes_as_string() ?? '';
```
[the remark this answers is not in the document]
JPry: Ahh, that's PHP 8+ 🤦🏻

src/Common/Admin/Entities/Heading.php
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Render the element.
+	 *
+	 * @return void
+	 */
+	public function render() {
+		printf(
+			'<h%1$d class="%2$s">%3$s</h%1$d>',
+			$this->level, // phpcs:ignore StellarWP.XSS.EscapeOutput
```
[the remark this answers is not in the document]
JPry: @Camwyn See [this comment](https://github.com/the-events-calendar/tribe-common/pull/2171#discussion_r1715819469).

I'll add a docblock to explain the lack of escaping, because I think this is one case where it is valid to skip escaping.

src/Common/Admin/Entities/Unordered_List.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Add a child to the container.
+	 *
+	 * @param List_Item $child The child to add.
+	 *
+	 * @return static
+	 * @throws InvalidArgumentException If the child is not an instance of the child type.
+	 */
+	public function add_child( $child ) {
```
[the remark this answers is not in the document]
JPry: This override is for accurate type-hinting. We want to only allow specific types of children (`List_Item` instances), not just any `Element` implementation.

src/Tribe/Settings_Tab.php
JPry: Yes we did 🤦🏻

src/Tribe/Field.php
```diff
@@ -314,43 +314,54 @@ public function __construct( $id, $field, $value = null ) {
...
+				$field = apply_filters( "tribe_field_output_{$this->type}", $field, $this->id, $this );
+
+				/**
+				 * Filter the field output by ID.
+				 *
+				 * @param string       $field        The field output.
+				 * @param string       $id           The field ID.
+				 * @param Tribe__Field $field_object The field object.
+				 */
+				echo apply_filters( "tribe_field_output_{$this->type}_{$this->id}", $field, $this->id, $this );
```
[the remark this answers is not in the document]
JPry: @Camwyn This is one I'm not surprised to see... I didn't change any behavior here, just added the filter docs. What would you recommend?

Basically, any given field type will be run through the appropriate method in this class, and escaping is handled there. But by passing through a filter, we're technically opening this up to unescaped output.

There are a few things I can think of, depending on how we want to handle this:

1. Ditch the filters. This obviously has ramifications for BC, our other plugins, and 3rd-party integrations.
2. Run the content through `wp_kses()`. This is probably the most correct thing to do, but it's possible the final output will be affected
3. Don't `echo` here, and instead handle the output within each of the methods directly.

There might be a few other options, but I'm having a difficult time determining which would be best.

src/Tribe/Traits/Array_Access.php
JPry: Thanks! I actually ended up refactoring the `Data` class to use this trait. I wanted the ability to add `implements ArrayAccess` easily to any class that needs it.

src/Tribe/Settings_Tab.php
JPry: > Should this function return a boolean just in case?

I don't think so. The reason for throwing in the first place is to ensure that we only have the data we expect in the class property, which is either a callback or `Settings_Sidebar` instance.

Where or whether the exception is caught is secondary. In fact, I would hope that we'd use a `catch` statement in an appropriate place to prevent an exception from bubbling up to the end user, if the circumstance warrants it.

For now, we'll be using it appropriately, and we can hope that any 3rd parties will discover any issues before publishing something.

If we really want to `catch` in the `add_sidebar()` method, I suppose we could do so, but I feel like it's a better development experience to surface errors in a way that is more difficult to ignore.

src/Tribe/Settings_Tab.php
JPry: This isn't actually a new filter, it was just relocated, and I added a docblock that was missing previously.

src/Common/Admin/Entities/Validate_Elements.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Validate the heading level.
+	 *
+	 * @param int $level The heading level.
+	 *
+	 * @return void
+	 * @throws InvalidArgumentException If the heading level is invalid.
+	 */
+	protected function validate_level( int $level ) {
+		if ( $level < 1 || $level > 6 ) {
```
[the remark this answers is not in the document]
JPry: > we might want to restrict it to a smaller subset for specific purposes.

I think that the validation you're talking about should be handled separately from this validation. Here, we want to make sure we're generating valid HTML elements. Ensuring that the heading type is eve more limited than that is something I feel should be handled closer to where we need that specific behavior, _if_ we need that specific behavior. 

Perhaps a good compromise would be to add a `$max_level` parameter, and the code that's being more limiting can set a different max.

JPry:
@Camwyn Just trying to surface [this comment/question](https://github.com/the-events-calendar/tribe-common/pull/2171#discussion_r1720141277). It applies to the this line:

```php
echo apply_filters( "tribe_field_output_{$this->type}_{$this->id}", $field, $this->id, $this );
```

src/Tribe/Field.php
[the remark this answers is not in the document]
Camwyn: Kses would be ok, but will likely need to be fed a list of form elements (inputs, etc) so it won't strip them out

JPry:
> Another thought:
> We need to do a new upsell.
>
> I am styling the exiting one to fit the new designs, but we need to add one. Wondering if that could fit this architecture - it's essentially a Container with a header, paragraph(s) a button and an image...

@Camwyn Take a look at b5c45295940980aa243ff94e3089b438d6caf67f. I converted the License section to use the new Elements. If we wanted to the upsell here (is that the one you're talking about?) we could add it right before the `do_action( 'tec_settings_tab_licenses' );` call.

It would look something like this (with example content):

```php
$licenses_tab->add_sidebar(
	function () {
		$sidebar = new Settings_Sidebar();
		$sidebar->add_section(
			( new Settings_Section() )
				->add_elements(
					[
						new Paragraph(
							__( 'Looking for additional functionality including recurring events, custom meta, community events, ticket sales, and more?', 'tribe-common' )
						),
						new Link(
							'https://example.com/add-ons', // todo: real link here.
							__( 'Check out the available add-ons.', 'tribe-common' )
						),
					]
				)
		);
	}
);
```

src/resources/postcss/tribe-common-admin/_tabbed-nav.pcss
```diff
@@ -29,10 +29,27 @@
...
 			background-color: var(--tec-color-background);
 			padding: var(--tec-spacer-2) var(--tec-spacer-5);
+			min-width: 100%;
+		}
+
+		.tribe-settings-form {
+			display: flex;
+			padding: 20px 32px 32px 32px;
+			align-items: flex-start;
+			gap: 32px;
```
Camwyn, line 49:
```suggestion
			gap: var(--tec-spacer-7);
```

src/resources/postcss/tribe-common-admin/_tabbed-nav.pcss
```diff
@@ -29,10 +29,27 @@
...
 
 		h1 {
 			background-color: var(--tec-color-background);
 			padding: var(--tec-spacer-2) var(--tec-spacer-5);
+			min-width: 100%;
+		}
+
+		.tribe-settings-form {
+			display: flex;
+			padding: 20px 32px 32px 32px;
```
Camwyn, line 47:
```suggestion
			padding: var(--tec-spacer-4) var(--tec-spacer-7) var(--tec-spacer-7);
```

src/resources/postcss/tribe-common-admin/_tabbed-nav.pcss
```diff
@@ -533,6 +551,18 @@
...
 
+	/* Sidebar styles */
+	.tribe-settings-sidebar {
+		align-items: flex-start;
+		align-self: stretch;
+		background-color: var(--tec-color-background);
+		display: flex;
+		flex-direction: column;
+		gap: 16px;
+		padding: 20px;
```
Camwyn, line 562:
```suggestion
		gap: var(--tec-spacer-3);
		padding: var(--tec-spacer-4);
```

Camwyn (changes requested):
Some small things

src/Tribe/Settings_Tab.php
```diff
@@ -120,19 +144,35 @@ public function __construct( $id, $name, $args = [] ) {
 			'show_save'        => true,
 			'display_callback' => false,
 			'network_admin'    => false,
-			'parent'           => null,
```
Camwyn, line 123: If you don't leave parent as a default value, we may get tabs that throw fatals for not passing one.

src/Common/Admin/Entities/Unordered_List.php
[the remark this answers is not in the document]
JPry: > You're not explicitly type hinting here, like `public function add_child( List_Item $child ) {`

Yes, that's right. With inheritance rules, it throws errors if the type changes. But what _does_ change is the docblock, and IDEs can pick up that distinction (`Element` vs. `List_Item`), and will provide the correct contextual information. Here's the visual:

Screenshot 2024-08-20 at 10 05 44

Hovering over the yellow line yields this message:

> Expected parameter of type `'TEC\Common\Admin\Entities\List_Item'`, `'TEC\Common\Admin\Entities\Plain_Text'` provided.

Essentially, this is one of the small details that makes things easier for any other developers working on this in the future.

src/Common/Admin/Entities/Link.php
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+
+	/**
+	 * Render the link.
+	 *
+	 * @return void
+	 */
+	public function render() {
+		printf(
```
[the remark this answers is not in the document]
JPry: Good catch

src/Tribe/Field.php
```diff
@@ -1112,5 +1272,46 @@ protected function sanitize_class_attribute( $class ) {
...
+				'doScreenReaderLabel' => 'do_screen_reader_label',
+			];
+
+			// Helper function to prepend the class name to the method name.
+			$prepend_class = function ( string $method_name ): string {
+				return sprintf( '%s::%s', __CLASS__, $method_name );
+			};
+
+			if ( array_key_exists( $name, $method_map ) ) {
+				_deprecated_function( $prepend_class( $name ), '4.3', $prepend_class( $method_map[ $name ] ) ); // phpcs:ignore StellarWP.XSS.EscapeOutput,WordPress.Security.EscapeOutput
```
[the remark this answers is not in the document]
JPry: Fair point. I was keeping it as-is when I consolidated the methods into the `__call()` magic method.

JPry:
Closing in favor of #2179

PR #2182, opened 2024-08-22
Fix/responsive styling

Camwyn:
Spreadsheet work

### DIFF
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -710,10 +710,17 @@ class Tribe__Settings {
 		}
 
 		$nav_id = $modal ? 'tec-settings-modal-nav' : 'tribe-settings-tabs';
+		$wrapper_classes = [
+			'tec-nav__wrapper' => true,
+			'tec-nav__wrapper--subnav-active' => false,
+		];
+		if ($this->get_tab( $this->get_current_tab() ) ->has_parent()) {
+			$wrapper_classes[ 'tec-nav__wrapper--subnav-active'  ] = true;
+		}
 
 		ob_start();
 		?>
-			<nav id="<?php echo esc_attr( $nav_id ); ?>" class="tec-nav__wrapper">
+			<nav id="<?php echo esc_attr( $nav_id ); ?>" <?php tribe_classes( $wrapper_classes ); ?>>
 				<ul class="tec-nav">
 					<?php if ( ! $modal ) : ?>
 					<li class="tec-nav__tab tec-nav__tab--skip-link">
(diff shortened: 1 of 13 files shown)


src/Tribe/Settings.php
```diff
@@ -798,7 +810,10 @@ public function generate_sidebar_modal_close(): void {
 	 */
 	public function generate_tab( Tribe__Settings_Tab $tab ) {
 		$url         = $this->get_tab_url( $tab->id );
-		$class       = [ 'tec-nav__tab' ];
+		$class       = [
+			'tec-nav__tab',
+			'tec-nav__tab--' . $tab->id,
```
JPry, line 815: Suggestion:

```suggestion
			"tec-nav__tab--{$tab->id}",
```

src/Tribe/Settings.php
```diff
@@ -864,9 +879,16 @@ public function add_child_tabs_to_nav( Tribe__Settings_Tab $tab ) {
 	 * @since TBD
 	 */
 	protected function get_modal_controls(): void {
+		$current_tab = $this->get_tab( $this->get_current_tab() );
+		if ( empty( $current_tab ) ) {
+			return;
+		}
```
JPry, line 885: Just a note that this is the second time I've seen this same block. It might not be able to be abstracted, but if it can be that might be a good thing to pull out.

src/Tribe/Settings_Tab.php
```diff
@@ -197,7 +197,7 @@ public function add_actions(): void {}
 	 * @since TBD
 	 */
 	public function add_filters() {
-		add_filter( 'tribe_settings_tabs', [ $this, 'add_tab' ] );
+		add_filter( 'tribe_settings_tabs', [ $this, 'add_tab' ], 10 );
```
JPry, line 200: This added parameter isn't needed, as it's the default value.

src/resources/postcss/tribe-common-admin/settings/_modal-nav.pcss
```diff
@@ -15,7 +15,7 @@
 		margin: 0;
 		padding: 0 var(--tec-spacer-7) var(--tec-spacer-5);
 
-		@media (--viewport-full) {
+		@media (min-width: 500px) {
```
JPry, line 18: Just curious why these aren't variables.

JPry (changes requested):

src/resources/postcss/tribe-common-admin/settings/_modal-nav.pcss
Camwyn: Was trying to figure out why they weren't being applied in the correct order.
Turns out it is something in our build process that moves media queries. The solve was actually to add max-widths on some of them - but I didn't go back through and switch them back

src/Tribe/Settings.php
Camwyn: I think I removed the other use of it but good to point out!
FYI: My habit is to abstract once I hit 3 uses.

JPry (approved):

PR #2192, opened 2024-08-30
Admin add-on and help page updates

Camwyn:
Remove Community Tickets from admin pages
Reword Community Events to just read "Community" 
Add a couple check items for Community

Current:
current

Changed (title case since the image was taken):
suggested

Removed CT (on right) here:
no-ct


src/Tribe/Plugins.php
```diff
@@ -39,7 +39,7 @@ class Tribe__Plugins {
 				'thickbox_url' => 'https://theeventscalendar.com/product/wordpress-events-calendar-pro/?TB_iframe=true',
 			],
 			[
-				'short_name'   => 'Community Events',
+				'short_name'   => 'Community',
```
JPry, line 42: Does this need to be translated?

JPry (approved):
I left one question, but it looks good

src/Tribe/Plugins.php
Camwyn: No - first plugin names aren't typically translated (especially in the admin) and second it's _mostly_ used as a key (although it _is_ put into notices as well)

PR #2225, opened 2024-10-11
Marketing Notices for Black Friday.

bordoni:
### 🎫 Ticket

[ET-2241] [TEC-5242]

### 🗒️ Description

As you review this, please make sure you keep in mind this is for a Black Friday.

Still haven't touched the existing tests here, so I am expecting them to break, will fix them tomorrow.

### 🎥 Artifacts 
https://i.bordoni.me/GqsJbYmj

--

Video update on Oct 14 with dismissal:
https://i.bordoni.me/XKC3GDY0

[ET-2241]: https://stellarwp.atlassian.net/browse/ET-2241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TEC-5242]: https://stellarwp.atlassian.net/browse/TEC-5242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Admin/Conditional_Content/Black_Friday.php
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		$date = $date->modify( '-3 days' );
+
+		return $date;
+	}
+
+	protected function should_display(): bool {
+		return true;
+	}
```
bordoni, line 56: I am fully aware this is here. For In branch QA purposes, will be removed once we pass QA.

dpanta94: Would be nice if we were able to QA appearance logic as well. A bit more complicated for sure, but it's possible to fake the current time even for QA

src/Common/Admin/Conditional_Content/Datetime_Conditional_Abstract.php
```diff
@@ -1,73 +1,78 @@
...
+use Tribe__Template as Template;
+use Tribe\Utils\Date_I18n;
 
 /**
  * Abstract class for conditional content.
  *
- * @since 4.14.7
+ * @since TBD
  */
 abstract class Datetime_Conditional_Abstract {
```
dpanta94, line 14: Aren't changes on this file breaking backwards compatibility ? e.g. ET might be extending this class and not have a delcared return type for method get_start_time being overwritten.

By the way this looks like it could be extending a Provider. The registration of Black_Friday is like just registering a provider with extra steps.

src/Tribe/Admin/Notice/Service_Provider.php
```diff
@@ -54,24 +52,5 @@ public function plugins_loaded() {
 		tribe( 'pue.notices' );
 		tribe( 'admin.notice.php.version' );
 		tribe( WP_Version::class );
-
-		if ( defined( 'TRIBE_HIDE_MARKETING_NOTICES' ) ) {
-			return;
-		}
```
dpanta94, line 60: It would still be nice to allow people to choose not to see those IMO

bordoni: Oh yeah they will be Able to do that I am going to Extend the `should_display()` and use the correct method `tec_should_hide_upsell()`

bordoni: Added

src/Common/Admin/Conditional_Content/Datetime_Conditional_Abstract.php
bordoni: Potentially, I will add back the older ones in the deprecated folder.
I've tested with all the other plugins and looked for it, there isn't anything on our codebase. But there might be something external. I am aggressively trying to reduce codebase haha.

JPry (changes requested):

src/Common/Admin/Settings_Sidebar_Section.php
```diff
@@ -0,0 +1,103 @@
...
+namespace TEC\Common\Admin;
+
+use TEC\Common\Admin\Entities\Image;
+
+/**
+ * Class Settings_Sidebar_Section
+ *
+ * @since TBD
+ */
+class Settings_Sidebar_Section extends Settings_Section {
```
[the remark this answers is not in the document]
bordoni: It's used in The Events Calendar sidebar.

src/Common/Admin/Entities/Link.php
```diff
@@ -33,19 +33,33 @@ class Link extends Base_Entity {
...
-	public function __construct( string $url, string $text, ?Classes $classes = null, ?Attributes $attributes = null ) {
+	public function __construct( string $url, $content, ?Classes $classes = null, ?Attributes $attributes = null ) {
 		$this->url  = $url;
-		$this->text = $text;
+
+		if ( is_string( $content ) ) {
+			$this->text = $content;
+		}
+
+		if ( $content instanceof Base_Entity ) {
```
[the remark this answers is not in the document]
bordoni: `Entity` doesn't exist

src/Common/Admin/Entities/Link.php
```diff
@@ -33,19 +33,33 @@ class Link extends Base_Entity {
...
-	 * @param string      $text       The text for the link.
-	 * @param ?Classes    $classes    The classes for the link.
-	 * @param ?Attributes $attributes The attributes for the link.
+	 * @param string                  $url        The URL for the link.
+	 * @param string|Base_Entity      $content    The text or entity for the link.
+	 * @param ?Classes                $classes    The classes for the link.
+	 * @param ?Attributes             $attributes The attributes for the link.
 	 */
-	public function __construct( string $url, string $text, ?Classes $classes = null, ?Attributes $attributes = null ) {
+	public function __construct( string $url, $content, ?Classes $classes = null, ?Attributes $attributes = null ) {
```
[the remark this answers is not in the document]
bordoni: For sure, but that is why I picked `Base_Entity` which is extended by `Container` so you can pick an individual one or the container, since the `render` from `Container` is still used in `__toString()` allowing me to not care if it's one or many.


```
// Opt 1
new Link( 'url', 'Text' );

// Opt 2
new Link( 'url', new Image( 'url' ... ) );

// Opt 3
$container = new Container( ... )->add_children( new Image( ... ), new Image( ... ) );
new Link( 'url', $container );
```

src/Common/Admin/Settings_Sidebar.php
```diff
@@ -42,64 +33,77 @@ class Settings_Sidebar extends Section {
...
 	 *
-	 * @since 6.1.0
+	 * @since TBD
 	 *
 	 * @param Section $section The section to add.
 	 *
 	 * @return void
 	 */
-	public function add_section( Section $section ) {
+	public function append_section( Section $section ): self {
```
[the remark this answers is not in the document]
bordoni: To add consistency with `prepend_` method.

src/Common/Admin/Settings_Sidebar.php
```diff
@@ -42,64 +33,77 @@ class Settings_Sidebar extends Section {
...
+	public function get_sections(): array {
+		/**
+		 * Filter the sidebar sections.
+		 *
+		 * @since TBD
+		 *
+		 * @param Section[]        $sections The sidebar sections.
+		 * @param Settings_Sidebar $sidebar  The sidebar object.
+		 */
+		return apply_filters( 'tec_settings_sidebar_sections', $this->sections, $this );
```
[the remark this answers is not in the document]
bordoni: We have both, the instance is supplied on a filter elsewhere. I just thought it would be nice to include a filter here.

The only reason I dont like saving back to the `$this->sections` is because it would re-apply the filter to the same `$this->section` if it was called twice back to back.

src/Common/Event_Automator/Admin/Tabs/Tabs_Provider.php
```diff
@@ -5,7 +5,7 @@
 use TEC\Common\Contracts\Service_Provider;
 
 /**
- * Class Service_Provider
+ * Class Controller
```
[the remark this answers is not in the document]
Camwyn: Note: I am under the assumption that if we start a class block with "Class XXX" the `XXX` is the _name of the class_, not it's type, abstract, etc.

If this said "Controller for settings tabs." I'd be ok with it.

src/Tribe/Admin/Notice/Service_Provider.php
```diff
@@ -54,24 +52,5 @@ public function plugins_loaded() {
 		tribe( 'pue.notices' );
 		tribe( 'admin.notice.php.version' );
 		tribe( WP_Version::class );
-
-		if ( defined( 'TRIBE_HIDE_MARKETING_NOTICES' ) ) {
```
Camwyn, line 58: You moved the notice instantiation to a controller, but you've removed this check in the process...we had this check for specific reasons (like network installs) - is that intentional?

Camwyn (changes requested):

readme.txt
```diff
@@ -286,7 +286,7 @@
 
 = [5.1.2] 2023-06-22 =
 
-* Fix - Lock our container usage(s) to the new Service_Provider contract in tribe-common. This prevents conflicts and potential fatals with other plugins that use a di52 container.
+* Fix - Lock our container usage(s) to the new Controller contract in tribe-common. This prevents conflicts and potential fatals with other plugins that use a di52 container.
```
Camwyn, line 289: Why are e retro-ing this readme?

src/Common/Admin/Conditional_Content/Datetime_Conditional_Abstract.php
```diff
@@ -1,73 +1,78 @@
...
 	 *
-	 * @since 4.14.7
-	 * @return \Tribe\Utils\Date_I18n - Date Object
+	 * @since TBD
+	 *
+	 * @return ?Date_I18n - Date Object
 	 */
-	protected function get_start_time() {
+	protected function get_start_time(): ?Date_I18n {
 		$date = Dates::build_date_object( $this->start_date, 'UTC' );
```
Camwyn, line 76: This returns `now` on fail (third param default) perhaps we should not do that and instead have it return false.

Then the lower return of null (and checks on it) could be for boolean false.

readme.txt
bordoni: It was a mistake

src/Tribe/Admin/Notice/Service_Provider.php
bordoni: No I didnt, that one was legacy. We moved to `tec_should_hide_upsell()`

src/Common/Admin/Entities/Link.php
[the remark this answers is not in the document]
JPry: @bordoni Sorry, that was supposed to be `Element`.

src/Common/Admin/Entities/Link.php
```diff
@@ -33,19 +33,33 @@ class Link extends Base_Entity {
...
+		$this->url = $url;
+
+		if ( is_string( $content ) ) {
+			$this->text = $content;
+		}
+
+		if ( $content instanceof Base_Entity ) {
+			$this->content = $content;
+		}
```
JPry, line 62: This set of checks for `$content` aren't strict enough. If something other than a `string` or `Base_Entity` is passed, then we don't have a check in place that covers that situation.

What I propose is that we allow for only `string` or `Base_Entity`, and convert a `string` into a `Plain_Text` object to handle compatibility with what is already written.

JPry, line 62:
```suggestion
		if ( is_string( $content ) ) {
			$this->content = new Plain_Text( $content );
		} elseif ( $content instanceof Base_Entity ) {
			$this->content = $content;
		} else {
			throw new InvalidArgumentException( 'Content must be a string or an instance of Base_Entity' );
		}
```

src/Common/Admin/Entities/Link.php
```diff
@@ -69,7 +83,7 @@ public function render() {
 			esc_url( $this->url ),
 			esc_attr( $this->get_classes() ),
 			$this->get_attributes(), // phpcs:ignore StellarWP.XSS.EscapeOutput,WordPress.Security.EscapeOutput
-			esc_html( $this->text )
+			$this->content ? (string) $this->content : esc_html( $this->text ) // phpcs:ignore StellarWP.XSS.EscapeOutput,WordPress.Security.EscapeOutput
```
JPry, line 86: As part of the above-mentioned proposed changes, I would say that the entirety of the `render()` method be replaced with this:

```php
		?>
		<a href="<?php echo esc_url( $this->url ); ?>"
			class="<?php echo esc_attr( $this->get_classes() ); ?>"
			<?php echo $this->get_attributes(); // phpcs:ignore StellarWP.XSS.EscapeOutput,WordPress.Security.EscapeOutput ?>>
			<?php $this->content->render(); ?>
		<?php
```

src/Common/Admin/Settings_Sidebar.php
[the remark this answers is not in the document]
JPry: > We have both, the instance is supplied on a filter elsewhere. I just thought it would be nice to include a filter here.

There *is* such a thing as too many filters in too many places where they don't belong. Adding a filter here invites a way to tweak the internals of this class instance in a way that is hard to control. 

If we want to add sections (which is a valid case), then we should use a filter *outside* of this class, and utilize the class methods to provide modifications to the class instance. Otherwise we wind up with spaghetti code that is hard to track down where changes to things happen.

src/Common/Admin/Settings_Sidebar.php
```diff
@@ -42,64 +37,94 @@ class Settings_Sidebar extends Section {
...
-	 * @param Image $image The image to set.
+	 * @param Section $section The section to add.
+	 * @param bool    $append  Whether to append the section or prepend it.
 	 *
-	 * @return void
+	 * @return self
 	 */
-	public function set_header_image( Image $image ) {
-		$this->header_image = $image;
+	public function add_section( Section $section, bool $append = true ): self {
```
JPry, line 61: I don't think we should add duality to how this method functions, and I don't think that we need to have both `append` and `prepend` methods. While I understand the use case of prepending a section, we don't have to have symmetry in the methods available. 

With that said, I think we should remove the new `append_section()` method, keep the new `prepend_section()` method, and remove the second parameter to this method.

src/Common/Admin/Settings_Sidebar_Section.php
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	public function render(): void {
+		?>
+		<div class="tec-settings-form__sidebar-section tec-settings-form__sidebar-header">
+			<?php do_action( 'tec_settings_sidebar_header_start' ); ?>
+			<?php
+			$this->render_header_image();
+			$this->render_title();
+			?>
+			<?php do_action( 'tec_settings_sidebar_header_end' ); ?>
```
JPry, line 97: This should be one block of PHP code, with comments included for the actions.

```suggestion
			<?php
			/**
			 * Fires before the sidebar header is rendered.
			 *
			 * @since TBD
			 */
			do_action( 'tec_settings_sidebar_header_start' );

			$this->render_header_image();
			$this->render_title();

			/**
			 * Fires after the sidebar header is rendered.
			 *
			 * @since TBD
			 */
			do_action( 'tec_settings_sidebar_header_end' );
			?>
```

JPry (changes requested):

src/Common/Admin/Settings_Sidebar.php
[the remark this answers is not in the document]
JPry: > I think that this gives us the opportunity to add a second param to `add_section`

Respectfully, I disagree. I think we should have less duplicity in our methods rather than more. It may be small, but it still introduces unnecessary complexity into this method.

src/Common/Admin/Settings_Sidebar.php
bordoni: Done. 🫀

src/Common/Admin/Conditional_Content/Black_Friday.php
```diff
@@ -0,0 +1,287 @@
...
+		 * @since TBD
+		 *
+		 * @param string       $slug The slug of the conditional content.
+		 * @param Black_Friday $this The Black Friday instance.
+		 */
+		do_action( 'tec_conditional_content_black_friday', 'settings-sidebar', $this );
+
+		/* translators: %1$s: Black Friday year */
+		$translated_title = sprintf(
+			esc_attr_x( '%1$s Black Friday Sale for The Events Calendar plugins, add-ons and bundles.', 'Alt text for the Black Friday Ad', 'tribe-common' ),
```
Camwyn, line 229: This should fix the phpcs issue:
```suggestion
		$translated_title = sprintf(
			/* translators: %1$s: Black Friday year */
			esc_attr_x( '%1$s Black Friday Sale for The Events Calendar plugins, add-ons and bundles.', 'Alt text for the Black Friday Ad', 'tribe-common' ),
```

src/Common/Admin/Conditional_Content/Black_Friday.php
```diff
@@ -0,0 +1,285 @@
...
+			esc_attr( $year )
+		);
+		
+		$container = new Container();
+
+		$button_attr = new Attributes(
+			[
+				'style'                                       => 'position: absolute; top: 0; right: 0; background: transparent; border: 0; color: #fff; padding: 0.5em; cursor: pointer;',
+				'data-tec-conditional-content-dismiss-button' => true,
+				'data-tec-conditional-content-dismiss-slug'   => $this->slug,
```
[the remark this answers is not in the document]
Camwyn: I've seen this too - apparently it breaks if one key is above some length. I haven't tested to find the "sweet spot"

src/Common/Admin/Conditional_Content/Controller.php
```diff
@@ -6,27 +6,33 @@
...
+use Tribe__Main as Common;
+
 /**
- * Conditional Content Provider.
+ * Conditional Content Controller.
  *
- * @since 4.14.7
+ * @since TBD
  */
-class Service_Provider extends Provider_Contract {
```
Camwyn, line 18: Where is this deprecation handled - in case some third party is using this?

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		$nonce        = tribe_get_request_var( 'nonce', false );
+		$nonce_action = $this->get_nonce_action();
+
+		if ( ! wp_verify_nonce( $nonce, $nonce_action ) ) {
+			wp_send_json( false );
+		}
+
+		// Send a JSON answer with the status of dismissal
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
		// Send a JSON answer with the status of dismissal.
```

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,187 @@
...
+		wp_send_json( $this->dismiss() );
+	}
+
+	/**
+	 * A Method to actually add the Meta value telling that this content has been dismissed.
```
Camwyn, line 104:
```suggestion
	 * A Method to actually add the Meta value recording that this content has been dismissed.
```

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,187 @@
...
+		if ( empty( $this->slug ) ) {
+			return false;
+		}
+
+		if ( is_null( $user_id ) ) {
+			$user_id = get_current_user_id();
+		}
+
+		// If this user has dismissed we don't care either
+		if ( $this->has_user_dismissed( $user_id ) ) {
```
Camwyn, line 122: Should this always bail on ID of `0`? (fallback value of `get_current_user_id()`)

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,187 @@
...
+		if ( empty( $this->slug ) ) {
+			return false;
+		}
+
+		if ( is_null( $user_id ) ) {
+			$user_id = get_current_user_id();
+		}
+
+		// If this user has dismissed we don't care either.
+		if ( ! $this->has_user_dismissed( $this->slug, $user_id ) ) {
```
Camwyn, line 150: see comment above about `0`

src/Common/Admin/Entities/Button.php
```diff
@@ -0,0 +1,57 @@
...
+	/**
+	 * Div constructor.
+	 *
+	 * @since TBD
+	 *
+	 * @param ?Classes    $classes    The classes for the div.
+	 * @param ?Attributes $attributes The attributes for the div.
+	 */
+	public function __construct( ?Classes $classes = null, ?Attributes $attributes = null ) {
+		if ( $classes ) {
```
Camwyn, line 31: ugh, the parent functions should handle passing empty values - how do I _remove_ a class/attribute?

src/Common/Admin/Entities/Link.php
```diff
@@ -27,25 +29,32 @@ class Link extends Base_Entity {
...
 
 	/**
-	 * The text for the link.
+	 * Content for the link if not a string.
 	 *
-	 * @var string
+	 * @var ?Base_Entity
 	 */
-	private string $text = '';
+	private ?Base_Entity $content = null;
```
Camwyn, line 36: I'm not sure what's going on here, but this is confusing. Is it a string?
Can it be a string?

src/admin-views/conditional_content/black-friday.php
```diff
@@ -1,36 +1,37 @@
...
- * @var string $button_link - the url the button links to
+ * @since TBD
+ *
+ * @var string $image_src Where the image is located.
+ * @var string $link      Where the image should link to.
+ * @var string $nonce     The nonce for the dismiss button.
+ * @var string $slug      The slug for the dismiss button.
  */
 
+$year = date_i18n( 'Y' );
```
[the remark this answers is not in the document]
Camwyn: Use `$sale_year` if you care

src/admin-views/conditional_content/black-friday.php
```diff
@@ -1,36 +1,37 @@
...
+/* translators: %1$s: Black Friday year */
```
Camwyn, line 14:
```suggestion
/* translators: %1$s: Black Friday sale year (numeric) */
```
Just clearly indicate it's a number

src/resources/postcss/tribe-common-admin/settings/_sidebar.pcss
```diff
@@ -10,22 +10,30 @@
...
 
 		@media (--viewport-xlarge) {
-			border: 1px solid var(--tec-color-border-default);
-			border-radius: var(--tec-border-radius-default);
 			display: flex;
 			margin: var(--tec-spacer-4) auto auto 0;
 			min-height: fit-content;
+
+
+			> .tec-settings-form__sidebar-section {
```
Camwyn, line 27: Extra line break:
```suggestion

			> .tec-settings-form__sidebar-section {
```

src/resources/postcss/tribe-common-admin/settings/_sidebar.pcss
```diff
@@ -44,37 +52,45 @@
 
 		img {
 			width: 100%;
+			display: block;
```
Camwyn, line 55:
```suggestion
			display: block;
			width: 100%;
```
alpha order

src/resources/postcss/tribe-common-admin/settings/_sidebar.pcss
```diff
@@ -10,22 +10,30 @@
...
+
+			> .tec-settings-form__sidebar-section {
+				display: flex;
+				flex-direction: column;
+				gap: var(--tec-spacer-3);
+				margin: 0 auto auto 0;
+				background-color: var(--tec-color-background);
+				border: 1px solid var(--tec-color-border-default);
+				border-radius: var(--tec-border-radius-default);
+				overflow: hidden;
```
Camwyn, line 35: Alpha order:
```suggestion
				background-color: var(--tec-color-background);
				border: 1px solid var(--tec-color-border-default);
				border-radius: var(--tec-border-radius-default);
				display: flex;
				flex-direction: column;
				gap: var(--tec-spacer-3);
				margin: 0 auto auto 0;
				overflow: hidden;
```

Camwyn (changes requested):

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,191 @@
...
+
+		return add_user_meta( $user_id, $this->meta_key, $this->slug );
+	}
+
+	/**
+	 * Removes the user meta that holds if this content has been dismissed.
+	 *
+	 * @since TBD
+	 *
+	 * @param int|null|string $user_id The user ID
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
	 * @param int|null|string $user_id The user ID.
```

src/Common/Admin/Conditional_Content/Dismissible_Trait.php
```diff
@@ -0,0 +1,191 @@
...
+	protected function dismiss( $user_id = null ): bool {
+		if ( empty( $this->slug ) ) {
+			return false;
+		}
+
+		if ( is_null( $user_id ) ) {
+			$user_id = get_current_user_id();
+		}
+
+		// If this user has dismissed we don't care either
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
		// If this user has dismissed we don't care either.
```

src/Common/Admin/Conditional_Content/Controller.php
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * @since TBD
+	 */
+	public function register() {
+		// This is specifically for the admin, bail if we're not in the admin.
+		if ( ! is_admin() ) {
+			return;
+		}
+
+		$this->container->singleton(  Black_Friday::class, Black_Friday::class, [ 'hook' ] );
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
		$this->container->singleton( Black_Friday::class, Black_Friday::class, [ 'hook' ] );
```

Camwyn (dismissed):

Camwyn (dismissed):

src/Common/Admin/Entities/Button.php
JPry: @Camwyn 

> the parent functions should handle passing empty values

The reason parent doesn't have its own constructor was to allow for the different needs of child elements. But you could easily update them to handle only classes and attributes.

> how do I remove a class/attribute?

What is a scenario where you need to remove a class or attribute?

PR #2232, opened 2024-10-22
Include a way to get request variables without sanatization

bordoni:
### 🎫 Ticket

[TEC-5283]

### 🗒️ Description

Resolves the problem where because we are using `tribe_get_request_var` it will strip all HTML.

### 📹 Artifact

https://i.bordoni.me/Zt3Zk5Bd

[TEC-5283]: https://stellarwp.atlassian.net/browse/TEC-5283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/functions/utils.php
+++ b/src/functions/utils.php
@@ -160,13 +160,13 @@ if ( ! function_exists( 'tribe_get_request_var' ) ) {
 	 *
 	 * @see   tec_get_request_var()
 	 *
-	 * @param string|array $var           The variable to check for.
+	 * @param string|array $request_var   The variable to check for.
 	 * @param mixed        $default_value The default value to return if the variable is not set.
 	 *
 	 * @return mixed
 	 */
-	function tribe_get_request_var( $var, $default_value = null ) {
-		return tec_get_request_var( $var, $default_value );
+	function tribe_get_request_var( $request_var, $default_value = null ) {
+		return tec_get_request_var( $request_var, $default_value );
 	}
 }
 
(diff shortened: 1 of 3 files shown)


dpanta94 (changes requested):
The code is good. I would prefer seeing another method though `tribe_get_request_var_raw` which would be used inside `tribe_get_request_var`, rather than a new parameter.

In most cases where we want a get param, we want to use the `tribe_get_request_var_raw`, if what we want to get needs to be saved or displayed we need `tribe_get_request_var` :D

dpanta94 (approved):

src/functions/utils.php
```diff
@@ -156,36 +156,95 @@
...
 		}
 
 		if ( isset( $_GET ) ) {
-			$requests[] = (array) $_GET;
+			$requests[] = (array) $_GET; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.NonceVerification.Recommended
 		}
 
 		if ( isset( $_POST ) ) {
-			$requests[] = (array) $_POST;
+			$requests[] = (array) $_POST; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.NonceVerification.Recommended
```
Camwyn: Just disable at the beginning of the function and re-enable at the end?

src/functions/utils.php
```diff
@@ -156,36 +156,95 @@ function tribe_exit( $status = '' ) {
...
-			return $default;
+			return $default_value;
 		}
 
-		$unsafe = Tribe__Utils__Array::get_in_any( $requests, $var, $default );
-		return tribe_sanitize_deep( $unsafe );
+		$unsafe = Tribe__Utils__Array::get_in_any( $requests, $request_var, $default_value );
+
+		// Return the value as is.
+		return $unsafe;
```
Camwyn, line 247:
```suggestion
		// Return the value as is.
		return Tribe__Utils__Array::get_in_any( $requests, $request_var, $default_value );
```

Camwyn (changes requested):
Small tweaks and maybe less phpcs noise?

dpanta94 (dismissed):
The code looks good - Why EVA tests are failing though ?

PR #2246, opened 2024-11-01
Ensure a more graceful deprecation of props in Settings class.

bordoni:
### 🎫 Ticket

[TEC-5312]

### 🗒️ Description

Customer came with problems of usage on certain props of Setting class, I found out that we were a bit too aggressive with our deprecation, we need to leave notices before we fully remove.

[TEC-5312]: https://stellarwp.atlassian.net/browse/TEC-5312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changelog/fix-TEC-5312-deprecated-props-still-in-use
+++ b/changelog/fix-TEC-5312-deprecated-props-still-in-use
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Include backwards compatibility for deprecated proprieties in the Settings class used in The Events Calendar and Event Tickets [TEC-5312]


src/Tribe/Settings.php
```diff
@@ -284,9 +284,59 @@ public function __construct() {
...
+		 *
+		 * @deprecated 6.1.0
+		 */
+		$this->menuName    = $this->menu_name;
+		$this->requiredCap = $this->required_cap;
+		$this->allTabs     = $this->all_tabs;
+		$this->defaultTab  = $this->default_tab;
+		$this->currentTab  = $this->current_tab;
+		$this->noSaveTabs  = $this->no_save_tabs;
+		$this->adminSlug   = $this->admin_slug;
```
dpanta94, line 299: Since they are being served by the magic method, why do we set them as well ? If they are set, the magic method won't be triggered and as a result the deprecation notice won't fire. right ?

dpanta94 (changes requested):

src/Tribe/Settings.php
bordoni: Yeah it's intentional for now until we deprecate the usage of those in ETP and ET. But didn't want to add complexity to the fext release

dpanta94 (approved):

dpanta94 (approved):

PR #2251, opened 2024-11-04
Project IAN (In-App Notifications) - Common Client

pattihis:
### 🎫 Ticket

[TEC-5165]

### 🗒️ Description

IAN is a notification system that will remotely communicate with a laravel app to send notifications to a website running TEC/ET. 

Screenshot_1

### ✔️ Checklist

[TEC-5165]: https://stellarwp.atlassian.net/browse/TEC-5165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Notifications/Notifications.php
+++ b/src/Common/Notifications/Notifications.php
@@ -125,20 +125,20 @@ final class Notifications {
 		}
 
 		/**
-		 * Filter allowing disabling of the Notifications icon by returning false.
+		 * Filter allowing disabling of the Notifications by returning false.
 		 *
 		 * @since TBD
 		 *
-		 * @param bool $show Whether to show the icon or not.
+		 * @param bool $show Whether to render the IAN sidebar or not.
 		 */
-		$show = (bool) apply_filters( 'tec_common_ian_show_icon', true, $slug );
+		$show = (bool) apply_filters( 'tec_common_ian_render', true, $slug );
 
 		if ( ! $show ) {
 			return;
 		}
 
 		$template = new Template();
-		$template->render_icon( [ 'slug' => $slug ], true );
+		$template->render_sidebar( [ 'slug' => $slug ], true );
 	}
 
 	/**
--- a/src/Common/Notifications/Template.php
+++ b/src/Common/Notifications/Template.php
@@ -35,16 +35,16 @@ class Template extends Tribe__Template {
 	}
 
 	/**
-	 * Render the notification icon.
+	 * Render the notification sidebar.
 	 *
 	 * @since TBD
 	 *
 	 * @param array $args Array of arguments that will ultimately be sent to the template.
 	 * @param bool  $output Whether or not to echo the HTML. Defaults to true.
 	 *
-	 * @return string HTML of notification icon.
+	 * @return string HTML of notification sidebar.
 	 */
-	public function render_icon( $args, $output = true ) {
+	public function render_sidebar( $args, $output = true ) {
 		$args = wp_parse_args(
 			$args,
 			[
@@ -55,7 +55,7 @@ class Template extends Tribe__Template {
 			]
 		);
 
-		return $this->template( 'icon', $args, $output );
+		return $this->template( 'sidebar', $args, $output );
 	}
 
 	/**
--- a/src/Tribe/Settings.php
+++ b/src/Tribe/Settings.php
@@ -13,6 +13,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 use TEC\Common\Admin\Entities\Element_With_Children;
 use TEC\Common\Admin\Entities\Field_Wrapper;
 use Tribe\Admin\Pages as Admin_Pages;
+use TEC\Common\Notifications\Controller;
 
 if ( did_action( 'tec_settings_init' ) ) {
 	return;
@@ -590,10 +591,15 @@ class Tribe__Settings {
 	public function do_page_header( $admin_page ): void {
 		?>
 		<h1>
+			<span>
 			<?php if ( $this->is_event_settings() ) : ?>
 				<?php echo wp_kses_post( $this->get_page_logo( $admin_page ) ); ?>
 			<?php endif; ?>
 			<?php echo esc_html( $this->get_page_title( $admin_page ) ); ?>
+			</span>
+			<?php if ( tribe( Controller::class )->is_ian_page() ) : ?>
+				<div class="ian-client" data-trigger="iconIan"></div>
+			<?php endif; ?>
 		</h1>
 		<?php
 	}
--- a/src/admin-views/notifications/sidebar.php
+++ b/src/admin-views/notifications/sidebar.php
@@ -11,9 +11,6 @@
  */
 
 ?>
-
-<div class="ian-client" data-trigger="iconIan"></div>
-
 <div class="ian-sidebar is-hidden" data-trigger="sideIan">
 	<div class="ian-sidebar__title">
 		<div class="ian-sidebar__title--left">
--- a/src/resources/js/ian-client.js
+++ b/src/resources/js/ian-client.js
@@ -12,12 +12,12 @@
 		Ian.feed = { read: [], unread: [] };
 
 		const init = () => {
+			wrapHeadings();
+
 			document.addEventListener("click", handleClick);
 			document.addEventListener("keydown", handleKeydown);
 
 			if (Ian.consent == "true") getIan();
-
-			wrapHeadings();
 		};
 
 		const wrapHeadings = () => {
@@ -28,11 +28,25 @@
 					const wrapper = document.createElement("div");
 					wrapper.className = "ian-header";
 
+					const innerWrapper = document.createElement("div");
+					innerWrapper.className = "ian-inner-wrapper";
+
+					const clientDiv = document.createElement("div");
+					clientDiv.className = "ian-client";
+					clientDiv.setAttribute("data-trigger", "iconIan");
+
 					heading.parentNode.insertBefore(wrapper, heading);
-					wrapper.appendChild(heading);
-					wrapper.appendChild(pageAction);
+					wrapper.appendChild(innerWrapper);
+					innerWrapper.appendChild(heading);
+					innerWrapper.appendChild(pageAction);
+					innerWrapper.appendChild(clientDiv);
 				}
 			});
+			Ian.icon = document.querySelector('[data-trigger="iconIan"]');
+			const settingsHeading = document.querySelector('.tribe_events_page_tec-events-settings h1');
+			if (settingsHeading) {
+				settingsHeading.appendChild(Ian.sidebar);
+			}
 		};
 
 		const handleClick = e => {
--- a/src/resources/postcss/ian-client.pcss
+++ b/src/resources/postcss/ian-client.pcss
@@ -1,10 +1,5 @@
 .ian-client {
-  position: absolute;
-  top: var(--tec-spacer-6);
-  left: calc(100% - var(--tec-spacer-12));
-
   width: 40px;
-
   height: 40px;
 
   background-image: url('../images/icons/bell.svg');
@@ -31,8 +26,8 @@
 }
 
 .ian-sidebar {
-  position: fixed;
-  top: 165px;
+  position: absolute;
+  top: 125px;
   right: 0;
 
   z-index: 1000;
@@ -47,12 +42,8 @@
 
   transition: right 0.5s ease-in-out;
 
-  @media screen and (max-width: 782px) {
-    top: 180px;
-  }
-
   @media screen and (max-width: 500px) {
-    top: 140px;
+    top: 86px;
   }
 
   &.is-hidden {
@@ -281,37 +272,67 @@
   }
 }
 
+.tec-events-admin-settings {
+
+  h1 {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+  }
+}
+
 .post-php.post-type-tribe_events,
 .edit-php.post-type-tribe_events {
   .ian-header {
     height: 60px;
     background: #fff;
-    width: calc(100% + 40px);
+    width: 100%;
     margin-left: -20px;
     margin-top: -10px;
     margin-bottom: 16px;
-    padding: 36px 0 0 20px;
-  }
-
-  .ian-sidebar {
-    top: 130px;
+    padding: 36px 20px 0 20px;
 
     @media screen and (max-width: 782px) {
-      top: 144px;
-      right: 24px;
+      height: 80px;
+    }
+  }
 
-      &.is-hidden {
-        right: -100%;
+  .ian-inner-wrapper {
+    clear: both;
+
+    h1.wp-heading-inline {
+      padding: 4px 0;
+
+      @media screen and (max-width: 782px) {
+        margin-bottom: 0;
       }
     }
   }
 
   .ian-client {
-    left: calc(100% - var(--tec-spacer-12));
-    top: var(--tec-spacer-8);
+    float: right;
+  }
+
+  .ian-sidebar {
+    top: 98px;
+
+    @media (min-width: 600px) and (max-width: 782px) {
+      top: 118px;
+      right: 4px;
+
+      &.is-hidden {
+        right: -100%;
+      }
+    }
 
     @media screen and (max-width: 600px) {
-      top: calc(var(--tec-spacer-8)*2.25);
+      top: 164px;
+      right: 4px;
+
+      &.is-hidden {
+        right: -100%;
+      }
     }
   }
 


src/Common/Ian/Conditionals.php
```diff
@@ -0,0 +1,106 @@
...
+						}
+					}
+				}
+
+				return true;
+			}
+		);
+
+		return array_values( $notifications );
+	}
```
bordoni, line 105: I would like to see this method split into a couple of classes, with an Abstract and Interface, so that we can extend and add more conditionals without bloating this method further.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+use TEC\Common\Admin\Conditional_Content\Dismissible_Trait;
+
+/**
+ * Class Ian_Client
+ *
+ * @since   TBD
+
+ * @package TEC\Common\Ian
+ */
+final class Ian_Client {
```
bordoni, line 23: This class is doing too much, lets the individual Notifications into a class that extends the Dismissable Trait.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+	/**
+	 * Boot the IAN Client.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function boot(): void {
+		// TODO - Get the organization, brand, and product from... where?
+		$this->api_url = 'https://ian.stellarwp.com/feed/organization/brand/product.json';
```
bordoni, line 63: Move the handling this definition of class to a method with filters.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+	 */
+	public function init(): void {
+		/**
+		 * Filter the base parent slugs for IAN.
+		 *
+		 * @since TBD
+		 *
+		 * @param array<string> $plugin_slugs The slugs for plugins that support IAN.
+		 */
+		$this->plugin_slugs = apply_filters( 'tec_common_ian_plugin_slugs', $this->plugin_slugs );
```
bordoni, line 91: Move to a separate method.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+	 * @return bool
+	 */
+	public function is_ian_page() {
+		$screen = get_current_screen();
+
+		if ( in_array( $screen->id, [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings' ], true ) ) {
+			return true;
+		}
+
+		return false;
```
bordoni, line 148: This needs to be filterable.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+		load_template(
+			$main->plugin_path . 'src/admin-views/ian/icon.php',
+			true,
+			[
+				'slug'  => $slug,
+				'main'  => $main,
+				'optin' => Conditionals::get_ian_opt_in(),
+				'url'   => Telemetry::get_permissions_url(),
+			]
+		);
```
bordoni, line 235: Use our own template classes.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+
+		if ( wp_verify_nonce( $nonce, 'common_ian_nonce' ) ) {
+
+			tribe_update_option( 'ian-client-opt-in', 1 );
+
+			wp_send_json_success( 'IAN opt-in successful', 200 );
+		} else {
+			wp_send_json_error( 'Invalid nonce', 403 );
+		}
+	}
```
bordoni, line 254: Translations are missign on these errors.

src/Common/Ian/Ian_Client.php
```diff
@@ -0,0 +1,386 @@
...
+				wp_send_json_error( 'Invalid slug', 403 );
+			}
+
+			$this->slug = $slug;
+
+			wp_send_json_success( $this->dismiss(), 200 );
+		} else {
+			wp_send_json_error( 'Invalid nonce', 403 );
+		}
+	}
```
bordoni, line 385: We need to have one nonce per notification, attach the nonce to the dimiss button itself.

src/Common/Ian/Provider.php
```diff
@@ -0,0 +1,132 @@
...
+use TEC\Common\Contracts\Service_Provider;
+
+/**
+ * Class Provider
+ *
+ * @since   TBD
+
+ * @package TEC\Common\Ian
+ */
+class Provider extends Service_Provider {
```
bordoni, line 21: Turn this into a Controller, we should avoid pure providers.

bordoni (changes requested):
Overall, there are a ton of methods here that are namespaced with IAN, when they are already in a class in the Ian Namespace, we should avoid that.

I would rename it to just `Notifications` as the namespace.

* Make sure to use `tec_get_request_var()` instead of `$_POST` variable.
* Splitting concerns into multiple classes.
* Add tests to the classes added.

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,347 @@
...
+	/**
+	 * Get the API URL for the In-App Notifications.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_api_url() {
+		$api = 'https://ian.stellarwp.com/feed/organization/brand/product.json';
+
```
jesseeproductions, line 129: What about setting this from The Events Calendar ( and Event Tickets or any of the premium plugins ). In TEC it would also have a filter to be able to easily target the url for that plugin's feed.

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,347 @@
...
+		/**
+		 * Filter the API URL for the In-App Notifications.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $api The API URL for the In-App Notifications.
+		 */
+		$api = apply_filters( 'tec_common_ian_api_url', $api );
```
jesseeproductions, line 137: Can we also add some parameters to this filter so someone would have more information on the source of the call? For instance trying to see the product calling it? This might not be necessary if we decide to implement the above.

src/Common/Notifications/Notifications.php
pattihis: Yes, it could [go here](https://github.com/the-events-calendar/the-events-calendar/pull/4798/files#diff-def87422643db7c06d83ff1115ce16c359b9d15d3f9155527c7fcc7d1650be6aR29), where we also have a function to decide which screen we are showing the notifications in. I am waiting for some decisions from @bordoni for that.

Thanks @jesseeproductions .

src/Common/Notifications/Notifications.php
pattihis: Indeed, I believe @Camwyn told me before to always pass `$this` as the last param because it will probably be needed down the road and to add context to the filter. Thanks for pointing it out, I will change that.

src/Common/Notifications/Controller.php
```diff
@@ -0,0 +1,131 @@
...
+ */
+class Controller extends Provider_Contract {
+
+	/**
+	 * Registers actions and filters.
+	 *
+	 * @since TBD
+	 */
+	public function register() {
+		$this->container->bind( Notifications::class, Notifications::class );
```
dpanta94, line 29: You just need to pass one parameter if the second is the same as the first.

dpanta94 (changes requested):
In general id prefer to see a Main controller registering individual smaller controllers. Smaller controllers on their own register hooks and unregister them as well.

I think the architecture is not very straight forward. It can be easy to confuse somebody. Would really prompt you to have a look on how we are doing things for Seating.

In ET check out the directory src/Tribe/Tickets/Seating/. You can see the main controller in that directory and then each one individually what they are offering to the application.

src/Common/Notifications/Controller.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Hook the IAN initialization.
+	 *
+	 * @since TBD
+	 */
+	public function hook_init(): void {
+		add_action( 'admin_init', [ $this, 'initialize' ], 5 );
+	}
```
dpanta94, line 76: There is no initialize method - you can remove this hook completely.

src/Common/Notifications/Controller.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->remove_actions();
+	}
+
+	/**
+	 * Add the action hooks.
+	 *
+	 * @since TBD
+	 */
+	public function add_actions() {
+		add_action( 'tribe_plugins_loaded', [ $this, 'boot' ], 50 );
```
dpanta94, line 45: What i see that this hook is trying to - which is the only thing that needs to be done in a hook is register assets.

So lets do that instead of booting. Also i did dig into when the boot method is called and its called when the whole app is being booted.

It goes like all the registration happens then when the container calls boot all the providers overloading the container call their boot method as well. Its not usefull in frameworks like WP, because iduring registration you can simply add a hook to execute later (when you want) a callback.

src/Common/Notifications/Controller.php
```diff
@@ -0,0 +1,128 @@
...
+	/**
+	 * Boot our internal IAN code.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function boot() {
+		$this->container->make( Notifications::class );
+	}
```
dpanta94, line 87: As a result i would rename the method - to avoid confusion with the containers boot method being called during the requests lifecycle - and here only register the assets Notifications provide.

No need for the assets to be registered inside Notifications either way, since Notifications is more a business logic interface.

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,276 @@
...
+				'localize'     => [
+					'name' => 'commonIan',
+					'data' => [
+						'ajax_url' => admin_url( 'admin-ajax.php' ),
+						'nonce'    => wp_create_nonce( 'common_ian_nonce' ),
+					],
+				],
+			]
+		);
+	}
```
dpanta94, line 100: So move this registration in the Controller

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,276 @@
...
+
+	/**
+	 * Get the IAN notifications.
+	 *
+	 * @since TBD
+	 */
+	public function get_feed() {
+		if ( ! wp_verify_nonce( tec_get_request_var( 'nonce' ), 'common_ian_nonce' ) ) {
+			wp_send_json_error( esc_html__( 'Invalid nonce', 'tribe-common' ), 403 );
```
dpanta94, line 226: To help you with testing, you will need to add a `return` after every wp_send_json which is bailing early.

Thats because you will need to mock these functions in order to test them to avoid their exit.

I will skip making this comment everywhere its applicable :D

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,276 @@
...
+		}
+		// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get
+		$response = wp_remote_get( $this->api_url );
```
dpanta94, line 229: How often is this being called ? Depending on the answer some caching may be of value.

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,276 @@
...
+		if ( is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) !== 200 ) {
+			wp_send_json_error( wp_remote_retrieve_response_message( $response ), wp_remote_retrieve_response_code( $response ) );
+		}
+
+		$body = json_decode( wp_remote_retrieve_body( $response ), true );
+		$feed = Conditionals::filter_feed( $body['notifications_by_area']['general-tec'] );
```
dpanta94, line 235: Use `??` operator for accessing $body's offsets. That may or may not be set

src/Common/Notifications/Notifications.php
```diff
@@ -0,0 +1,276 @@
...
+	public function opt_in() {
+		if ( wp_verify_nonce( tec_get_request_var( 'nonce' ), 'common_ian_nonce' ) ) {
+
+			tribe_update_option( 'ian-notifications-opt-in', 1 );
+
+			wp_send_json_success( esc_html__( 'Notifications opt-in successful', 'tribe-common' ), 200 );
+		} else {
+			wp_send_json_error( esc_html__( 'Invalid nonce', 'tribe-common' ), 403 );
+		}
+	}
```
dpanta94, line 217: Lets transform this one as well to early return style avoiding the if else as a result.

src/Common/Notifications/Template.php
```diff
@@ -0,0 +1,105 @@
...
+		if ( empty( $this->template ) ) {
+			$this->template = new Tribe__Template();
+			$this->template->set_template_origin( Tribe__Main::instance() );
+			$this->template->set_template_folder( 'src/admin-views/notifications' );
+			$this->template->set_template_context_extract( true );
+			$this->template->set_template_folder_lookup( false );
+		}
+
+		return $this->template;
+	}
```
dpanta94, line 51:
```suggestion
	/**
	 * Notifications Template Constructor
	 *
	 * @since TBD
	 */
	private function __construct() {
			$this->set_template_origin( Tribe__Main::instance() );
			$this->set_template_folder( 'src/admin-views/notifications' );
			$this->set_template_context_extract( true );
			$this->set_template_folder_lookup( false );
		}
	}
```

Since we are extedning we can use the methods of the parent. Make sure you make this a singleton instance so the container builds it only once.

src/Common/Notifications/Template.php
```diff
@@ -0,0 +1,105 @@
...
+			[
+				'slug'  => $args['slug'],
+				'main'  => Tribe__Main::instance(),
+				'optin' => Conditionals::get_opt_in(),
+				'url'   => Telemetry::get_permissions_url(),
+			]
+		);
+
+		$template = $this->get_template();
+		return $template->template( 'icon', $args, $output );
```
dpanta94, line 75:
```suggestion
		return $this->template( 'icon', $args, $output );
```

src/Common/Notifications/Template.php
```diff
@@ -0,0 +1,105 @@
...
+				'dismissible' => $args['dismissible'] ?? true,
+				'slug'        => $args['slug'] ?? '',
+				'title'       => $args['title'] ?? '',
+				'html'        => $args['html'] ?? '',
+				'actions'     => $args['actions'] ?? [],
+			]
+		);
+
+		$template = $this->get_template();
+		return $template->template( 'notification', $args, $output );
```
dpanta94, line 103:
```suggestion
		return $this->template( 'notification', $args, $output );
```

dpanta94 (changes requested):
A lot of improvements! Kudos @pattihis.

I also want to make sure that when we are responding to AJAX requests we are not checking who is the current user (e.g. `current_user_can()`). I want to make sure thats expected.

I think

- amending my commnets.
- writting tests
- maybe one more round of review after it

and we are ready to merge.

src/Common/Notifications/Conditionals.php
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	public static function check_plugin_version( array $plugins ): bool {
+		if ( empty( $plugins ) ) {
+			return true;
+		}
+
+		foreach ( $plugins as $plugin ) {
+			$pieces = explode( '@', $plugin );
+
+			if ( ! is_plugin_active( $pieces[0] . '/' . $pieces[0] . '.php' ) ) {
```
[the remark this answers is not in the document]
jesseeproductions: This is a good point. What about using Tribe__Dependency to compare versions and plugin names? All that information is registered on plugin loading and this way we do not have to guess on plugin paths.

`array (
  'Tribe__Events__Main' => 
  array (
    'class' => 'Tribe__Events__Main',
    'version' => '6.7.1',
    'path' => '/Users/brianjessee/Local Sites/tribe/events/app/public/wp-content/plugins/the-events-calendar/the-events-calendar.php',
    'dependencies' => 
    array (
      'addon-dependencies' => 
      array (
        'Tribe__Events__Pro__Main' => '7.1.0-dev',
        'Tribe__Events__Filterbar__View' => '5.5.7-dev',
        'Tribe__Events__Community__Main' => '5.0.4-dev',
        'Tribe__Tickets__Main' => '5.10.0-dev',
        'Tribe__Tickets_Plus__Main' => '5.9.0-dev',
        'Tribe__Events__Tickets__Eventbrite__Main' => '4.6.14-dev',
        'Tribe__Events__Community__Tickets__Main' => '4.9.3-dev',
        'Tribe\\Events\\Virtual' => '1.15.5-dev',
        'TEC\\Event_Automator' => '1.3.1-dev',
      ),
    ),
  ),
  'Tribe__Events__Filterbar__View' => 
  array (
    'class' => 'Tribe__Events__Filterbar__View',
    'version' => '5.5.7',
    'path' => '/Users/brianjessee/Local Sites/tribe/events/app/public/wp-content/plugins/events-filterbar/the-events-calendar-filter-view.php',
    'dependencies' => 
    array (
      'parent-dependencies' => 
      array (
        'Tribe__Events__Main' => '6.7.0-dev',
      ),
    ),
  ),
  'Tribe__Events__Pro__Main' => 
  array (
    'class' => 'Tribe__Events__Pro__Main',
    'version' => '7.2.0',
    'path' => '/Users/brianjessee/Local Sites/tribe/events/app/public/wp-content/plugins/events-pro/events-calendar-pro.php',
    'dependencies' => 
    array (
      'parent-dependencies' => 
      array (
        'Tribe__Events__Main' => '6.7.0-dev',
      ),
    ),
  ),
)`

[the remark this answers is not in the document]
jesseeproductions: We could send class names in the notification conditions from Laravel since we know those are consistent across all installs.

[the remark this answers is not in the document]
bordoni: Passing classes creates a weird association on IAN. I know it's a good way of solving it, but I am not sure it's ideal, I would love if we could use a standard way in WordPress.

dpanta94 (changes requested):
Final change! All the rest looks good :D

bordoni (approved):

dpanta94 (approved):

PR #2268, opened 2024-11-14
React 18 compatibility

bordoni:
Ticket: [TEC-5322](https://stellarwp.atlassian.net/browse/TEC-5322)
Update asset version, registration and usage to align with React 18.

[TEC-5322]: https://stellarwp.atlassian.net/browse/TEC-5322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package-safelist.json
```diff
@@ -50,5 +50,7 @@
   "vendor/vendor-prefixed/autoload-classmap.php",
   "vendor/vendor-prefixed/autoload.php",
   "vendor/vendor-prefixed/stellarwp/**/*",
+  "node_modules/php-date-formatter/js/php-date-formatter.js",
+  "node_modules/php-date-formatter/js/php-date-formatter.min.js",
```
dpanta94, line 54: Can we include those outside of node_modules ?

src/Tribe/Main.php
```diff
@@ -258,6 +258,7 @@ public function load_assets() {
 				[ 'tribe-attrchange', 'vendor/attrchange/js/attrchange.js' ],
 				[ 'tec-ky-module', 'vendor/ky/ky.js', [], null, [ 'module' => true ] ],
 				[ 'tec-ky', 'vendor/ky/tec-ky.js', [ 'tec-ky-module' ], null, [ 'module' => true ] ],
+				[ 'tec-common-php-date-formatter', 'node_modules/php-date-formatter/js/php-date-formatter.js' ],
```
dpanta94, line 261: the file source would need to be replaced here as well

dpanta94 (changes requested):
Just a file movement in order for our final zips to include `php-date-formatter.js`.

Everything else looks great!

package-safelist.json
lucatume: I think the build system should support this, changing the way dependencies are  managed, in `node_modules` and through `npm` to go with something that is fixed (in `vendor` or `includes`) seems working around a bug.

Furthermore, this is both included directly, and imported in the blocks.

Could the build system be fixed?

PR #2269, opened 2024-11-14
Help Hub - DQA 44,29,23,11

redscar:
### 🗒️ Description
Fixes -

- Consent container is not clickable in the AI Chatbot area
- Open accordion separator is in the wrong place
- In Resources content is too indented on the left and not enough on the right
- Resources is missing section headings on mobile
https://docs.google.com/spreadsheets/d/19_VNSbR426WqHBNS9DghOuyQsua5ZO_eTDgWg9JZco8/edit?gid=0#gid=0

### 🎥 Artifacts 

image
image
image

### DIFF
--- a/src/resources/js/admin/help-page.js
+++ b/src/resources/js/admin/help-page.js
@@ -109,6 +109,7 @@ tribe.helpPage = tribe.helpPage || {};
 			{
 				active: true,
 				collapsible: true,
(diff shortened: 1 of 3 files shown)


Camwyn (dismissed):
Don't need to change but it does mean a smaller CSS file

Camwyn (approved):

PR #2307, opened 2024-12-02
[TECTRIA-549] Help Hub - Valid License Transient Issue

redscar:
### 🎫 Ticket

[TECTRIA-549]

### 🗒️ Description

When installing a free plugin and going to Help Hub `is_any_license_valid` would cache that there were no valid licenses. Once installing a valid license `is_any_license_valid` would continue to return false.

By deleting the transient when we check a plugins status we can confirm that `is_any_license_valid` always gives us the most up to date value.

We also took into account Uplink, Dimi introduced a new hook that will fire when a plugin is licensed.

### ✔️ Checklist

[TECTRIA-549]: https://stellarwp.atlassian.net/browse/TECTRIA-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c62c55b2f96b21c6942947c11925cc1f",
+    "content-hash": "4a063d7c5ba67c09c3ff597cc59e3620",
     "packages": [
         {
             "name": "firebase/php-jwt",
@@ -782,16 +782,16 @@
         },
         {
             "name": "stellarwp/uplink",
-            "version": "v2.2.1",
+            "version": "v2.2.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/stellarwp/uplink.git",
-                "reference": "57eb6264994662cb8da368cc3a9227ec716edc70"
+                "reference": "5bc1f115efe629dd4244bff08809aea45ed9d8f1"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/stellarwp/uplink/zipball/57eb6264994662cb8da368cc3a9227ec716edc70",
-                "reference": "57eb6264994662cb8da368cc3a9227ec716edc70",
+                "url": "https://api.github.com/repos/stellarwp/uplink/zipball/5bc1f115efe629dd4244bff08809aea45ed9d8f1",
+                "reference": "5bc1f115efe629dd4244bff08809aea45ed9d8f1",
                 "shasum": ""
             },
             "require": {
@@ -843,9 +843,9 @@
             "description": "A library that integrates a WordPress product with the StellarWP Licensing system.",
             "support": {
                 "issues": "https://github.com/stellarwp/uplink/issues",
-                "source": "https://github.com/stellarwp/uplink/tree/v2.2.1"
+                "source": "https://github.com/stellarwp/uplink/tree/v2.2.2"
             },
-            "time": "2024-09-30T22:46:40+00:00"
+            "time": "2024-12-05T19:13:16+00:00"
         }
     ],
     "packages-dev": [
@@ -926,15 +926,15 @@
             "type": "project",
             "extra": {
                 "autotagger": true,
+                "mirror-repo": "Automattic/jetpack-changelogger",
                 "branch-alias": {
                     "dev-trunk": "4.2.x-dev"
                 },
-                "mirror-repo": "Automattic/jetpack-changelogger",
-                "version-constants": {
-                    "::VERSION": "src/Application.php"
-                },
                 "changelogger": {
                     "link-template": "https://github.com/Automattic/jetpack-changelogger/compare/${old}...${new}"
+                },
+                "version-constants": {
+                    "::VERSION": "src/Application.php"
                 }
             },
             "autoload": {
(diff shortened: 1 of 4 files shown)


src/Tribe/PUE/Checker.php
```diff
@@ -244,51 +244,58 @@ public function is_key_valid() {
...
-						break;
-					}
+			// Revalidate if we haven't found a valid license yet.
+			foreach ( self::$instances as $checker ) {
+				$license  = get_option( $checker->get_license_option_key() );
+				$response = $checker->validate_key( $license );
+				// Is it valid?
+				if ( ! empty( $response['status'] ) ) {
+					set_transient( self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY, $valid_slug, HOUR_IN_SECONDS );
+					return true;
```
dpanta94, line 291: Combine the 2 foreach into 1 :D

src/Tribe/PUE/Checker.php
```diff
@@ -345,6 +353,8 @@ public function set_key_status( $valid ) {
 			// We set a transient in addition to an option for compatibility reasons.
 			// @todo remove transient in a major feature release where we release all plugins.
 			set_transient( $this->pue_key_status_transient_name, $status, $this->check_period * HOUR_IN_SECONDS );
+
+			delete_transient( self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY );
```
dpanta94, line 357: Lets save resources - at this point we can say with confidence if we have one valid or not. So if its valid don't delete rather update. If it's invalid delete.

dpanta94 (changes requested):

src/Tribe/PUE/Checker.php
redscar: I made the change. My original thought process behind always deleting it was to make sure that `is_any_license_valid` is in sync with the licenses. For example, what if `set_key_status` returns false, but we still have `self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY` set to valid?

codingmusician (dismissed):

src/Tribe/PUE/Checker.php
```diff
@@ -345,6 +350,10 @@ public function set_key_status( $valid ) {
 			// We set a transient in addition to an option for compatibility reasons.
 			// @todo remove transient in a major feature release where we release all plugins.
 			set_transient( $this->pue_key_status_transient_name, $status, $this->check_period * HOUR_IN_SECONDS );
+
+			if ( ! tribe_is_truthy( $valid ) ) {
+				delete_transient( self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY );
+			}
```
dpanta94, line 356: I think we care about being aware when a license is valid mostly not when invalid.

What is the downside if the transient is falsy saying there is one valid while there isn't ?

Also make sure you `set_transient( self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY, $status, HOUR_IN_SECONDS );` when the license is valid.

dpanta94 (changes requested):

src/Tribe/PUE/Checker.php
dpanta94: I think we can respond to that by knowing what would go wrong if there are no valid licenses but the transient still says there is one valid.

src/Tribe/PUE/Checker.php
```diff
@@ -388,6 +404,8 @@ public function hooks() {
 			add_action( 'admin_enqueue_scripts', [ $this, 'maybe_display_json_error_on_plugins_page' ], 1 );
 			add_action( 'admin_init', [ $this, 'general_notifications' ] );
 
+			add_action( 'wp_loaded', [ $this, 'monitor_uplink_actions' ] );
```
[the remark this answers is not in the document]
dpanta94: This hook is simply adding a listener to every plugin registered through uplink to perform an action if ONLY a NEW license is provided. Nothing more.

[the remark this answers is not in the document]
dpanta94: Check https://github.com/stellarwp/uplink/pull/87

the hook only fires if a new connection occurs

dpanta94:
Since we added the generic hook, lets bumb uplink requirements to 2.2.2 and use that ;) https://github.com/stellarwp/uplink/pull/92

dpanta94:
Also definiely not load on FE and if you can hook very specifically even better !

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c62c55b2f96b21c6942947c11925cc1f",
+    "content-hash": "4a063d7c5ba67c09c3ff597cc59e3620",
```
dpanta94, line 7: It seems like the lock file has been updated by running `composer update`. That shouldn't happen since other dependencies are being updated like that. You should revert your lock file and then execute `composer update --lock` so that lock gets updated only by `.json` updates.

src/Tribe/PUE/Checker.php
```diff
@@ -244,51 +246,60 @@ public function is_key_valid() {
...
+
+			if ( $transient_value === $invalid_slug ) {
+				// Transient explicitly marked as invalid, no recheck.
+				return false;
 			}
 
-			// Check our local transient/cache first.
+			if ( self::has_valid_uplink_license() ) {
+				set_transient( self::IS_ANY_LICENSE_VALID_TRANSIENT_KEY, $valid_slug, HOUR_IN_SECONDS );
+				return true;
```
dpanta94, line 273: Because we are checking now if did_action that could only be true when=>

- Current request updated an uplink license
- is_any_license_valid is being executed after license validation - which i dont think is the case

So i would prefer to rather add a listener for the uplink license updated rather than checking if it was done prior to this.

dpanta94 (changes requested):

src/Tribe/PUE/Checker.php
```diff
@@ -2075,5 +2125,23 @@ public function is_valid_key_format() {
 
 			return true;
 		}
+
+		/**
+		 * Monitors Uplink and hooks into the plugin connected action.
+		 *
+		 * This method loops through the registered Uplink plugins and hooks into the
```
dpanta94, line 2132: Description is no longer accurate

tests/integration/Tribe/PUE/Checker_Test.php
```diff
@@ -129,7 +146,329 @@ public function it_should_check_uplink_before_pue() {
 		$this->assertEquals( $key, $pue_instance->get_key() );
 	}
 
-	public function test_replacemnt_key_update_in_multisite_context(): void {
+	public function test_replacemnt_key_update_in_multisite_context(): void {}
```
dpanta94, line 149: Lets either delete it or write a test about it.

dpanta94 (approved):
Approved with minor comments

dpanta94 (approved):

PR #2321, opened 2024-12-11
Fix/uplink activation of plugins

dpanta94:
When a resource is found registered in uplink also check its key against uplink. Not just PUE.

Also adds a listener for uplink successfull connections in order to clear notices related to invalid/expired licenses.

### DIFF
--- a/changelog/fix-uplink-activation-of-plugins
+++ b/changelog/fix-uplink-activation-of-plugins
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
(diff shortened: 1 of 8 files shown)


redscar (commented):
Look's good. I think we should add some tests for `is_key_valid()` to make sure that the uplink piece works. 

If you want me to write the tests, let me know. I've recently became familiarized with the `Checker.php` code :laughing:

bordoni (approved):

rafsuntaskin (approved):

PR #2338, opened 2025-01-06
Provider registration actions should fire once or not at all when provider is inactive

dpanta94:
### 🎫 Ticket

[TCMN-178]

### 🗒️ Description

Container would always fire registration actions of controllers for every time they would register and even if they were inactive.

As a result the following scenario would result in a fatal:

In ET -> Deactivate the main seating controller using a filter or const making the is_active method return false.
In ETP -> The main seating controller registers on the registration action of the ET's main seating controller.
We would have a fatal because the ET's main seating controller registration action would fire even when the controller is inactive.

### 🎥 Artifacts 
Very low level change - making tests passing a valid artifact.

### ✔️ Checklist

[TCMN-178]: https://stellarwp.atlassian.net/browse/TCMN-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Contracts/Container.php
```diff
@@ -32,28 +34,37 @@ public function get( $id ) {
...
 	 * @return void
 	 *
 	 * @throws \TEC\Common\lucatume\DI52\ContainerException If the provider class is marked as deferred but
 	 *                                                      does not provide a set of deferred registrations.
 	 */
-	public function register( $serviceProviderClass, ...$alias ) {
-		// Register the provider with the parent container.
-		parent::register( $serviceProviderClass, ...$alias );
+	public function register( $service_provider_class, ...$alias ) {
+		try {
```
lucatume, line 48: Throwing exceptions in the `Controller` seems risky since it introduces possible crashes where there were none before.
An idem-potent function (`register`) now becomes a careful-how-you-use-it function.

The purpose of this fix is to avoid firing the registration actions if the Controller (this is only about Controllers) is already registered or not active.

I think this could be achieved modifying only this class with code like this one:

```php
public function register($service_provider_class, ...$alias ) {
	// Register the provider with the parent container.
	parent::register( $service_provider_class, ...$alias );

	if (
		is_a($service_provider_class, Controller::class, true)
		&& !$this->getVar($service_provider_class . '_registered', false)
	) {
		// The Controller did not register (inactive, already registered): do not fire the registration actions.
		return;
	}

	/**
	 * Fires when a service provider is registered by the container.
	 *
	 * @param string        $service_provider_class The service provider class name.
	 * @param array<string> $alias                The alias(es) the service provider was registered with.
	 *@since 5.1.4
	 *
	 */
	do_action( 'tec_container_registered_provider', $service_provider_class, $alias );

	/**
	 * Fires a class-specific action when a service provider is registered by the container.
	 *
	 * @since 5.1.4
	 *
	 * @param array<string> $alias The alias(es) the service provider was registered with.
	 */
	do_action( 'tec_container_registered_provider_' . $service_provider_class, $alias );

	if (
		// Back compat with older definition of Service Provider.
		! property_exists( $service_provider_class, 'registration_action' )
		// New definition of Service Provider: default action is empty.
		|| empty( $service_provider_class::$registration_action )
	) {
		return;
	}

	/**
	 * Fires a custom action defined by the Service Provider when it's registered.
	 *
	 * @since 5.1.4
	 */
	do_action( $service_provider_class::$registration_action, $service_provider_class, $alias );
}
```

This would avoid introducing the potentially destructive exception throwing in the code.
What do you think?

lucatume: Another check that might be done could be with `did_action`.

lucatume (approved):

PR #2351, opened 2025-01-17
FBAR-341 Fix Filter Bar translations

andrasguseo:
### 🎫 Ticket

[FBAR-341]

### 🗒️ Description

After WP 6.7 the translations from the `lang` folder of Filter Bar would not load. This PR aims to fix that.

### 🎥 Artifacts 
[Screen Recording](https://docs.google.com/videos/d/1mrJF9W8ODlc4XvMmov0FD1SSAf4Fv-Qkb0bwV6cY2Bo/edit?usp=sharing)
- Part 1 - Explaining the original issue and the issue with the first proposal.
- Part 2 - Recommending a possible solution.
- Part 3 - Demonstrating that the solution works.

### ✔️ Checklist

[FBAR-341]: https://stellarwp.atlassian.net/browse/FBAR-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Main.php
```diff
@@ -585,9 +586,14 @@ public function load_text_domain( $domain, $dir = false ) {
...
 
+		// Try to load translations from wp-content/languages/plugins
 		$loaded = load_plugin_textdomain( $domain, false, $plugin_rel_path );
 
-		if ( $dir !== false && ! $loaded ) {
+		// If translation files are not found in wp-content/languages/plugins, then load from the plugin's language folder.
+		$filename = $domain . '-' . $locale . '.mo';
+		$file     = $plugin_rel_path . $filename;
+
+		if ( ! file_exists( $file ) && $dir !== false ) {
```
Camwyn, line 596: A bit of explanation here:
`load_plugin_textdomain` will **always** return true unless you pass it a non-string for `$domain`.

What we really want to know is: will the "override" translation file (in `WP_LANG_DIR`) actually be loaded. If it doesn't sexists, it will not, so we use the plugin one.

Camwyn: Added @lucatume and @bordoni as reviewers because of their experience with translations.

src/Tribe/Main.php
```diff
@@ -585,9 +586,14 @@ public function load_text_domain( $domain, $dir = false ) {
 		 */
 		$plugin_rel_path = apply_filters( 'tribe_load_text_domain', $plugin_rel_path, $domain, $locale, $dir );
 
+		// Try to load translations from wp-content/languages/plugins
 		$loaded = load_plugin_textdomain( $domain, false, $plugin_rel_path );
```
redscar, line 591: With the changes below, I don't think we need `$loaded` anymore.

bordoni (approved):
I looks like a good fix, I would love to get plenty of QA here.

pattihis:
This fix also resolves https://stellarwp.atlassian.net/browse/ET-2285

pattihis (dismissed):

src/Tribe/Main.php
```diff
@@ -566,40 +566,50 @@ public function should_load_common_admin_css() {
...
+		$file     = $plugin_rel_path . $filename;
+
+		// If translation files are not found in the custom folder, then load from the plugin's language folder.
+		if (
+			! file_exists( $file )
+			&& $dir !== false
+			&& $plugin_rel_path !== $dir
+		) {
 			return load_plugin_textdomain( $domain, false, $dir );
 		}
```
dpanta94, line 614: Shouldn't this be above line 602 ?

And current line 602 be executed only when those conditionals evaluate to false, or they failed to load the textdomain ?

tests/integration/Tribe/Language_Test.php
```diff
@@ -0,0 +1,141 @@
...
+use Tribe__Main;
+use Tribe__Events__Main;
+
+class Language_Test extends WPTestCase {
+
+	/**
+	 * @dataProvider load_text_domain_data_provider
+	 * @test
+	 */
+	public function it_should_load_text_domain_based_on_conditions( Closure $setup, string $expected_result, ?string $expected_message = null ): void {
```
jesseeproductions, line 18: The $expected_message is not used. You could remove it. 

Or use it and remove the conditional check for success and failure.

tests/integration/Tribe/Language_Test.php
```diff
@@ -0,0 +1,141 @@
...
+		// Execute the setup logic and retrieve the result.
+		$result = $setup();
```
dpanta94, line 20: I would also like to evaluate how many times we call `load_plugin_textdomain` in every case. Thats where my comments on prod code are aimed for.

if the first call of load_plugin_textdomain is good there shouldn't be a second.

Can you try to test that scenario ?

dpanta94 (changes requested):
Requested an extra scenario and left a question! Looks really good though!!

src/Tribe/Main.php
```diff
@@ -566,40 +566,50 @@ public function should_load_common_admin_css() {
...
-		 * @param string      $locale Which Language we will load.
-		 * @param string|bool $dir    If there was a custom directory passed on the method call.
+		 * @param string      $domain          The plugin domain we are trying to load.
+		 * @param string      $locale          The language we will load.
+		 * @param string|bool $dir             If there was a custom directory passed on the method call.
 		 */
 		$plugin_rel_path = apply_filters( 'tribe_load_text_domain', $plugin_rel_path, $domain, $locale, $dir );
 
+		// Try to load translations from the plugin's language folder or a custom folder.
 		$loaded = load_plugin_textdomain( $domain, false, $plugin_rel_path );
```
codingmusician, line 602: I would remove this and at the end of the method just return `load_plugin_textdomain( $domain, false, $plugin_rel_path )`

andrasguseo: @codingmusician @dpanta94 (your comment about moving above line 602).
We offer a filter on line 599 for the user to define a custom folder, so we need to try loading that textdomain first.
If the user's custom folder fails, then we still try the plugin's lang folder.
If it's not filtered, then `$dir` - which is the plugin's lang folder usually (coming from us) - goes into line 602. Then the check on line 608 fails due to the last condition and is not executed anymore.
Also, the minimal changes offers easy backwards-compatibility.

And I agree, the `$loaded =` can be removed.

src/Tribe/Main.php
```diff
@@ -566,40 +566,50 @@ public function should_load_common_admin_css() {
...
 
-		if ( $dir !== false && ! $loaded ) {
+		$filename = $domain . '-' . $locale . '.mo';
+		$file     = $plugin_rel_path . $filename;
+
+		// If translation files are not found in the custom folder, then load from the plugin's language folder.
+		if (
+			! file_exists( $file )
```
dpanta94, line 609:
```suggestion
		// If translation files are not found in the custom folder, then load from the plugin's language folder.
		if (
			! $loaded
```

src/Tribe/Main.php
```diff
@@ -566,40 +566,50 @@ public function should_load_common_admin_css() {
...
+			&& $dir !== false
+			&& $plugin_rel_path !== $dir
+		) {
 			return load_plugin_textdomain( $domain, false, $dir );
 		}
```
codingmusician, line 615: My suggestion:

```suggestion
		$filename = $domain . '-' . $locale . '.mo';
		$file     = $plugin_rel_path . $filename;

		// If translation files are not found in the custom folder, then load from the plugin's language folder.
		if (
            $dir !== false
			&& $plugin_rel_path !== $dir
			&& ! file_exists( $file )
		) {
			return load_plugin_textdomain( $domain, false, $dir );
		}

        return load_plugin_textdomain( $domain, false, $plugin_rel_path );

```
Remove the `return $loaded;` below this.

src/Tribe/Main.php
andrasguseo: @dpanta94 `load_plugin_textdomain` doesn't check if the file exists. It'll return `false` only when the passed `$domain` parameter is not a string. Otherwise it'll return `true`; So checking for `$loaded` doesn't make sense unftly.

src/Tribe/Main.php
andrasguseo: @codingmusician Restructured the checks in the if.

src/Tribe/Main.php
```diff
@@ -566,44 +566,52 @@ public function should_load_common_admin_css() {
...
+		// Load textdomain from a custom folder or the plugin's language folder.
+		if ( $mo_file_exists ) {
+			return load_plugin_textdomain( $domain, false, $plugin_rel_path );
 		}
 
-		return $loaded;
+		// If translation files are not found in the custom folder, then load textdomain from the plugin's language folder.
+		load_plugin_textdomain( $domain, false, $dir );
+
+		return true;
```
codingmusician, line 614: Since `$domain` is being passed in let's utilize `load_plugin_textdomain` to determine whether to return true/false.
```suggestion
		// If translation files are not found in the custom folder, then load textdomain from the plugin's language folder.
		return load_plugin_textdomain( $domain, false, $dir );
```

src/Tribe/Main.php
```diff
@@ -566,44 +566,52 @@ public function should_load_common_admin_css() {
...
+		 * @param string      $domain          The plugin domain we are trying to load.
+		 * @param string      $locale          The language we will load.
+		 * @param string|bool $dir             If there was a custom directory passed on the method call.
 		 */
 		$plugin_rel_path = apply_filters( 'tribe_load_text_domain', $plugin_rel_path, $domain, $locale, $dir );
 
-		$loaded = load_plugin_textdomain( $domain, false, $plugin_rel_path );
+		$filename       = $domain . '-' . $locale . '.mo';
+		$file           = trailingslashit( WP_PLUGIN_DIR ) . trailingslashit( $plugin_rel_path ) . $filename;
+		$mo_file_exists = file_exists( $file );
```
dpanta94, line 604:
```suggestion
```

src/Tribe/Main.php
```diff
@@ -566,44 +566,52 @@ public function should_load_common_admin_css() {
...
 
-		if ( $dir !== false && ! $loaded ) {
-			return load_plugin_textdomain( $domain, false, $dir );
+		// Load textdomain from a custom folder or the plugin's language folder.
+		if ( $mo_file_exists ) {
```
dpanta94, line 607:
```suggestion
		if ( file_exists( $file ) ) {
```

No point in this local var

src/Tribe/Main.php
redscar: I do agree with you. However, when I was working on the test `load_plugin_textdomain` was returning false in certain scenarios.

Since this is the fallback, we could potentially either add an additional check to make sure that the fallback should work properly, or always return true.

I'm open for either option. The additional check could be checking for the `.mo` file in the `WP_LANG_DIR/plugins/*` directory. If it exists, we return true, if it doesn't we return false.

dpanta94 (changes requested):
Some optimizations

tests/integration/Tribe/Language_Test.php
```diff
@@ -0,0 +1,211 @@
...
+	protected $mo_file_path;
+
+	protected static $load_plugin_textdomain_call_count = 0;
+
+	/**
+	 * @before
+	 * @return void
+	 */
+	protected function setup(): void {
+		global $wp_textdomain_registry, $l10n;
```
dpanta94, line 37: take backups and restore them, what we are doing here should not leak :D

tests/integration/Tribe/Language_Test.php
```diff
@@ -0,0 +1,211 @@
...
+	/**
+	 * Data provider for `it_should_load_text_domain_based_on_conditions`.
+	 *
+	 * @return \Generator
+	 */
+	public function load_text_domain_data_provider(): \Generator {
+		yield 'Valid .mo file in custom directory' => [
+			function () {
+				$text_domain        = 'the-events-calendar';
+				$locale             = get_locale();
```
dpanta94, line 128: hardcode a locale other than english would be my suggestion.

tests/integration/Tribe/Language_Test.php
```diff
@@ -0,0 +1,211 @@
...
+				$text_domain = 'the-events-calendar';
+				$locale      = get_locale();
+				$temp_dir    = sys_get_temp_dir() . '/wp-languages/plugins/';
+				$plugin_dir  = WP_PLUGIN_DIR . '/the-events-calendar/lang/';
+				$pot_file    = $plugin_dir . $text_domain . '.pot';
+				$mo_file     = $temp_dir . $text_domain . '-' . $locale . '.mo';
+
+				// Override WP_LANG_DIR for this test.
+				if ( defined( 'WP_LANG_DIR' ) ) {
+					$this->set_fn_return( 'WP_LANG_DIR', $temp_dir );
```
dpanta94, line 163: set_const_value wihtout the need to check if defined prior

src/Tribe/Main.php
```diff
@@ -566,44 +566,49 @@ public function should_load_common_admin_css() {
...
+	 * First, it tries to load language files from a custom folder when set. If it fails it falls back to load from $dir.
 	 *
 	 * @since  4.0.1 Introduced.
 	 * @since  4.2   Included $domain and $dir params.
+	 * @since  TBD   Fix text domain loading logic after changes in WordPress 6.7.
 	 *
 	 * @param string       $domain The text domain that will be loaded.
 	 * @param string|false $dir    What directory should be used to try to load if the default doesn't work.
 	 *
 	 * @return bool  If it was able to load the text domain.
```
Camwyn, line 579:
```suggestion
	 * @return bool  If it was able to register the text domain.
```
To whit: `load_plugin_textdomain` now just registers the path and does not handle _laoding_ it. We retrun true if the domain is already loaded, or if we successfully _register_ the path

Camwyn (approved):

PR #2395, opened 2025-02-13
Fix the tests for Tribe__Updater as we can no longer instantiate the base class directly.

Camwyn:
Add a getter and setter for the TEC\Common\Updater class.

[TCMN-181]

Created against `master` so folks can easily merge it into their current projects.

[TCMN-181]: https://stellarwp.atlassian.net/browse/TCMN-181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Updater.php
+++ b/src/Common/Updater.php
@@ -58,6 +58,24 @@ class Updater extends Tribe__Updater {
 	}
 
 	/**
+	 * Allows setting the version we are using for running tasks.
+	 * 
+	 * @param string $version
+	 */
+	public function set_version( string $version ) {
+		$this->current_version = $version;
+	}
+
+	/**
+	 * Gets the current version we are using for running tasks.
+	 * 
+	 * @return string The current version we are using.
+	 */
+	public function get_version(): string {
+		return $this->current_version;
+	}
+
+	/**
 	 * Hook into admin init and run the update process.
 	 * 
 	 * @since 5.6.1.1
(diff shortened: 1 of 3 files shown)


src/Common/Updater.php
```diff
@@ -57,6 +57,24 @@ public function __construct( $current_version = null ) {
...
+	 */
+	public function set_version( string $version ) {
+		$this->current_version = $version;
+	}
+
+	/**
+	 * Gets the current version we are using for running tasks.
+	 * 
+	 * @return string The current version we are using.
+	 */
```
dpanta94, line 73:
```suggestion
	/**
	 * Gets the current version we are using for running tasks.
	 * 
	 * @since TBD
	 *
	 * @return string The current version we are using.
	 */
```

src/Common/Updater.php
```diff
@@ -57,6 +57,24 @@ public function __construct( $current_version = null ) {
 		$this->current_version = ! empty( $current_version ) ? $current_version : Tribe__Main::VERSION;
 	}
 
+	/**
+	 * Allows setting the version we are using for running tasks.
+	 * 
+	 * @param string $version
```
dpanta94, line 63:
```suggestion
	 * @param string $version The version to set.
```

dpanta94 (commented):
And we probably need a changelog because of the new methods.

dpanta94 (approved):

bordoni (approved):

PR #2396, opened 2025-02-15
Add notifications sidebar for ET

pattihis:
### 🎫 Ticket

[ET-2294]

### 🗒️ Description

Loads IAN notifications in ET. Goes along with 
https://github.com/the-events-calendar/event-tickets/pull/3563/

### ✔️ Checklist

[ET-2294]: https://stellarwp.atlassian.net/browse/ET-2294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Notifications/Controller.php
+++ b/src/Common/Notifications/Controller.php
@@ -111,7 +111,7 @@ class Controller extends Controller_Contract {
 	 */
 	public function is_ian_page() {
 		$screen  = get_current_screen();
-		$allowed = [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings', 'tickets_page_tec-tickets-settings' ];
+		$allowed = [];
 
 		/**
 		 * Filter the allowed pages for the Notifications icon.
--- a/src/Common/Notifications/Controller.php
+++ b/src/Common/Notifications/Controller.php
@@ -111,7 +111,7 @@ class Controller extends Controller_Contract {
 	 */
 	public function is_ian_page() {
 		$screen  = get_current_screen();
-		$allowed = [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings' ];
+		$allowed = [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings', 'tickets_page_tec-tickets-settings' ];
 
 		/**
 		 * Filter the allowed pages for the Notifications icon.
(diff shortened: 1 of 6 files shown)


src/Common/Notifications/Controller.php
```diff
@@ -111,7 +111,7 @@ public function register_ian() {
 	 */
 	public function is_ian_page() {
 		$screen  = get_current_screen();
-		$allowed = [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings' ];
+		$allowed = [ 'tribe_events', 'edit-tribe_events', 'tribe_events_page_tec-events-settings', 'tickets_page_tec-tickets-settings' ];
```
Camwyn, line 114: It would really be better if ET hooked into the filter below and added itself. Ditto for TEC pages (but that's another ticket)

src/resources/js/ian-client.js
```diff
@@ -496,6 +496,8 @@
 			}
 		};
 
+		const whichPlugin = () => document.body.classList.contains('tickets_page_tec-tickets-settings') ? 'et' : 'tec';
```
Camwyn, line 499: Basing this on a class makes me nervous. First, we are making TCMN _more_ entangled in the parent plugin (ET in this case) And second, classes change so should not be treated a "source of truth" for selectors (in many cases).

Can we pass grab the page param from the URL, or something similar?
Can we pass these things from ET (localized data maybe?) so that if ET changes, they change automatically?

src/Common/Notifications/Controller.php
Camwyn: We should always try to minimize the amount of plugin names, etc we use in TCMN - we have a hook, use it and let the plugin itself handle the strings.

Camwyn (changes requested):

src/resources/js/ian-client.js
pattihis: > making TCMN more entangled in the parent plugin

This does makes sense but on the other hand, TCMN is only used by TEC and ET, and that's why I went for this conditional.

> And second, classes change

This is not just a css class, it's the menu page slug so it will not easily change. For this to happen we would need to change the way we are registering our admin pages. In which case the `get_current_screen()` will also be affected. I think this is safer than a window variable but I don't mind changing it.

Camwyn: OK, I don't want this hung up - it has to be done today for the MR sprint.

We didn't get past the timing issues quickly enough, let's go with the class for now and circle back later.

Camwyn (approved):

PR #2446, opened 2025-03-26
Fix/improve memory usage context

bordoni:
Attempting to reduce the amount of memory usage, with the actual solution for the problem reported here:

https://github.com/the-events-calendar/the-events-calendar/pull/4986

### DIFF
--- a/src/Tribe/Context.php
+++ b/src/Tribe/Context.php
@@ -238,15 +238,6 @@ class Tribe__Context {
 	protected $request_cache = [];
 
 	/**
-	 * Whether this context should use the default locations or not.
-	 * This flag property is set to `false` when a context is obtained using
-	 * the `set_locations` method; it will otherwise be set to `true`.
-	 *
-	 * @var bool
-	 */
-	protected $use_default_locations = true;
-
-	/**
 	 * An instance of the post state handler.
 	 *
 	 * @since 5.0.13
@@ -258,7 +249,7 @@ class Tribe__Context {
 	/**
 	 * Tribe__Context constructor.
 	 *
-	 * since 5.0.13
+	 * @since 5.0.13
 	 *
 	 * @param Post_Request_Type|null $post_state An instance of the post state handler.
 	 */
@@ -444,22 +435,7 @@ class Tribe__Context {
 	public function get_locations() {
 		$this->populate_locations();
 
-		$locations = $this->use_default_locations
-			? array_merge( self::$locations, $this->override_locations )
-			: $this->override_locations;
-
-		if ( $this->use_default_locations ) {
-			/**
-			 * Filters the locations registered in the Context.
-			 *
-			 * @since 4.10.2
-			 *
-			 * @param $locations array           An array of read and write location in the shape of the `Tribe__Context::$locations` one,
-			 *                                   `[ <location> => [ 'read' => <read_locations>, 'write' => <write_locations> ] ]`.
-			 * @param $context   Tribe__Context  Current instance of the context.
-			 */
-			$locations = apply_filters( 'tribe_context_locations', $locations, $this );
-		}
+		$locations = array_merge( self::$locations, $this->override_locations );
 
 		return $locations;
 	}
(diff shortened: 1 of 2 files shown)


src/Tribe/Context.php
```diff
@@ -389,6 +392,8 @@ public function get( $key, $default = null, $force = false ) {
 		// Only cache if the value was found.
 		if ( $found ) {
 			$this->request_cache[ $key ] = $value;
+			codecept_debug( 'context cache' );
```
lucatume, line 395: Left-over from debug.

src/Tribe/Context.php
```diff
@@ -1320,6 +1309,22 @@ public function dangerously_repopulate_locations() {
...
+	 * Unless you very specifically know what you are doing **DO NOT USE THIS METHOD**!
+	 *
+	 * Please keep in mind this will set force the context to repopulate all locations for the whole request, expensive
+	 * and very dangerous overall since it could affect all this things we hold dear in the request.
+	 *
+	 * With great power comes great responsibility: think a lot before using this.
+	 *
+	 * @since TBD
+	 */
+	public function dangerously_reset_static_properties() {
```
lucatume, line 1323: Add the `: void` return type?

src/Tribe/Context.php
```diff
@@ -1160,7 +1144,7 @@ public function to_array(  ) {
 	 *
 	 * @return array
 	 */
-	public function get_state( array $fields = null, $whitelist = true ) {
+	public function get_state( ?array $fields = null, $whitelist = true ) {
```
lucatume, line 1147: Here and elsewhere.
Since you're here: add the `array` return type and document the `@return array` value.

src/Tribe/Context.php
```diff
@@ -1285,6 +1269,11 @@ public function get_orm_args( array $fields = null, $whitelist = true ) {
 	 * @since 4.9.8
 	 */
 	protected function populate_locations() {
+		// in this instance we don't want to prepoulate the locations.
```
lucatume, line 1272:
```suggestion
		// In this instance we don't want to prepoulate the locations.
```

tests/wpunit/Tribe/ContextTest.php
```diff
@@ -1636,6 +1636,7 @@ public function should_not_change_values_after_repopulating_with_cache() {
 		$value_before_reset = $context->get( $context_key( '__closure__' ) );
 
 		add_filter( 'tribe_context_locations', static function( $locations ) use ( $context_key ) {
+			codecept_debug( 'static filter two' );
```
lucatume, line 1639: Remove this: left-over from debug.

tests/wpunit/Tribe/ContextTest.php
```diff
@@ -1623,8 +1620,11 @@ public function should_not_change_values_after_repopulating_with_cache() {
 		tribe_update_option( '__after_repopulate__', '__value_after_repopulate__' );
 
 		$context = new Context;
+		// This is to ensure the static properties are reset so the test before does not affect the test after.
+		$context->dangerously_reset_static_properties();
```
lucatume, line 1624: It looks like this, being a call to a static method, should be moved to a `@before/@after` method to avoid leaking state between test methods in this class and other test cases.

tests/wpunit/Tribe/ContextTest.php
```diff
@@ -1729,6 +1736,8 @@ public function should_overwrite_locations_will_be_repopulated_when_using_defaul
 
 		$value_overwrite_before_reset = $context->get( $context_key( '__closure_overwrite__' ) );
 		$value_before_reset = $context->get( $context_key( '__closure__' ) );
+		codecept_debug( 'value before reset' );
```
lucatume, line 1739: More left-over from debug.

tests/wpunit/Tribe/ContextTest.php
```diff
@@ -1707,6 +1712,8 @@ public function should_overwrite_locations_will_be_repopulated_when_using_defaul
 		tribe_update_option( '__before_repopulate__', '__value_before_repopulate__' );
 		tribe_update_option( '__after_repopulate__', '__value_after_repopulate__' );
 
+		tribe_context()->dangerously_reset_static_properties();
```
lucatume, line 1715: This call tells me the an option update will not trigger a complete or partial invalidation of the cached values, correct? 
While an immutable-in-the-context-of-a-request state is more efficient, it's also implicitly asking of code that updates the state (option, request vars, cached values) and then uses a value from the Context that should be affected by it to know about the immutability or know about the API required to invalidate the cache (the `dangerous` function).
This burden of knowledge on the client code makes this a leaky abstraction; what about using the ephemeral cache to store the Context state, invalidate on the usual post or option updates (likely happening in a WRITE request) using the cache listeners and be fairly sure the cache would likely never be invalidated once cached in the context of a READ request?

src/Tribe/Context.php
lucatume: I would update the name to  `dangerously_reset_state`; the `static properties` part suggests **how** it's done; it should suggest the **what**.

lucatume (changes requested):
I've left comments where I found things, but I have two main concerns:
1. The context was born to have a fetch-once-on-demand strategy: once a location had been accessed and resolved once, its value would be cached. Why is that not working?
2. I think the explicit cache invalidation is not a good API; the Context should just know when to cache and invalidate and manage that internally. Otherwise it becomes a use-with-care API that will make it hard to debug and use.

lucatume (approved):
Nice!

Approved (should EVA tests pass?)

bordoni:
I will try to get it to pass later, but they come and go.

PR #2496, opened 2025-04-12
Fine-grained disable the block editor

JPry:
- **Add initial block editor controller**
- **Add methods for block loading for post type**
- **Use ::class to reference the registered classes**
- **Register the Block_Logic provider**
- **Mark the editor class as deferred**
- **Remove no-op method**
- **Tweak Regexes**
- **Correct docblock parameters**
- **Simplify dynamic block stripping**
- **Fix typo**
- **Use interpolation and clean up some methods**
- **Apply formatting to the file**

### 🎫 Ticket

[TICKET_ID]

### ✔️ Checklist

### DIFF
--- a/src/Common/Editor/Block_Logic.php
+++ b/src/Common/Editor/Block_Logic.php
@@ -11,6 +11,7 @@ declare( strict_types=1 );
 
 namespace TEC\Common\Editor;
 
+use Exception;
 use TEC\Common\Contracts\Provider\Controller;
 use WP_Screen;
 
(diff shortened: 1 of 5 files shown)


src/Common/Editor/Block_Logic.php
```diff
@@ -0,0 +1,203 @@
...
+
+	/**
+	 * Return the value of should_load_blocks passed through a filter.
+	 *
+	 * @since TBD
+	 *
+	 * @param bool   $should_load The value of should_load_blocks.
+	 * @param string $reason      The reason why we are returning this value.
+	 *
+	 * @return bool
```
lucatume, line 120:
```suggestion
	 * @return bool Whether blocks should be loaded.
```

src/Common/Editor/Block_Logic.php
```diff
@@ -0,0 +1,203 @@
...
+use WP_Screen;
+
+/**
+ * Class Block_Logic
+ *
+ * @since TBD
+ */
+class Block_Logic extends Controller {
+
+	/** @var ?WP_Screen */
```
lucatume, line 25: Add `@since annotation` here.

src/Common/Editor/Block_Logic.php
```diff
@@ -0,0 +1,203 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $post_type The post type to check.
+	 *
+	 * @return bool
+	 */
+	public function should_load_blocks_for_post_type( string $post_type = '' ): bool {
+		// Set up a static variable to ensure we don't check the same post type multiple times.
+		static $checked_post_types = [];
```
lucatume, line 151: Use `tribe_cache()[ $cache_key ]` (`ArrayAccess` API)  for memoization. Static variables will not flush on `wp_cache_flush` calls making programmatic cache invalidation difficult (includes test, but not only those).

src/Common/Editor/Block_Logic.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @throws Exception If the current post type cannot be determined.
+	 */
+	private function determine_current_post_type(): string {
+		// If we have the screen object, use its post type.
+		if ( null !== $this->screen ) {
+			return $this->screen->post_type;
+		}
+
+		// Try to use the global post object.
+		if ( isset( $GLOBALS['post']->post_type ) ) {
```
lucatume, line 197: Will this work reliably in REST API and Ajax context?

src/Common/Editor/Block_Logic.php
```diff
@@ -0,0 +1,203 @@
...
+	 * Return the value of should_load_blocks passed through a filter.
+	 *
+	 * @since TBD
+	 *
+	 * @param bool   $should_load The value of should_load_blocks.
+	 * @param string $reason      The reason why we are returning this value.
+	 *
+	 * @return bool
+	 */
+	private function return_should_load_blocks( bool $should_load, string $reason ): bool {
```
lucatume, line 122: Not a stopper, but why not `filter_should_load_blocks`?

src/Common/Editor/Block_Logic.php
lucatume: On the topic of memoization: is this worth it? Are there so many calls to this method to justify it and run the risk of serving a stale result (e.g. because the post types have been filtered later)?
It looks like the expensive call in this method is the one to `$this->determine_current_post_type()`; should that method memoized?
I'm asking this question as I do not have an order of how many calls are made to this method.

JPry: > On the topic of memoization: is this worth it? Are there so many calls to this method to justify it and run the risk of serving a stale result (e.g. because the post types have been filtered later)?

My thought process here was that every place that calls `should_load_blocks()` could potentially get to this method as well. Because this method runs `apply_filters()`, and that isn't necessarily a light call, I wanted to minimize the number of times the logic needs to run. 

With that said, I can switch to `tribe_cache()` instead.

> I'm asking this question as I do not have an order of how many calls are made to this method.

At the moment it's only called in one place. However, I was trying to be defensive about it ahead of time, knowing how the calls are likely to expand over time.

src/Common/Editor/Block_Logic.php
JPry: > Will this work reliably in REST API and Ajax context?

That's a good question: I think it would make sense to have a separate case for those situations outside of this method, since the post type isn't as relevant in those contexts.

PR #2525, opened 2025-04-23
Country List with Stripe info

pattihis:
### 🎫 Ticket

[ET-2339]

### 🗒️ Description

Under the ET Onboarding Wizard work, we centralize our Currency and Country lists in Common. We also use the [whodat Stripe API](https://whodatdev.theeventscalendar.com/commerce/v1/stripe/countries) endpoint to pull info about the payment gateway. These methods will also be used by TEC Onboarding Wizard, to centralize our data.

### ✔️ Checklist

[ET-2339]: https://stellarwp.atlassian.net/browse/ET-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Lists/Country.php
+++ b/src/Common/Lists/Country.php
@@ -337,7 +337,7 @@ class Country {
 	 */
 	public function get_gateway_countries(): array {
 		$cache     = tribe( 'cache' );
-		$cache_key = 'paymentgateway_enabled_countries';
+		$cache_key = 'payment_gateway_enabled_countries';
 
 		// Try to get from cache first.
 		$cached_data = $cache->get( $cache_key );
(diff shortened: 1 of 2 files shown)


src/Common/Lists/Country.php
```diff
@@ -327,4 +327,77 @@ public function find_country_by_value( $value ): ?string {
...
+
+		// If API error or invalid response, return all countries with default values.
+		if ( is_wp_error( $response ) || ! isset( $stripe_data['countries'] ) ) {
+			foreach ( $base_countries as $continent => $countries ) {
+				foreach ( $countries as $code => $name ) {
+					$result[ $code ] = [
+						'name'       => $name,
+						'group'      => $continent,
+						'has_paypal' => true,
+						'has_stripe' => true,
```
Camwyn, line 368: I think we should default to false, and let the API handle overwriting that

src/Common/Lists/Country.php
```diff
@@ -327,4 +327,77 @@ public function find_country_by_value( $value ): ?string {
...
+					'group'      => $continent,
+					'has_paypal' => true,
+					'has_stripe' => false,
+					'has_square' => true,
+				];
+
+				// Check if this country exists in Stripe data and is active.
+				if ( isset( $stripe_data['countries'][ $code ] ) &&
+					isset( $stripe_data['countries'][ $code ]['is_active'] ) &&
+					$stripe_data['countries'][ $code ]['is_active'] === true ) {
```
Camwyn, line 394: This can just be:
```suggestion
					! empty( $stripe_data['countries'][ $code ]['is_active'] ) {
```
since the values should be boolean?

Camwyn: Actually all of this should be covered by 
`$result[ $code ]['has_stripe'] = $stripe_data['countries'][ $code ]['is_active'] ?? false;`

src/Common/Lists/Currency.php
```diff
@@ -136,18 +148,18 @@ public static function get_currency_list(): array {
 				'symbol' => '$',
 				'entity' => '&#36;',
 			],
```
Camwyn, line 150: I'm all for adding, but why are we _removing_ currencies?

Camwyn (changes requested):

src/Common/Lists/Country.php
pattihis: Yes we do, a bit further down. Here is the case that the API is unreachable or response is not what we expected so I thought we default them all to true and let the user configure what he wants.

PR #2561, opened 2025-05-12
Create supporting methods in PHP for Square and Country by currency

dpanta94:
Supporting work for https://github.com/the-events-calendar/event-tickets/pull/3803

### DIFF
--- a/src/Common/Lists/Country.php
+++ b/src/Common/Lists/Country.php
@@ -329,6 +329,21 @@ class Country {
 	}
 
 	/**
+	 * Get a country by its currency.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $currency The currency.
+	 *
+	 * @return array|null The country or null if not found.
+	 */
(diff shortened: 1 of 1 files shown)


src/Common/Lists/Country.php
```diff
@@ -25,254 +294,50 @@ class Country {
...
+	}
+
+	/**
+	 * Get a formatted list of countries.
+	 *
+	 * @since TBD
+	 *
+	 * @return array<string,array<string,mixed>> The formatted list of countries.
+	 */
+	public function get_country_list_formatted(): array {
```
Camwyn, line 314: This is not "formatting" - this is adding data.
The functionality is great, but the function name is misleading.

Camwyn (changes requested):

src/Common/Lists/Country.php
Camwyn: "with additional data"
"with gateway info"
something that tells folks it's not just "making it pretty" but adding information/data

PR #2605, opened 2025-05-22
Fix - Issues after sync with latest elektra.

rafsuntaskin:
### 🗒️ Description
* Fixed redundant logic for determining page slug.
* Fixed PHP warning where screen is null.


andrasguseo (approved):
Looks good.

PR #2639, opened 2025-06-17
refactor(Classy) clean up store/registry logic

lucatume:
This restores the previously commented out tests and simplifies the registry/store logic.


JPry (approved):

src/resources/packages/classy/functions/classy.ts
```diff
@@ -108,19 +107,6 @@ export function toggleElementVisibility( document: Document | null = null ): voi
 export async function initApp( document: Document | null = null ): Promise< void > {
 	document = document ?? window.document;
 	const classyRoot = createRoot( getOrCreateElement( document ) );
-	const registry: WPDataRegistry = await createClassyRegistry();
-
-	// Set the module registry instance.
-	setRegistry( registry );
-
+	const registry = await getDefaultRegistry();
```
JPry, line 110: Should this be typed as `WPDataRegistry`, or is that undesirable since we're getting a Promise back?

lucatume: It's already typed from the promise return value.
Typing it explicitly will not hurt, though.

PR #2640, opened 2025-06-19
classy/component integration testing

lucatume:
The changes remove the hack I had previously used to get hold of the WordPress default registry (the one used by WordPress to handle all its stores in the Block Editor context) to align with the default WordPress API based around the use of the `withRegistry` HOC.

The refactor creates a Classy provider: a combination of a registry consumer (the `registry` instance provided by the `withRegistry` HOC) and a `RegistryProvider` (again from WordPress core) to pass-down, and capture, the current default registry.

The change was prompted by the idea to simplify the code and leveraging the fact that the `withRegistry` HOC will get, and if not already present, build, the default registry to provide it to the wrapped components. For the time being this removes the need to curate and set up a dedicated registry for the Classy application both in and out of the Block Editor context.

Functionally, the components and fields already part of the application are not affected.

This makes possible integration testing of components, in Common and in other plugins, that implicitly (through the `useRegistry` hook) or explicitly (through the `useSelect` and `useDispatch` hooks) rely on the registry instance.


src/resources/packages/classy/components/Provider.tsx
```diff
@@ -0,0 +1,92 @@
...
+		// Register the base Classy store.
+		registry.registerStore( STORE_NAME, storeConfig );
+
+		/**
+		 * Fire an action when the Classy application has initialized.
+		 * Initialized means the application got hold of the WordPress registry and registered the base store.
+		 *
+		 * @since TBD
+		 */
+		doAction( 'tec.classy.initialized' );
```
JPry, line 38: Should we rename this action, or reloacte it, since it's now within a more generalized component?

src/resources/packages/classy/components/Classy.tsx
```diff
@@ -241,12 +215,18 @@ function ClassyApplication() {
...
 }
 
-export function Classy( { registry }: { registry: WPDataRegistry } ) {
+/**
+ * The Classy application rendered in the context of the Classy provider that will provide the application
+ * with access to the WordPress default registry.
+ *
+ * @since TBD
+ *
+ * @constructor
```
JPry, line 224: Just curious for my own learning: When should we be using the `@constructor` doc tag in a JS/React context?

JPry (approved):
Just a few small things and questions; no blocking issues

src/resources/packages/classy/components/Provider.tsx
lucatume: Since this component is the one that initializes the Classy application, I would stay it can stay here.

PR #2641, opened 2025-06-20
fix(Classy) CTA font size

lucatume:
This fixes the font size of the Call-to-action text in the context of the Classy UI.
Based on the current value of the font sizes: `16px -> 14px`.


lucatume:
I've also added a missing bottom border in the modal header.

JPry (approved):

PR #2648, opened 2025-06-25
fix(Classy) hide BE welcome guide

lucatume:
When looking at a post type supported by Classy, then the Block Editor
welcome guide should be hidden.

Hiding the Welcome Guide on the frontend either using CSS or JS would
"flash" it.
This code updates the user meta controlling the persisted user
preferences without changing them, making sure the welcome guide will
not show if the user is editing a post type managed by Classy and has
never dismissed the welcome guide before.

A catch: this is the best solution I could come up with that would
cleanly do this, but a user that is editing, as their first post, a post
managed by classy will have the preference saved in **local storage** of
the browser, and will not see the Welcome Guide on other posts.
If the user has already seen the welcome guide, then they will not be
affected.

A user that wishes to see the welcome guide again can just do so from
the standard Block Editor menu.

[Demo](https://share.cleanshot.com/FCL1TBVC)

### DIFF
--- a/src/Common/Classy/Controller.php
+++ b/src/Common/Classy/Controller.php
@@ -17,6 +17,7 @@ use Tribe__Date_Utils as Date_Utils;
 use Tribe__Main as Common;
 use WP_Block_Editor_Context;
 use WP_Post;
+use wpdb;
 
 /**
  * Class Controller.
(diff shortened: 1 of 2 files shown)


JPry:
One more thought that occurred to me after I left the previous comment: Would it be better to try to modify the setting for the welcome guide from within Classy JS instead of PHP? That's where the editor is manipulating the setting anyway, so perhaps we could work with it there and bypass the issue I flagged.

PR #2672, opened 2025-07-14
Improves checkout timing with Shepherd

dpanta94:
[TCMN-185]

Improves TicketsCommerce checkout timing by offloading tasks to Shepherd.

This change enhances performance by optimizing checkout processes

[TCMN-185]: https://stellarwp.atlassian.net/browse/TCMN-185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


composer.json
```diff
@@ -33,10 +33,10 @@
     "stellarwp/db": "^1.0.3",
     "stellarwp/installer": "^1.1.0",
     "stellarwp/models": "dev-main",
-    "stellarwp/schema": "^1.1.8",
+    "stellarwp/schema": "^2.0",
+    "stellarwp/shepherd": "0.0.2",
     "stellarwp/telemetry": "^2.3.4",
-    "stellarwp/uplink": "2.2.2",
-    "woocommerce/action-scheduler": "3.9.2"
```
Camwyn, line 39: If you remove it, this will fail: https://github.com/the-events-calendar/tribe-common/blob/master/src/Tribe/Main.php#L527

Camwyn (changes requested):

composer.json
dpanta94: It won't, because we do the same in Shepherd. But to be clearer i'll remove the requirement in our own codebase too.

Action scheduler will continue to be available through Shepherd.

bordoni (approved):

PR #2677, opened 2025-07-15
Registers REST API infrastructure

dpanta94:
Initializes the REST API infrastructure for the project.

This includes:
- Registers a REST controller and versioned namespace.
- Sets up abstract endpoint classes and interfaces.
- Provides documentation generation based on registered endpoints and definitions.
- Registers an endpoint to serve the OpenAPI documentation.

YY-XXXX

### DIFF
--- a/src/Common/REST/TEC/V1/Abstracts/Endpoint.php
+++ b/src/Common/REST/TEC/V1/Abstracts/Endpoint.php
@@ -1,0 +1,158 @@
+<?php
+/**
+ * Endpoint class.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\REST\TEC\V1\Abstracts
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Common\REST\TEC\V1\Abstracts;
+
+use TEC\Common\REST\TEC\V1\Contracts\Endpoint_Interface;
+use TEC\Common\REST\TEC\V1\Contracts\Readable_Endpoint;
+use TEC\Common\REST\TEC\V1\Contracts\Creatable_Endpoint;
+use TEC\Common\REST\TEC\V1\Contracts\Updatable_Endpoint;
+use TEC\Common\REST\TEC\V1\Contracts\Deletable_Endpoint;
+use TEC\Common\REST\TEC\V1\Controller;
+use WP_REST_Server;
+use WP_REST_Request;
+
+/**
+ * Endpoint class.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\REST\TEC\V1\Abstracts
+ */
+abstract class Endpoint implements Endpoint_Interface {
+	/**
+	 * Alias for PUT, PATCH transport methods together.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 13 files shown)


src/Common/REST/TEC/V1/Abstracts/Endpoint.php
```diff
@@ -0,0 +1,158 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param WP_REST_Request $request The request object.
+	 *
+	 * @return bool
+	 */
+	public function can_read( WP_REST_Request $request ): bool {
+		return true;
+		return current_user_can( 'manage_options' );
```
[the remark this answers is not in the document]
dpanta94: This is for local dev only. By default i want only to allow admins, but for my local dev it helps me to always allow!

Will fix before merge!

src/Common/REST/TEC/V1/Abstracts/Endpoint.php
```diff
@@ -0,0 +1,158 @@
...
+	 * Returns whether the user can create the object.
+	 *
+	 * @since TBD
+	 *
+	 * @param WP_REST_Request $request The request object.
+	 *
+	 * @return bool
+	 */
+	public function can_create( WP_REST_Request $request ): bool {
+		return $this->can_read( $request );
```
Camwyn, line 130: I'm assuming these are local dev too? 😉

src/Common/REST/TEC/V1/Controller.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+class Controller extends Controller_Contract {
+	/**
+	 * The version of the REST API.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const VERSION = '1';
```
Camwyn, line 32: If this is _always_ going to be an int, we should type it that way so no one puts in `1.5` or `1.0.1`

🤔 Probably best in these cases to put a long description for the var that tells devs what is allowable?

Camwyn: PS: I'm really thinking of setting up expectations on the client side as well...

src/Common/REST/TEC/V1/Endpoints/OpenApiDocs.php
```diff
@@ -0,0 +1,88 @@
...
+ * @since TBD
+ *
+ * @package TEC\Common\REST\TEC\V1\Endpoints
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Common\REST\TEC\V1\Endpoints;
+
+// phpcs:disable StellarWP.Classes.ValidClassName.NotSnakeCase
```
Camwyn, line 14: Why? Does it _need_ to be in CamelCase?

dpanta94: Yes, if you visit https://www.openapis.org/ the OpenAPI is a name and its written like that. The `Docs` part of the name could be snake case but i'd have to ignore in any case.

Camwyn: OK, I figured it was a branding thing.

PR #2679, opened 2025-07-17
KV Cache - first iteration

lucatume:
This objective of this first iteration of the key-value cache API is to provide a transparent API to store computationally intensive and possible large string values across requests.

When object-caching it's available the API will use it, else it will use a custom table with a pretty simple structure.

The unique API is the `tec_kv_cache()` function, see more in the documentation.

Following a WP VIP lead, I've set the minimum accepted expiration time to 300 seconds, 5 minutes, else this is not the cache to use.

Furthermore the cache key column is set to a length of 191 chars following the example of the `wp_postmeta.meta_key` column: 255 chars long, but only 191 are indexed. I've picked the lower value to play on the safe side.

The feature first application would be to adress an ET performance issue, but it comes with tests.

The current implementation is a first step to test the waters and has a dumb-cache approach that will work, but might end up storing redundant values. E.g.:
```
- my_plugin_all_items -> JSON array of 3 items
- my_plugin_item_23 -> JSON object representing item 23
- my_plugin_item_89 -> JSON object representing item 89
- my_plugin_item_66 -> JSON object representing item 66
```

The items `23, 66, 89` are repeated inside the `all_items` key, it's redundant.
_Some_ system avoiding this would be nice.

### DIFF
--- a/src/Common/Key_Value_Cache/Key_Value_Cache_Table.php
+++ b/src/Common/Key_Value_Cache/Key_Value_Cache_Table.php
@@ -1,0 +1,215 @@
+<?php
+/**
+ * The API of the key-value cache.
+ *
+ * @since TBD
+ *
+ * @package Common\Key_Value_Cache;
+ */
+
+namespace TEC\Common\Key_Value_Cache;
+
+use Exception;
+use TEC\Common\Key_Value_Cache\Table\Schema;
+use TEC\Common\StellarWP\DB\DB;
+
+/**
+ * Class Key_Value_Cache_Table.
+ *
+ * @since TBD
+ *
+ * @package Common\Key_Value_Cache;
+ */
+class Key_Value_Cache_Table implements Key_Value_Cache_Interface {
+	use Key_Value_Cache_Methods_Trait;
+
+	/**
+	 * Whether a key exists in the cache.
+	 *
+	 * This method tells nothing about the nature of the value, it only checks whether the key exists in the cache.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $key The key to check.
+	 *
+	 * @return bool Whether the key exists in the cache.
+	 */
+	public function has( string $key ): bool {
+		$current_time = time();
+
+		try {
+			$exists = DB::get_var(
+				DB::prepare(
+					'SELECT COUNT(*) FROM %i WHERE `cache_key` = %s AND (`expiration` = 0 OR `expiration` > %d)',
+					Schema::table_name(),
+					$key,
+					$current_time
+				)
+			);
+		} catch ( Exception $e ) {
+			return false;
+		}
+
+		return $exists > 0;
+	}
+
+	/**
+	 * Returns a cached value if it exists.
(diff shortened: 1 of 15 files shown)


src/Common/Key_Value_Cache/Key_Value_Cache_Table.php
```diff
@@ -0,0 +1,215 @@
...
+	 *
+	 * @return bool Whether the key exists in the cache.
+	 */
+	public function has( string $key ): bool {
+		$current_time = time();
+
+		try {
+			$exists = DB::get_var(
+				DB::prepare(
+					'SELECT COUNT(*) FROM %i WHERE `cache_key` = %s AND (`expiration` = 0 OR `expiration` > %d)',
```
dpanta94, line 43: It seems like we dont support excpiration 0 anymore (5 mins min), so maybe we should amend the queries ?

src/Common/Key_Value_Cache/Key_Value_Cache_Table.php
```diff
@@ -0,0 +1,215 @@
...
+	public function set( string $key, string $value, int $expiration = 300 ): bool {
+		// Bail if the key is longer than 191 characters.
+		if ( strlen( $key ) > 191 ) {
+			return false;
+		}
+
+		if ( $expiration < 300 ) {
+			// To stick with WP VIP min requirements, we're not going to store values with an expiration under 5 minutes.
+			return false;
+		}
```
dpanta94, line 115: The silent failure is the desired approach ?

src/Common/Key_Value_Cache/Key_Value_Cache_Table.php
```diff
@@ -0,0 +1,215 @@
...
+
+		try {
+			DB::query( "TRUNCATE TABLE {$table_name}" );
+		} catch ( Exception $e ) {
+			do_action(
+				'tribe_log',
+				'error',
+				'Key Value Cache flush error: ' . $e->getMessage()
+			);
+		}
```
dpanta94, line 183: There is a truncate method available through stellarwp/schema. There is also an empty_table one which can help with avoiding Truncate during tests.

src/Common/Key_Value_Cache/Table/Schema.php
```diff
@@ -0,0 +1,87 @@
...
+use TEC\Common\StellarWP\Schema\Tables\Contracts\Table;
+
+/**
+ * Class Schema.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\Key_Value_Cache\Table;
+ */
+class Schema extends Table {
```
dpanta94, line 21: You can skip this, but we also provide an internal table contract at `TEC\Common\Integrations\Custom_Table_Abstract`.

For which you can provide a definition of columns and all the select/insert/update/delete queries would be provided to you by the `Custom_Table_Query_Methods` trait.

dpanta94 (changes requested):

PR #2681, opened 2025-07-19
JSON Packer API

lucatume:
Ticket (for context): [ETP-1021](https://stellarwp.atlassian.net/browse/ETP-1021)

This PR introduces the JSON Packer API.

The idea is to provide an API that allows getting a JSON string representation of values of any type.
The API boils down to the `tec_json_pack` and `tec_json_unpack` functions with a strict-by-default policy for object encoding (i.e. no class beyond some default ones allowed unless specified).

The PR builds on the one introducing the Key-Value cache API (#2679) and supersedes it to complete its usefulness with one API that reads and writes strings from/to the cache, the key-value cache API, and one that transforms any value to a string and viceversa.

Finally, building on the ability to transform anything into a string, the PR removes the `set_json`/`get_json` and `set_serialized`/`get_serialized` methods from the key-value cache API. Those were first-iteration leaky abstractions that would not really prevent thoughtless storing of objects of any class into the database and the side-effect of their serialization.

[ETP-1021]: https://stellarwp.atlassian.net/browse/ETP-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Json_Packer/Json_Packer.php
```diff
@@ -0,0 +1,523 @@
...
+		// Use reflection to access all properties.
+		$reflection = new ReflectionClass( $object_to_pack );
+		// This variable starts as an accumulator.
+		/** @var ReflectionProperty[][] $properties */
+		$properties = [];
+
+		// Get all properties from the class and its parents.
+		do {
+			$properties[] = $reflection->getProperties();
+		} while ( $reflection = $reflection->getParentClass() );
```
dpanta94, line 254: I am not very fond of this. This is going to store private and protected properties as well. If we are storing an object why not first check if it implements `JsonSerializable` or `Serializable` ?

if not just normally wp_json_encode the object getting only the public properties.

dpanta94 (changes requested):
I also left comments on #2679 which should be applicable

src/Common/Json_Packer/Json_Packer.php
lucatume: If the methods from `Serializable` or `JsonSerializable` remove properties that will need re-initialization to work, then the object cannot be rebuilt into a working instance during the unpacking operation.

There are number of objects we use as data-transfer objects that have private and protected properties exposed through methods. Those objects would need to be rebuilt from scratch, or each one of those objects would require the ad-hoc implementation of an interface (`Serializable`, `JsonSerializable`) or other.
I can see how the easy way out would be to make all the object properties `public` (not ideal). 

One of the objectives is to make this API immediately usable with object that we already use, without having to modify each one of them.

`wp_json_encode` does not maintain the class information; the data will be recovered, but not the objects that use that data. Again: the objective here is to have an API that does that.

Furthermore, the fact that the `pack`/`unpack` operations replace classes that cannot be packed/unpacked with `stdClass` adds a layer of safety over the `unserialize` methods.

I think the `allowed_classes` required parameter is a good layer of protection.

dpanta94 (approved):
Looks good!

dpanta94 (dismissed):

PR #2683, opened 2025-07-22
Add utility method to validate URL.

rafsuntaskin:
### 🎫 Ticket

Issue - 14

### 🗒️ Description
* Check if a string is a valid URL.
* Added tests for all possible allowed URLs.

### DIFF
--- a/src/resources/packages/classy/functions/index.ts
+++ b/src/resources/packages/classy/functions/index.ts
@@ -1,2 +1,3 @@
 export * from './dateUtils';
+export * from './utils';
 export { sortOptionsForDisplay } from './sortOptionsForDisplay';
--- a/src/resources/packages/classy/functions/utils.ts
+++ b/src/resources/packages/classy/functions/utils.ts
@@ -1,0 +1,57 @@
+/**
+ * Validates a URL string.
+ *
+ * @since TBD
+ *
+ * @param url - The URL string to validate.
+ * @returns True if the URL is valid, false otherwise.
+ */
+export function isValidUrl( url: string ): boolean {
+	if ( ! url || url.trim() === '' ) {
+		return true; // Allow empty URLs
+	}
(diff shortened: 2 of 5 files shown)


src/resources/packages/classy/functions/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './dateUtils';
+export * from './utils';
```
lucatume, line 2: Rename the file to `url` and export it as such. Let's avoid generic `utils` where possible.

src/resources/packages/classy/functions/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param url - The URL string to validate.
+ * @returns True if the URL is valid, false otherwise.
+ */
+export function isValidUrl( url: string ): boolean {
+	if ( ! url || url.trim() === '' ) {
+		return true; // Allow empty URLs
+	}
+
+	try {
+		// Add protocol if missing for validation
```
lucatume, line 15: Here and elsewhere, add full-stops at the end of line comments.

lucatume: About comments, please trim them removing the ones explaining what the code is doing. Comment code if the "what" and "why" you're doing it is not obvious.

tests/classy_jest/components/__snapshots__/TimeZone.spec.tsx.snap
```diff
@@ -641,3 +641,297 @@ exports[`TimeZone Component renders the UTC timezone choice if the timezone stri
   </div>
 </div>
 `;
+
```
lucatume, line 644: Why did this change?

lucatume (changes requested):
Some name and comment tweaks.

PR #2684, opened 2025-07-29
test(extensions) add the Suite Env extension

lucatume:
This adds the `Suite_Env` Codeception extension to allow for a cross-compatible (i.e. across Codeception versions) control of what happens during the suites loading of modules, initialization and execution.

See attached docs.


JPry (approved):
Looks like a great addition!

PR #2694, opened 2025-08-07
Fix REST API entity update to separate set_args and save calls

Camwyn:
## Fix REST API entity update to separate set_args and save calls

This change separates the set_args() and save() method calls in the Update_Entity_Response trait to allow for proper method chaining and better debugging of entity updates.

Related to multiple venue support implementation.

Also

## Fix TypeError in Update_Entity_Response trait 

Correct an issues where array_keys() was being called on a repository object instead of using the entity ID directly.

Issue: Line 78 called array_keys()[0] where  was a repository object (e.g., Tribe__Events__Pro__Repositories__Venue) rather than an array.

Solution: Use the  parameter directly instead of trying to extract it from the repository object.

Error details:
- TypeError: array_keys() expects parameter 1 to be array, object given
- Occurred during venue/organizer updates via REST API
- Data was being saved correctly but response was returning HTTP 500

Tested scenarios:
✅ Venue updates now return HTTP 200 with proper response data
✅ Organizer updates work correctly
✅ Invalid IDs still return appropriate 403/404 errors

### DIFF
--- a/src/Common/REST/TEC/V1/Traits/Update_Entity_Response.php
+++ b/src/Common/REST/TEC/V1/Traits/Update_Entity_Response.php
@@ -75,7 +75,7 @@ trait Update_Entity_Response {
 			$this->get_formatted_entity(
 				$this->get_orm()->by_args(
 					[
-						'id'     => array_keys( $entity )[0],
+						'id'     => $id,
 						'status' => 'any',
 					]
 				)->first()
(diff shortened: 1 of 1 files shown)


dpanta94 (changes requested):

dpanta94 (approved):
Production code looks good. The test seems mostly AI.
It doesn't follow our usual test approach. And it adds quite a bit of unnecessary lines.

I would also include it in an abstract that all the other rest related tests are extending in order to run for every plugin that integrates with the rest.

PR #2711, opened 2025-08-15
Enhances REST API security and data handling

dpanta94:
- Implements password protection for REST API entities, hiding sensitive data unless authenticated.
- Introduces whitelisting for specific fields to remain visible even when password protected.
- Forgives a WP core issue where parameters that are part of the Query or the Path may be converted to string.
- Refactors entity data transformation filters for clarity and flexibility.
- Introduces `get_wp_validator` to Parameters to have WP validator separate from the regular one.


Camwyn (approved):
Looks good

PR #2724, opened 2025-08-22
feat(REST API): Update common library documentation

dpanta94:
## Summary

This PR updates the REST API to leverage WordPress core's built-in validation and sanitization system instead of implementing our own custom validation logic.

### Key Change: WordPress Core Validation Integration

The main architectural change is that we now:
- **Delegate validation to WordPress REST API**: Let WordPress core handle parameter validation and sanitization
- **Use RequestBodyCollection with bridge pattern**: The `to_query_argument_collection()` method converts our OpenAPI-structured request bodies into WordPress-compatible validation rules
- **Maintain single source of truth**: Schema definitions drive both OpenAPI documentation AND WordPress validation rules

### Validation Pipeline Refactoring
- Removed custom validation implementations in favor of WordPress core's system
- Introduced `RequestBodyCollection::to_query_argument_collection()` bridge method
- Parameters are now pre-sanitized by WordPress before reaching endpoint handlers
- Endpoint methods receive `array $params` instead of raw `WP_REST_Request`

### Documentation Updates
- Created comprehensive validation-sanitization.md documenting the new pipeline
- Updated all abstract class documentation to reflect actual implementation
- Enhanced interfaces.md with accurate method signatures
- Updated parameter-types.md with collection usage patterns
- Added migration guide from legacy array-based parameters

## Why This Change?

1. **Leverage battle-tested code**: WordPress core's validation is mature and well-tested
2. **Reduce maintenance burden**: No need to maintain custom validation logic
3. **Better compatibility**: Aligns with WordPress REST API standards
4. **Single source of truth**: Schema definitions drive both documentation and validation

## Breaking Changes

- `create_args()` and `update_args()` now return `RequestBodyCollection` instead of `QueryArgumentCollection`
- Endpoint operation methods receive pre-sanitized `array $params` instead of `WP_REST_Request`


src/Common/REST/TEC/V1/Documentation/OpenAPI_Schema.php
```diff
@@ -334,36 +325,20 @@ public function validate( array $data = [] ): self {
...
+		 * Filters the schema defined parameters.
+		 *
+		 * @since TBD
+		 *
+		 * @param array $new_data The new data.
+		 * @param OpenAPI_Schema $this The schema.
+		 *
+		 * @return array The filtered data.
+		 */
+		return (array) apply_filters( 'tec_rest_schema_filter', $new_data, $this );
```
JPry, line 342: For this filter to be more useful, I would suggest including `$data` as a parameter, and/or `$params`.

src/Common/REST/TEC/V1/Parameter_Types/Definition_Parameter.php
```diff
@@ -99,6 +100,53 @@ public function get_sanitizer(): Closure {
...
+		$collections = $this->get_collections();
+
+		$filtered_data = [];
+
+		/** @var PropertiesCollection $collection */
+		foreach ( $collections as $collection ) {
+			/** @var Parameter $property */
+			foreach ( $collection as $property ) {
+				$param_name = $property->get_name();
+				$argument   = $this->get_name() ? $this->get_name() . '.' . $param_name : $param_name;
```
JPry, line 124: Minor: I find it easier to read interpolation, especially in the context of a ternary.

```suggestion
				$argument   = $this->get_name() ? "{$this->get_name()}.{$param_name}" : $param_name;
```

src/Common/REST/TEC/V1/Parameter_Types/Definition_Parameter.php
```diff
@@ -99,6 +100,53 @@ public function get_sanitizer(): Closure {
...
+
+				if ( $property->is_required() && ! isset( $data[ $param_name ] ) ) {
+					// translators: %s is the name of the property.
+					$exception = new InvalidRestArgumentException( sprintf( __( 'Property %s is required', 'the-events-calendar' ), $argument ) );
```
JPry, line 128: We have this exact same Exception type and messaging in a few places. I think it would be better to add a static method to the `InvalidRestArgumentException` class that takes the argument name generates all of the details. Then call the static method to throw the exception.

JPry (changes requested):

docs/REST/TEC/V1/abstract-classes.md
```diff
@@ -1,147 +1,464 @@
...
+
+- `get_priority()`: Returns registration priority
+- `to_array()`: Converts to OpenAPI schema array
+
+#### Usage Example
 
-    // Implement RUD_Endpoint methods...
+```php
+class Event_Definition extends Definition {
+    public function get_name(): string {
```
JPry, line 148: One other nitpick: since these examples are in the context of WordPress, they should use tabs instead of spaces for indenting.

docs/REST/TEC/V1/interfaces.md
```diff
@@ -24,54 +24,59 @@ Base interface that all endpoints must implement. Defines:
...
 
 For endpoints that support PUT/PATCH requests:
 
-- `update(WP_REST_Request $request): WP_REST_Response`
-- `update_args(): Collection`
-- `update_schema(): OpenAPI_Schema`
-- `can_update(WP_REST_Request $request): bool`
+- `update(array $params = []): WP_REST_Response` - Handle update operation with sanitized parameters
+- `can_update(WP_REST_Request $request): bool` - Permission check for update operations
+- `update_args(): RequestBodyCollection` - Define request body structure (changed from QueryArgumentCollection)
```
lucatume, line 53: Here and elsewhere it look like this method should use `params` over `args`.
Parameters are those in the signature (what the method seems to be returning) while arguments are those in the concrete call.

src/Common/REST/TEC/V1/Contracts/OpenAPI_Schema.php
```diff
@@ -112,22 +112,13 @@ public function add_response( int $code, Closure $description_provider, ?Headers
...
-	 * Sanitizes the schema.
-	 *
-	 * @since 6.9.0
+	 * @param array $data The data to filter.
 	 *
-	 * @return array The sanitized data.
+	 * @return array The filtered schema.
 	 */
-	public function sanitize(): array;
+	public function filter( array $data = [] ): array;
```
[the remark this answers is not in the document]
lucatume: What about `filter_before_request`? The word `filter` is in line with WordPress naming conventions, but alone is a bit obscure. Adding some temporal, order-of-operation part might be helpful here.

src/Common/REST/TEC/V1/Parameter_Types/Definition_Parameter.php
```diff
@@ -117,22 +166,23 @@ public function validate( array $data = [] ): self {
...
 				if ( $property->is_required() && ! isset( $data[ $param_name ] ) ) {
-					// translators: %s is the name of the property.
-					$exception = new InvalidRestArgumentException( sprintf( __( 'Property %s is required', 'the-events-calendar' ), $argument ) );
-					$exception->set_argument( $argument );
-					$exception->set_details( __( 'The property is required but missing.', 'the-events-calendar' ) );
-					$exception->set_internal_error_code( 'tec_rest_required_property_missing' );
-					throw $exception;
+					throw InvalidRestArgumentException::create(
+						// translators: %s is the name of the property.
+						sprintf( __( 'Property %s is required', 'the-events-calendar' ), $argument ),
```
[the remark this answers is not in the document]
dpanta94: Great idea, but not something i have the time to implement currently. Maybe as the project moves on, we can add this.

PR #2741, opened 2025-08-29
Introduced model and upgraded Custom table abstracts

dpanta94:
### 🎫 Ticket

[TICKET_ID]

### ✔️ Checklist

### DIFF
--- a/src/Common/Contracts/Provider/Controller.php
+++ b/src/Common/Contracts/Provider/Controller.php
@@ -12,6 +12,7 @@ namespace TEC\Common\Contracts\Provider;
 
 use TEC\Common\Contracts\Service_Provider;
 use Tribe__Log as Log;
+use TEC\Common\Contracts\Container;
 
 /**
  * Class Controller.
--- a/src/Common/Abstracts/Custom_Table_Abstract.php
+++ b/src/Common/Abstracts/Custom_Table_Abstract.php
@@ -1,0 +1,232 @@
+<?php
+/**
+ * Abstract for Custom Tables.
+ *
+ * @since TDB
+ *
+ * @package TEC\Common\Abstracts
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Common\Abstracts;
+
+use TEC\Common\StellarWP\DB\DB;
+use TEC\Common\Traits\Custom_Table_Query_Methods;
+use DateTimeInterface;
+use TEC\Common\Integrations\Custom_Table_Abstract as Base_Table;
+
+/**
+ * Class Custom_Table_Abstract
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\Abstracts
+ */
+abstract class Custom_Table_Abstract extends Base_Table {
+	use Custom_Table_Query_Methods;
+
+	/**
+	 * The PHP type for an integer.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PHP_TYPE_INT = 'int';
+
+	/**
+	 * The PHP type for a string.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PHP_TYPE_STRING = 'string';
+
+	/**
+	 * The PHP type for a float.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PHP_TYPE_FLOAT = 'float';
+
+	/**
+	 * The PHP type for a boolean.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PHP_TYPE_BOOL = 'bool';
+
+	/**
+	 * The PHP type for a datetime.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const PHP_TYPE_DATETIME = DateTimeInterface::class;
+
+	/**
+	 * The column type for a bigint.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_BIGINT = 'bigint';
+
+	/**
+	 * The column type for a varchar.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_VARCHAR = 'varchar';
+
+	/**
+	 * The column type for a text.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_TEXT = 'text';
+
+	/**
+	 * The column type for a longtext.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_LONGTEXT = 'longtext';
+
+	/**
+	 * The column type for a boolean.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_BOOLEAN = 'boolean';
+
+	/**
+	 * The column type for a timestamp.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const COLUMN_TYPE_TIMESTAMP = 'timestamp';
+
+	public const SQL_RESERVED_DEFAULTS = [
+		'CURRENT_TIMESTAMP',
+		'CURRENT_DATE',
+		'CURRENT_TIME',
+	];
+
+	/**
+	 * Constructor.
+	 *
+	 * @since TBD
+	 */
+	public function __construct() {
+		$this->db        = DB::class;
+		$this->container = tribe();
+	}
+
+	/**
+	 * Returns the table creation SQL in the format supported
+	 * by the `dbDelta` function.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The table creation SQL, in the format supported
+	 *                by the `dbDelta` function.
+	 */
+	public function get_definition() {
+		global $wpdb;
+		$table_name      = static::table_name( true );
+		$charset_collate = $wpdb->get_charset_collate();
+		$uid_column      = static::uid_column();
+
+		$columns = static::get_columns();
+
+		$columns_definitions = [];
+		foreach ( $columns as $column => $definition ) {
+			$column_sql = "`{$column}` {$definition['type']}";
+
+			if ( ! empty( $definition['length'] ) ) {
+				$column_sql .= "({$definition['length']})";
+			}
+
+			if ( ! empty( $definition['unsigned'] ) ) {
+				$column_sql .= ' UNSIGNED';
+			}
+
+			$column_sql .= ! empty( $definition['nullable'] ) ? ' NULL' : ' NOT NULL';
+
+			if ( ! empty( $definition['auto_increment'] ) ) {
+				$column_sql .= ' AUTO_INCREMENT';
+			}
+
+			if ( ! empty( $definition['default'] ) ) {
+				$column_sql .= ' DEFAULT ' . ( in_array( $definition['default'], self::SQL_RESERVED_DEFAULTS, true ) || in_array( $definition['php_type'], [ self::PHP_TYPE_INT, self::PHP_TYPE_BOOL, self::PHP_TYPE_FLOAT ], true ) ? $definition['default'] : "'{$definition['default']}'" );
+			}
+
+			$columns_definitions[] = $column_sql;
+		}
+
+		$columns_sql = implode( ',' . PHP_EOL, $columns_definitions );
+
+		return "
+			CREATE TABLE `{$table_name}` (
+				{$columns_sql},
+				PRIMARY KEY (`{$uid_column}`)
+			) {$charset_collate};
+		";
+	}
+
+	/**
+	 * Add indexes after table creation.
+	 *
+	 * @since TBD
+	 *
+	 * @param array<string,string> $results A map of results in the format
+	 *                                      returned by the `dbDelta` function.
+	 *
+	 * @return array<string,string> A map of results in the format returned by
+	 *                              the `dbDelta` function.
+	 */
+	protected function after_update( array $results ) {
+		foreach ( static::get_columns() as $column => $definition ) {
+			if ( empty( $definition['index'] ) ) {
+				continue;
+			}
+
+			$this->add_index( $column );
+		}
+
+		return $results;
+	}
+
+	/**
+	 * Helper method to check and add an index to a table.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $index_name The name of the index.
+	 * @param string $columns    The columns to index. Defaults to the index name.
+	 *
+	 * @return void
+	 */
+	public function add_index( string $index_name, string $columns = '' ): void {
+		$this->check_and_add_index( [], $index_name, $columns ? $columns : $index_name );
+	}
+}
--- a/src/Common/Abstracts/Model_Abstract.php
+++ b/src/Common/Abstracts/Model_Abstract.php
@@ -1,0 +1,140 @@
+<?php
+/**
+ * The model abstract.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\Abstracts;
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Common\Abstracts;
+
+use TEC\Common\Contracts\Model;
+use TEC\Common\StellarWP\DB\DB;
+use RuntimeException;
+
+/**
+ * The model abstract.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\Abstracts;
+ */
+abstract class Model_Abstract implements Model {
+	/**
+	 * The model's ID.
+	 *
+	 * @since TBD
+	 *
+	 * @var int
+	 */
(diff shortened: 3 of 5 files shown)


src/Tribe/Repository.php
```diff
@@ -466,14 +466,14 @@ public function __construct() {
 	/**
 	 * {@inheritdoc}
 	 */
-	public function get_default_args() {
+	public function get_default_args(): array {
```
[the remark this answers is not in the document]
dpanta94: You are not wrong... I think restoring backwards compat is the way to go.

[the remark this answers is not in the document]
dpanta94: Those have been addressed

changelog/feat-crud-foundation-for-prs
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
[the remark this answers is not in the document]
dpanta94: Also amended

docs/Abstracts/Custom_Table_Repository.md
```diff
@@ -0,0 +1,468 @@
...
+Some methods from the base Repository Interface are not supported:
+
+- `in()` and `not_in()` - Use column-specific operators instead
+- `parent()`, `parent_in()`, `parent_not_in()` - Not applicable to custom tables
+- `sort()` - Use `order_by()` instead
+- `collect()` - Not implemented
+
+## See Also
+
+- [Model Abstract](./Model_Abstract.md)
```
JPry: This path isn't correct.

docs/Abstracts/Custom_Table_Repository.md
```diff
@@ -0,0 +1,468 @@
...
+- [Model Contract](../Contracts/Model.md)
```
JPry: This path isn't correct either.

docs/Contracts/Custom_Table_Repository_Interface.md
```diff
@@ -0,0 +1,305 @@
...
+## Common Pitfalls
+
+1. **Forgetting Parent Constructor**: Always call `parent::__construct()` in implementations
+2. **Invalid Schema Callbacks**: Ensure all schema entries are callable
+3. **Model Class Changes**: Keep model class references up to date
+4. **Schema Conflicts**: Avoid overriding auto-generated schema entries unintentionally
+
+## See Also
+
+- [Repository Interface](./Repository_Interface.md)
```
JPry: This path is wrong

src/Common/Abstracts/Custom_List_Table.php
```diff
@@ -0,0 +1,184 @@
...
+	public function column_cb( $item ): void {
+		$show = current_user_can( 'manage_options' );
+
+		if ( ! $show ) {
+			return;
+		}
+
+		?>
+		<input id="cb-select-<?php echo esc_attr( $item->getPrimaryValue() ); ?>; ?>" type="checkbox" name="item_id[]" value="<?php echo esc_attr( $item->getPrimaryValue() ); ?>" />
+		<label for="cb-select-<?php echo esc_attr( $item->getPrimaryValue() ); ?>; ?>">
```
JPry: This (and the line above) has an extra PHP closing tag that will mess up the attribute.

```suggestion
		<label for="cb-select-<?php echo esc_attr( $item->getPrimaryValue() ); ?>">
```

JPry: Also, will the result of `$item->getPrimaryValue()` ever include spaces? Would we want to ensure the value we get is ready for inclusion as an attribute beyond `esc_attr()`?

src/Common/Abstracts/Custom_List_Table.php
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Returns whether the list is completely empty.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function is_empty(): bool {
+		$repo = $this->get_repository();
+		$repo->set_use_default_args( false );
```
JPry: Seeing this pattern a few times, it feels like there should be a better way. I think it would be ideal to see if we can eliminate this pattern.

src/Common/Abstracts/Custom_List_Table.php
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Returns the arguments to query the items.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	private function get_args(): array {
+		$args = [
+			'orderby' => tec_get_request_var( 'orderby', '' ) ? $this->get_orderby() : 'created',
```
JPry: The `$this->get_orderby()` method already checks `tec_get_request_var()`, so I think we can simplify this to avoid calling the function twiceover needlessly.

```suggestion
			'orderby' => $this->get_orderby() ?: 'created',
```

src/Common/Abstracts/Custom_List_Table.php
```diff
@@ -0,0 +1,184 @@
...
+			'order'   => tec_get_request_var( 'order', '' ) ? $this->get_order() : 'DESC',
```
JPry: Similar to the live above, it looks like we can mostly rely on `$this->get_order()`. The only difference is that we want to default to `DESC` here instead of `ASC` that exists in the `get_order()` method.

I think it might be worth overriding the `get_order()` method in this class and defaulting to `DESC`. Alternatively, have some kind of default value for `order` in the class that can be overriddent in a child class, then we can just use the method directly.

src/Common/Abstracts/Custom_List_Table.php
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	/**
+	 * Returns the instance arguments.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	abstract protected function get_instance_args(): array;
```
JPry: "Instance arguments" seems a little vague. Could more explanation be added to clarify exactly what that means? Or perhaps a different name?

src/Common/Abstracts/Custom_Table_Repository.php
```diff
@@ -0,0 +1,1081 @@
...
+namespace TEC\Common\Abstracts;
+
+use TEC\Common\Contracts\Custom_Table_Repository_Interface as Repository_Interface;
+use TEC\Common\StellarWP\SchemaModels\Contracts\SchemaModel as Model;
+use TEC\Common\StellarWP\Schema\Tables\Contracts\Table as Table_Interface;
+use TEC\Common\StellarWP\SchemaModels\Relationships\ManyToManyWithPosts;
+use TEC\Common\StellarWP\DB\DB;
+use Tribe__Promise as Promise;
+use RuntimeException;
+use Generator;
```
JPry: 2 small things:

1. Alphabetize the imports. This applies to all files.
2. Include `WP_Post` as an import. Not strictly needed for the code, but `WP_Post` _is_ referenced in docblocks.

```suggestion
use Generator;
use WP_Post;
```

src/Common/Abstracts/Custom_Table_Repository.php
```diff
@@ -0,0 +1,1081 @@
...
+					$this->add_schema_entry(
+						$column,
+						function ( $value ) use ( $column, $operator ) {
+							return [
+								'column'   => $column,
+								'value'    => $value,
+								'operator' => $operator,
+							];
+						}
+					);
```
JPry: Should there be a `continue` statement here? Or do we intend to add a second schema entry?

src/Common/Abstracts/Custom_Table_Repository.php
```diff
@@ -0,0 +1,1081 @@
...
+					}
+				);
+			}
+		}
+
+		$model = tribe( $this->get_model_class() );
+
+		$relationships = $model->getRelationshipCollection();
+
+		foreach ( $relationships->getAll() as $key => $relationship ) {
```
JPry: This is getting to be a rather large block of code, especially for one method. I think it would be better to move this block and the block above into separate methods.

src/Common/Contracts/Provider/Controller.php
```diff
@@ -12,6 +12,7 @@
 
 use TEC\Common\Contracts\Service_Provider;
 use Tribe__Log as Log;
+use TEC\Common\Contracts\Container;
```
JPry: Alphabetize :slightly_smiling_face:

src/Common/Contracts/Repository_Interface.php
```diff
@@ -0,0 +1,176 @@
...
+	public function set_update_fields_aliases( array $update_fields_aliases );
+
+	/**
+	 * Creates a post of the type managed by the repository with the fields
+	 * provided using the `set` or `set_args` methods.
+	 *
+	 * @since 4.9.5
+	 * @since TBD Added Model as a possible return type.
+	 *
+	 * @return WP_Post|Model|false|null The created post object or `false` if the creation
```
JPry: The comment mentions objects or `false`, but what about returning `null`?

Do we need an `@throws` tag?

src/Common/Libraries/Shepherd.php
```diff
@@ -38,6 +43,34 @@ public function do_register(): void {
...
+					<?php // Translators: %s is the query error message. ?>
+					<p><?php printf( esc_html__( 'The below query failed with the message(s): %s', 'tribe-common' ), '<code>' . esc_html( implode( '<br>', $error->getQueryErrors() ) ) . '</code>' ); ?></p>
+					<p><code><?php echo esc_html( $error->getQuery() ); ?></code></p>
+				</div>
+				<?php
+			}
+		)
+			->urgency( 'error' )
+			->dismissible( false )
+			->inline( true );
```
JPry:
```suggestion
			->inline();
```

src/Common/REST/TEC/V1/Abstracts/Post_Entity_Endpoint.php
```diff
@@ -169,7 +169,7 @@ public function get_post_type_object(): WP_Post_Type {
 	 *
 	 * @return array
 	 */
-	protected function format_post_entity_collection( array $posts ): array {
+	protected function format_entity_collection( array $posts ): array {
```
JPry: Do we need to account for the old method name?

Also, we should update the `@since` tag to flag that the method was renamed.

src/Common/REST/TEC/V1/Contracts/Post_Entity_Endpoint_Interface.php
```diff
@@ -76,7 +76,7 @@ public function get_formatted_entity( WP_Post $post ): array;
 	 *
 	 * @since 6.9.0
 	 *
-	 * @return Tribe__Repository__Interface
+	 * @return Repository_Interface
 	 */
-	public function get_orm(): Tribe__Repository__Interface;
+	public function get_orm(): Repository_Interface;
```
JPry: The `@since` tag needs updating with the new return type.

src/Common/REST/TEC/V1/Traits/Read_Custom_Archive_Response.php
```diff
@@ -0,0 +1,129 @@
...
+	 *
+	 * @param array $params The sanitized parameters to use for the request.
+	 *
+	 * @return WP_REST_Response The response object.
+	 */
+	public function read( array $params = [] ): WP_REST_Response {
+		$page     = absint( $params['page'] ?? 1 );
+		$per_page = absint( $params['per_page'] ?? $this->get_default_posts_per_page() );
+
+		/** @var Repository_Interface $query */
```
JPry: We don't need this, since `build_query()` has it documented as a return type.

```suggestion
```

src/Common/REST/TEC/V1/Traits/Read_Custom_Archive_Response.php
```diff
@@ -0,0 +1,129 @@
...
+
+		/**
+		 * Filters the data that will be returned for an entities archive request.
+		 *
+		 * @since TBD
+		 *
+		 * @param array $data   The retrieved data.
+		 * @param array $params The sanitized parameters to use for the request.
+		 */
+		$data = apply_filters( 'tec_rest_' . $this->get_model_class() . '_archive', $data, $params );
```
JPry: 2 things:

1. Add `get_model_class()` as an abstract.
2. String interpolation is easier to read, IMO

```suggestion
		$data = apply_filters( "tec_rest_{$this->get_model_class()}_archive", $data, $params );
```

src/Common/REST/TEC/V1/Traits/Read_Custom_Archive_Response.php
```diff
@@ -0,0 +1,129 @@
...
+
+		$total_pages = $per_page > 0 ? (int) ceil( $total / $per_page ) : 1;
+		$current_url = $this->get_current_rest_url();
```
JPry: Another method to add as an abstract requirement.

src/Common/REST/TEC/V1/Traits/Read_Custom_Entity_Response.php
```diff
@@ -0,0 +1,58 @@
...
+		if ( ! $id ) {
+			return new WP_REST_Response(
+				[
+					'error' => __( 'The entity could not be read.', 'tribe-common' ),
+				],
+				404
+			);
+		}
+
+		$entity = $this->get_orm()->by_primary_key( $id );
```
JPry: Add `get_orm()` as an abstract.

src/Common/REST/TEC/V1/Traits/Read_Custom_Entity_Response.php
```diff
@@ -0,0 +1,58 @@
...
+		if ( ! $entity ) {
+			return new WP_REST_Response(
+				[
+					'error' => __( 'The entity could not be read.', 'tribe-common' ),
+				],
+				404
+			);
+		}
+
+		return new WP_REST_Response( $this->get_formatted_entity( $entity ), 200 );
```
JPry: Add `get_formatted_entity()` as an abstract.

JPry (changes requested):

src/Common/Abstracts/Custom_List_Table.php
dpanta94: I chose to accept a $default arg in the methods to resolve this.

src/Common/Abstracts/Custom_Table_Repository.php
```diff
@@ -0,0 +1,1081 @@
...
+	 *
+	 * @param string $field The field to pluck from each result.
+	 *
+	 * @return array The plucked models.
+	 */
+	public function pluck( $field ): array {
+		$method  = 'get_' . $field;
+		$results = [];
+		foreach ( $this->all( true ) as $model ) {
+			$results[] = $model->$method();
```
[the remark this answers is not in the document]
dpanta94: I chose to be loud about this. StellarWP\SchemaModel supports magic methods with `get_` and `set_` for all their columns and relationships. So if we end up here having a not available method, something has gone very wrong.

src/Common/Contracts/Repository_Interface.php
dpanta94: I think the null was just a mistake. i couldn't see it so i removed it. I added the throws tag.

src/Common/Libraries/Shepherd.php
```diff
@@ -24,8 +26,11 @@ class Shepherd extends Controller_Contract {
 	 * @since 6.9.0
 	 */
 	public function do_register(): void {
+		$hook_prefix = tribe( Libraries_Provider::class )->get_hook_prefix();
 		Config::set_container( $this->container );
```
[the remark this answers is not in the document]
dpanta94: It's not correct. That would have thrown a fatal. php storm.... :P

src/Common/REST/TEC/V1/Abstracts/Custom_Entity_Endpoint.php
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Returns whether the user can read the object.
+	 *
+	 * @since TBD
+	 *
+	 * @param WP_REST_Request $request The request object.
+	 *
+	 * @return bool
+	 */
+	public function can_read( WP_REST_Request $request ): bool {
```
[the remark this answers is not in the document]
dpanta94: We don't but it may not be visible that its being passed to the method if i remove it. I prefer it there honestly.

src/Common/REST/TEC/V1/Abstracts/Post_Entity_Endpoint.php
dpanta94: No, anything in the new REST API we declare that backwards compatibility is not a guarantee.

src/Common/Abstracts/Custom_List_Table.php
JPry: Perhaps something like `get_object_query_args()`? Since it's used as part of the querying process, that seems to make the most sense to me.

Will all child classes definitely need to have a custom array of arguments? If not, I would recommend changing this method to be a normal `protected` method that returns an empty array. Child classes that need to override can still do so.

JPry (changes requested):
Just a couple small things left

PR #2746, opened 2025-09-04
Style - add error color class.

rafsuntaskin:
* Add a class for error color.

### DIFF
--- a/src/resources/packages/classy/style.pcss
+++ b/src/resources/packages/classy/style.pcss
@@ -744,6 +744,11 @@
 	width: 24px;
 }
 
+.classy-error-color {
+	color: var(--tec-color-gutenberg-alert-red);
+	fill: var(--tec-color-gutenberg-alert-red);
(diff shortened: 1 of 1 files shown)


JPry (dismissed):
This is good, but I think we should add our own `--classy-*` variable that maps to this color in the `:root` group. For example, in ET I used `--classy-color-destructive` mapped to this color. Maybe something more appropriate for its usage as an error color would be good.

rafsuntaskin:
> This is good, but I think we should add our own `--classy-*` variable that maps to this color in the `:root` group. For example, in ET I used `--classy-color-destructive` mapped to this color. Maybe something more appropriate for its usage as an error color would be good.

@JPry That's indeed a more structured approach, but it seems like an overkill for now. We already have a similar accent-color class and are just following the similar pattern. Will keep this in mind if we need more variables like this but for this instance I am going to dismiss the review.

PR #2780, opened 2025-09-25
TEC-5583 Don't show opt in modal if onboarding is completed

sdokus:
### 🎫 Ticket

[TEC-5583]

### 🗒️ Description

The Telemetry opt-in is being prompted in the onboarding wizard, so if a user completes the onboarding wizard, we should assume that they have opted out of Telemetry and not re-prompt them when they visit the Settings page. 

This PR adds logic to achieve that, while still accounting for if the user clicks `Skip guided setup`, the modal should still pop up because we are not sure if they consciously chose to opt out or not. 

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s). More info [here](https://docs.theeventscalendar.com/developer/git/changelogs/#process)
- [ ] Code is covered by **NEW** `wpunit` or `integration` tests.
- [ ] Code is covered by **EXISTING** `wpunit` or `integration` tests.
- [ ] Are all the **required** tests passing?
- [ ] Automated code review comments are addressed.
- [ ] Have you added Artifacts?
- [x] Check the base branch for your PR.
- [x] Add your PR to the project board for the release.

PR #2787, opened 2025-10-01
TEC-5583 Don't set opt-in option prematurely

sdokus:
### 🎫 Ticket

[TEC-5583]

### 🗒️ Description

This PR removes a call in the IAN notifications that was unintentionally setting the opt-in option to 'false' even if the user had not been shown any Telemetry opt-in modals yet. This, in turn, had been causing the logic in `calculate_modal_status` to incorrectly show the modal after the user had completed the Onboarding wizard because the setting was not updated since the wizard saw that value as already being set. 

This PR is preferred over https://github.com/the-events-calendar/tribe-common/pull/2780 because it addresses the root of the issue instead of just acting as a workaround. This PR is also working in tandem with a PR in TEC (which I will open shortly and then link here). 

### 🎥 Artifacts 

https://drive.google.com/file/d/1C3xiTotOaGLWQ31GMOxzjuZNbBhTpLVr/view?usp=sharing

### ✔️ Checklist

[TEC-5583]: https://stellarwp.atlassian.net/browse/TEC-5583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Notifications/Conditionals.php
+++ b/src/Common/Notifications/Conditionals.php
@@ -37,7 +37,7 @@ class Conditionals {
 		 *
 		 * @param bool $opted_in Whether the user has opted in to the IAN notifications.
 		 */
-		return (bool) apply_filters( 'tec_common_ian_opt_in', (bool) self::opt_in_is_active() );
+		return (bool) apply_filters( 'tec_common_ian_opt_in', self::opt_in_is_active() );
 	}
 
 	/**


src/Common/Telemetry/Telemetry.php
```diff
@@ -587,20 +589,22 @@ public function calculate_optin_status( $opted = null ) {
...
 		$status = array_filter( $stati );
+
 		return (bool) array_pop( $status );
 	}
 
 	/**
 	 * Calculate the optin status for the TEC plugins from various sources.
 	 *
 	 * @since 5.1.1.1
+	 * @since TBD Change bail check to check if anything has been set, not just if the value istruthy.
```
Camwyn, line 600:
```suggestion
	 * @since TBD Change option check to prevent false negatives when the option is `false` (user has opted out).
```

Camwyn (changes requested):
Docblock change

dpanta94 (changes requested):

dpanta94 (approved):

PR #2876, opened 2026-04-20
Feat/consolidated licensing

dpanta94:
### 🎫 Ticket

[SCON-651]

### ✔️ Checklist

[SCON-651]: https://stellarwp.atlassian.net/browse/SCON-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Common/Integrations/Harbor/EventAggregator.php
+++ b/src/Common/Integrations/Harbor/EventAggregator.php
@@ -9,7 +9,8 @@
 
 namespace TEC\Common\Integrations\Harbor;
 
-// phpcs:ignore StellarWP.Classes.ValidClassName.NotSnakeCaseClass
+// phpcs:disable StellarWP.Classes.ValidClassName.NotSnakeCase
+
 
 /**
  * The Event Aggregator Harbor integration.
--- a/src/Common/Integrations/Harbor/EventAggregator.php
+++ b/src/Common/Integrations/Harbor/EventAggregator.php
@@ -27,7 +27,7 @@ class EventAggregator extends Integration_Controller {
 	 * @return void
 	 */
 	protected function do_register(): void {
-		add_filter( 'tec_events_aggregator_consolidation_took_over', [ $this, 'filter_consolidation_took_over' ] );
+		add_filter( 'tec_events_aggregator_harbor_took_over', [ $this, 'filter_harbor_took_over' ] );
 	}
 
 	/**
@@ -38,7 +38,7 @@ class EventAggregator extends Integration_Controller {
 	 * @return void
 	 */
 	public function unregister(): void {
-		remove_filter( 'tec_events_aggregator_consolidation_took_over', [ $this, 'filter_consolidation_took_over' ] );
+		remove_filter( 'tec_events_aggregator_harbor_took_over', [ $this, 'filter_harbor_took_over' ] );
 	}
 
 	/**
@@ -48,7 +48,7 @@ class EventAggregator extends Integration_Controller {
 	 *
 	 * @return bool
 	 */
-	public function filter_consolidation_took_over(): bool {
+	public function filter_harbor_took_over(): bool {
 		return (bool) $this->harbor->get_unified_license_key_if_feature_enabled( 'event-aggregator' );
 	}
 }
--- a/src/Common/Integrations/Harbor/EventAggregator.php
+++ b/src/Common/Integrations/Harbor/EventAggregator.php
@@ -17,7 +17,7 @@ use stdClass;
  *
  * @package TEC\Common\Integrations\Harbor
  */
-class EA extends Integration_Controller {
+class EventAggregator extends Integration_Controller {
 	/**
 	 * Register the controller.
 	 *
@@ -26,7 +26,6 @@ class EA extends Integration_Controller {
 	 * @return void
 	 */
 	protected function do_register(): void {
-		add_filter( 'tribe_aggregator_api', [ $this, 'filter_aggregator_api' ] );
 		add_filter( 'tec_events_aggregator_consolidation_took_over', [ $this, 'filter_consolidation_took_over' ] );
 	}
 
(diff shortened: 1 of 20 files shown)


src/Common/Integrations/Harbor/PUE.php
```diff
@@ -0,0 +1,301 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param false|array|WP_Error $response    The response.
+	 * @param array                $parsed_args The parsed arguments.
+	 * @param string               $url         The URL.
+	 *
+	 * @return false|array
+	 */
+	public function filter_pre_http_request( $response, array $parsed_args, string $url ) {
```
johnhooks, line 132: Couldn't we use this hook?

https://github.com/stellarwp/uplink/blob/ba3fbaf406f5473c3a94d4d1b0aeeef93e78c4b6/src/Uplink/Resources/Plugin.php#L47-L50

johnhooks: Perhaps it doesn't cover all the needs. I find filtering on pre_http_request to be pretty heavy handed, though perhaps necessary without changes to Uplink.

src/Common/Integrations/Harbor/EventAggregator.php
```diff
@@ -0,0 +1,54 @@
...
+class EventAggregator extends Integration_Controller {
+	/**
+	 * Register the controller.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
+		add_filter( 'tec_events_aggregator_consolidation_took_over', [ $this, 'filter_consolidation_took_over' ] );
```
johnhooks:
```suggestion
		add_filter( 'tec_events_aggregator_harbor_took_over', [ $this, 'filter_harbor_took_over' ] );
```

If we name it "harbor" that would match up with what is taking over, we will forget that this was the "consolidation" project.

dpanta94: Fair point, will amend that

src/Common/Integrations/Harbor/PUE.php
dpanta94: Well the pre_http_request is required to take over in all communicaitons with licensing. Specifically for uplink i use `stellarwp/uplink/tec/license_get_key` to filter the stored license key to be the consolidated one instead.

Regarding your concern about it being heavy handed, i think the early bail logic that is in place would make the added filter almost adding no load on when its not supposed to take over.

src/Common/Libraries/Harbor.php
```diff
@@ -0,0 +1,250 @@
...
+			$filters_removed = true;
+		}
+
+		$slugs_added = [];
+
+		foreach ( $plugins as $plugin ) {
+			$license_object = $plugin->get_license_object();
+			$licenses[]     = [
+				'key'        => $license_object->get_key(),
+				'slug'       => $this->get_harbor_product_slug( $plugin->get_slug() ),
```
jonwaldstein, line 115: it doesn't look like calling `get_harbor_product_slug` is really necessary?

src/Common/Libraries/Harbor.php
```diff
@@ -0,0 +1,250 @@
...
+	 * @param string $product The product slug.
+	 *
+	 * @return bool
+	 */
+	public function is_product_licensed( string $product ): bool {
+		if ( ! lw_harbor_has_unified_license_key() ) {
+			return false;
+		}
+
+		return lw_harbor_is_feature_enabled( $product );
```
jonwaldstein: I think you're looking for either `lw_harbor_is_product_license_active` or `lw_harbor_is_feature_available`.

- `lw_harbor_is_product_license_active` would be the main product (TEC)

- `lw_harbor_is_feature_available` would be _features_ included in the plan that are available to you.  enabled is specifically if someone has the feature enabled/activated on their site but _available_ means included in their license plan.

You could even use a combo of the two if need be.

src/Common/Libraries/Harbor.php
```diff
@@ -0,0 +1,250 @@
...
+	 * Get the unified license key.
+	 *
+	 * @since TBD
+	 *
+	 * @return string|null The unified license key, or null if no key is found.
+	 */
+	public function get_unified_license_key(): ?string {
+		if ( ! lw_harbor_has_unified_license_key() ) {
+			return null;
+		}
```
jonwaldstein: `lw_harbor_get_unified_license_key` already handles this so not really necessary

src/Common/Libraries/Harbor.php
dpanta94: Why not ?

src/Common/Libraries/Harbor.php
dpanta94: Definitely not `lw_harbor_is_product_license_active` since i do wan to evaluate individual features.

In this case, and from my testing, enabled and available yield exactly the same results.

A not enabled product won't trigger this method. At the same time, services seem to answer whether they are available either way.

src/Common/Integrations/Harbor/PUE.php
```diff
@@ -0,0 +1,301 @@
...
+use TEC\Common\StellarWP\Uplink\Resources\Resource as Uplink_Resource;
+
+/**
+ * The PUE Harbor integration.
+ *
+ * @since TBD
+ *
+ * @package TEC\Common\Integrations\Harbor
+ */
+class PUE extends Integration_Controller {
```
jonwaldstein, line 29: @dpanta94 can you explain why this controller is necessary? maybe you have some requirements that i'm not aware of, but ideally you would leave uplink logic alone and let harbor live alongside it.

src/Common/Libraries/Harbor.php
jonwaldstein: Oh sorry it looked like there was no difference in slugs at first glance, now I see promoter and seating are different. Do you know if that slug change was intentional?

src/Common/Integrations/Harbor/PUE.php
dpanta94: What does this controller do:

1. It intercepts any calls to licensing v2 validate endpoint for products that are INCLUDED already by harbor. That saves a lot of traffic, since that product that would be validated is already validated by harbor. I also let harbor manage my caching fully.

How does it do that ?

It adds a generic pre_http_request filter.

2. It hooks into the pre_option and uplink's get license key functions in order to return the unified license key for products that are already authorized and valid for this site via harbor.

Why ?

It does that, to prevent multiple code changes against multiple codebases where these options are used to communicate with licensing or TEC services.

For example, Event aggregator expects the license to be sent when TEC plugins communicate with it. Having the pre_option hook in place ensures that the unified license is passed to it IF EA is provided by it, without affecting existing code.

src/Common/Libraries/Harbor.php
dpanta94: I dont im afraid :(

src/Common/Libraries/Harbor.php
dpanta94: i replaced with lw_harbor_is_feature_available - just for clarity

src/Common/Integrations/Harbor/PUE.php
jonwaldstein: Thanks, I think we concluded that you can choose your own adventure per-product and TEC is the most tied in with uplink so do what you think is best here.

PR #2916, opened 2026-06-11
Update required and add tested up to versions

sdokus:
[skip-changelog][skip-phpcs]


tribe-common.php
```diff
@@ -3,7 +3,8 @@
  * Plugin Name: Tribe Common
  * Description: An event settings framework for managing shared options.
  * Version: 6.11.4
- * Requires at least: 6.7
+ * Tested up to: 7.0
```
dpanta94: I think tested up to goes into the readme.txt

dpanta94 (approved):

PR #2924, opened 2026-06-16
Fix tests in `bucket/rsvp-merge` branch

sdokus:
_Ticket: TBD — I can add one if we want to track this_

### 🗒️ What this fixes

Two changes, both aimed at getting the test suites to run reliably again on this branch.

**1. The tests couldn't even start.**

The migrations library we recently pulled in quietly depends on a small package, `myclabs/php-enum`. That package needs a tiny helper file (a `Stringable` shim for older PHP versions), but it accidentally leaves that file out of the copy we download by default. Our build step then can't find the file and stops partway through, so the prefixed Common classes never get generated. The end result: any plugin that loads Common crashes on startup with `Interface ... Template_Engine not found`, and no tests run at all.

The fix recreates that one missing file via a small patch — using the very same patch mechanism we already use for the Faker library.

**2. A test that randomly passed or failed.**

A Repository test checked that a query returned its results in one exact order. But that query never asks the database for a particular order, so the database is free to hand the results back in any order — which made the test pass or fail at random. I changed those checks to compare the results regardless of order, which is what the test actually means to verify. (This only surfaced now because change #1 is what lets these tests run again in the first place.)

> ⚠️ **Heads up: #1 is a workaround, not the root cure.** The real bug is in how the `php-enum` package publishes its releases — it points to a file it doesn't actually ship. The proper long-term fix belongs upstream, or in having the migrations library not depend on `php-enum`. Calling this out because the workaround needs to travel with the migrations work when it merges into `main`: `main` doesn't use that library yet, so it isn't broken there today, but it will be the moment this merges forward.

### ✔️ How to test

1. Delete `vendor/` and run `composer install` (with dev dependencies) on this branch.
   - **Before:** the build stops with an error about `php-enum/stubs/Stringable.php`.
   - **After:** the build finishes cleanly and the migrations classes are present.
2. Run the test suites — they start up and run (instead of crashing immediately), and the Repository test no longer fails intermittently on ordering.

[skip-changelog]
[skip-phpcs]

### DIFF
--- a/tests/wpunit/Tribe/RepositoryTest.php
+++ b/tests/wpunit/Tribe/RepositoryTest.php
@@ -172,13 +172,13 @@ class RepositoryTest extends WPTestCase {
 			$this->repository()->where( 'meta_in', 'author_name', [ 'Lewis Carroll' ] )->get_ids(),
 			"meta_in should support single values with space"
 		);
-		$this->assertEquals(
+		$this->assertEqualSets(
 			[ $book_2_id, $book_3_id ],
 			$this->repository()->where( 'meta_not_in', 'author_name', 'Lewis Carroll' )->get_ids(),
 			"meta_not_in should support single values with space"
 		);
 
-		$this->assertEquals(
+		$this->assertEqualSets(
 			[ $book_2_id, $book_3_id ],
 			$this->repository()->where( 'meta_not_in', 'author_name', [ 'Lewis Carroll' ] )->get_ids(),
 			"meta_not_in should support single values with spaces in array"
@@ -193,7 +193,7 @@ class RepositoryTest extends WPTestCase {
 
 		$for_kids_term_id = get_term_by( 'name', 'for kids', 'audience' )->term_id;
 
-		$this->assertEquals(
+		$this->assertEqualSets(
 			[ $book_1_id, $book_2_id ],
 			$this->repository()->where( 'term_id_in', 'audience', $for_kids_term_id )->get_ids(),
 			"term_id_in should support single values"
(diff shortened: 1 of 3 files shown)


tests/wpunit/Tribe/RepositoryTest.php
```diff
@@ -172,13 +172,13 @@ public function should_build_meta_in_not_in_query_correctly(): void {
 			$this->repository()->where( 'meta_in', 'author_name', [ 'Lewis Carroll' ] )->get_ids(),
 			"meta_in should support single values with space"
 		);
-		$this->assertEquals(
+		$this->assertEqualSets(
```
estevao90: Sorry, I don't know the project, but I did not understand why we need to remove the strict comparison here. Could you please explain it?

sdokus: Good question! This test is just checking that the correct set of IDs is in the array, but doesn't care about the order - this test was randomly failing just because the array would sometimes return the IDs in a different order than the expected results, which isn't a real failure.

tests/wpunit/Tribe/RepositoryTest.php
```diff
@@ -172,13 +172,13 @@ public function should_build_meta_in_not_in_query_correctly(): void {
...
 			[ $book_2_id, $book_3_id ],
 			$this->repository()->where( 'meta_not_in', 'author_name', 'Lewis Carroll' )->get_ids(),
 			"meta_not_in should support single values with space"
 		);
 
-		$this->assertEquals(
+		$this->assertEqualSets(
```
dpanta94: Maybe specify order by instead of just comparing the Set ?

The fact that it can return in different order from time to time is weird to me. If it would fail even when order by is specified, i would look into race conditions

dpanta94 (changes requested):

dpanta94 (approved):
Amazing work @sdokus 🚀 !

PR #2929, opened 2026-06-24
[SMTNC-1646] PHP Compatibility improvements

dpanta94:
### 🎫 Ticket

SMTNC-1646

### 🗒️ Description

Improves php compatibility for php version from 7.4 up to 8.4.

Breaking contract by adding explicitly nullable `?` is not a breaking change.

PHPCompatibility and PHPCompatibilityWP are included via the `sh` script because the latest stable version of PHPCompatibilityWP does not cover 8+ compatibility concerns.

### 🎥 Artifacts 
Tests continue passing are part of the artifact.

Test ensuring that the cotract being broken is not considered a breaking change.

CleanShot 2026-06-25 at 00 33 40

### ✔️ Checklist


src/Tribe/Utils/Collection_Interface.php
```diff
@@ -16,7 +16,7 @@
  * @since 4.9.14
  * @package Tribe\Utils
  */
-interface Collection_Interface extends \ArrayAccess, \SeekableIterator, \Countable, \Serializable, \JsonSerializable {
+interface Collection_Interface extends \ArrayAccess, \SeekableIterator, \Countable, \JsonSerializable {
```
defunctl, line 19: Again, we should really look around and see what is implementing this interface and checking for `$whatever instanceof \Serializable`...

src/Tribe/Log/File_Logger.php
```diff
@@ -155,7 +155,7 @@ public function log( $entry, $type = Tribe__Log::DEBUG, $src = '' ) {
 			return;
 		}
 
-		fputcsv( $this->handle, [ date_i18n( 'Y-m-d H:i:s' ), $entry, $type, $src ] );
+		fputcsv( $this->handle, [ date_i18n( 'Y-m-d H:i:s' ), $entry, $type, $src ], ',', '"', '\\' );
```
defunctl: Same for this and below, I think `\\` should be an empty string? Would really need to test it.

https://www.php.net/manual/en/function.fputcsv.php

src/Common/Site_Health/Fields/Post_Status_Count_Field.php
```diff
@@ -25,12 +25,13 @@ class Post_Status_Count_Field extends Generic_Info_Field {
 	/**
 	 * Configure all the params for a generic field.
 	 *
+	 * @since TBD Made second argument $post_type explicity nullable.
```
defunctl:
```suggestion
	 * @since TBD Made second argument $post_type explicitly nullable.
```

defunctl (changes requested):
I would say in general, many of these changes could cause problems for 3rd party implementations where someone extended a class and they no longer have the same method signatures or wrapped different functions, e.g. things like `src/Tribe/Process/Handler.php` that have abstract methods that have been changed etc...

There is just no way to know for sure what exists out there, so regardless all of these needs to be released with caution/warning maybe a long lived beta or something?

src/Tribe/Utils/Collection_Interface.php
dpanta94: Nothing does that in our codebase. I think if that is used externally it would be very very limited. I dont think it should be blocking us.

.github/workflows/csv-fputcsv-compat.yml
```diff
@@ -0,0 +1,45 @@
...
+  fputcsv-snapshot:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version: [ '7.4', '8.0', '8.1', '8.2', '8.3', '8.4' ]
+    name: PHP ${{ matrix.php-version }}
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
```
estevao90:
```suggestion
        uses: actions/checkout@v6
```

defunctl (approved):

PR #2960, opened 2026-07-27
Legacy license field improvements

pramodjodhani:
### 🎫 Ticket

[SMTNC-1862](https://linear.app/nexcess/issue/SMTNC-1862/tec-tec-license-page-improvements)

### 🗒️ Description

1. If unified license key is set, the legacy license fields are made readonly and disabled.
2. Warn user away from entering unified license key to the legacy license fields. Guide to the LW license manager.
3. Fix bug where incorrect message is show when user tried to enter legacy license when already unified license key is set.

### 🎥 Artifacts 

https://www.loom.com/share/94f08029f9fc4ac0a4a1f08915e1e778

### ✔️ Checklist


src/Common/Libraries/Harbor.php
```diff
@@ -265,6 +274,53 @@ public function is_product_licensed( string $product ): bool {
...
+	 * @since 6.12.0
+	 *
+	 * @param string $tec_product_slug The TEC product slug.
+	 *
+	 * @return bool
+	 */
+	public function is_license_field_managed_by_harbor( string $tec_product_slug ): bool {
+		return $this->is_product_licensed(
+			$this->get_harbor_product_slug( $tec_product_slug )
+		);
```
d4mation: This specific check could be risky, right? It just checks whether their Unified Key _covers_ the Product Slug. Not whether it is _active_ for their site.

If they choose specifically not to activate the Product License on their site via their Unified Key, then they should be able to enter a Legacy Key still.

`lw_harbor_is_product_license_active()` may be the correct underlying function to call here. `$this->is_product_license()` calls `lw_harbor_is_feature_available()` under the hood which is a different check.

pramodjodhani: @d4mation I have made this change [`07c8007` (this PR)](https://github.com/the-events-calendar/tribe-common/pull/2960/commits/07c8007e059f38f11b49174402e69c4973c9cabf)

Questions: 

1. Are entitlements same as capabilities? Do we need to to rename the function?
2. Do you think this function could be helpful for other plugins as well and can be added to Harbor's global functions?

d4mation: Ah, good catch on this! I was working on LearnDash Core recently where I used `lw_harbor_is_product_license_active()` and it just _happened_ to be the same slug as the Product in Portal+Licensing. 

In this case, you need to be checking for Capabilities (plugins, services, etc.) that are provided via the Product. What you have done looks correct from a logic standpoint, but it will *_need to be a_* global function in Harbor to avoid calling the version frozen within Tribe Common. https://github.com/stellarwp/harbor/blob/main/docs/architecture/fat-leader-thin-instance.md

It might be a good idea to expose a few functions here: 

1. One to get all the Products from the License Repository
2. One that does what you're doing and checking if a given Capability Slug is _*covered*_ by the entered Unified Key, whether or not the Product is active for the domain
3. One that does the same as above, but also checks whether the Product is active for the domain (exactly what you're doing)

pramodjodhani: @d4mation Sure, I'm on it. Do these names make sense to you:

```
lw_harbor_get_licensed_products()
lw_harbor_is_capability_licensed( string $capability_slug ): bool
function lw_harbor_is_capability_license_active( string $capability_slug ): bool
```

d4mation: I think we should follow some existing naming patterns:

- `lw_harbor_is_capability_license_available( string $slug ): bool`
- `lw_harbor_is_capability_license_active( string $slug ): bool`

But I otherwise prefer your simplification here :) 

I am less sure about `lw_harbor_get_licensed_products()` though. Unless you think you need it, we should skip it for now. We would need to flatten it to an array of string slugs to make it make sense in a multi-Harbor environment because each version of Harbor would be expecting a different Strauss Prefix on whatever Object it would normally want to return.

pramodjodhani: @d4mation [Done](https://github.com/the-events-calendar/tribe-common/pull/2960/commits/c401e0279a091060bd58be54fe49095dd236db70) and [Done](https://github.com/stellarwp/harbor/pull/187)

src/Common/Libraries/Harbor.php
```diff
@@ -265,6 +275,65 @@ public function is_product_licensed( string $product ): bool {
 		return lw_harbor_is_feature_available( $product );
 	}
 
+	/**
+	 * Whether a license key uses the unified Liquid Web format.
+	 *
+	 * @since 6.12.0
```
dpanta94, line 281: Prefer adding TBD for the version since we aren't sure the version that this code is going to be included in.

src/Common/Integrations/Harbor/PUE.php
```diff
@@ -56,6 +59,154 @@ public function unregister(): void {
...
+	/**
+	 * Get the stored product license key without Harbor option overlays.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $option_name The option name.
+	 *
+	 * @return string
+	 */
+	private function get_stored_product_license_key( string $option_name ): string {
```
dpanta94, line 196: Some type hinting would be nice in the methods of this class.

src/Tribe/PUE/Checker.php
```diff
@@ -1238,6 +1238,23 @@ public function validate_key( string $key, bool $network = false ): array {
...
+			 * Returning a non-null array short-circuits remote validation and is used as
+			 * the AJAX/validation response. Return null to continue normal validation.
+			 *
+			 * @since 6.12.0
+			 *
+			 * @param array|null          $response Early response, or null to continue.
+			 * @param string              $key      The license key being validated.
+			 * @param Tribe__PUE__Checker $checker  The PUE checker instance.
+			 */
+			$pre_validate = apply_filters( 'tribe_puc_pre_validate_key', null, $key, $this );
```
dpanta94, line 1253: Probably a typo in the filter name and using legacy prefix.

Prefer: `tec_common_pue` over `tribe_puc` :D

composer.json
```diff
@@ -31,7 +31,7 @@
     "stellarwp/assets": "^1.5",
     "stellarwp/container-contract": "^1.0.4",
     "stellarwp/db": "^1.0.3",
-    "stellarwp/harbor": "^1.4",
+    "stellarwp/harbor": "dev-SMTNC-1895",
```
dpanta94, line 34: Lets make sure we release new version of harbor before merging so we dont ship a dev version.

src/resources/postcss/utilities
dpanta94, line 1: Make sure you update common through its own PR - not by a PR in tec or et.

dpanta94 (changes requested):

src/Common/Libraries/Harbor.php
```diff
@@ -265,6 +275,65 @@ public function is_product_licensed( string $product ): bool {
...
+	/**
+	 * Whether a TEC product license field is managed by Harbor.
+	 *
+	 * When true, the unified license key should be shown read-only and should not
+	 * be validated through legacy PUE per-product fields.
+	 *
+	 * Checks that the product's license is actually active/valid for this site, not
+	 * merely that the customer's tier entitles them to it — a customer can be entitled
+	 * to a product without having activated it here, in which case a legacy per-product
+	 * key should still be accepted.
```
johnhooks, line 300: Sweet, this was my question in your Loom @pramodjodhani this is what I would expect.

johnhooks (approved):
@pramodjodhani I approve. From my perspective this handles the juggling of unified and stellar license key correctly. 

Though please get full approval from @dpanta94 before merging, he knows the ins and outs of `tribe-common`, I do not.

src/Common/Libraries/Harbor.php
pramodjodhani: Good catch. I have added to to my claude skill as well so its not missed in future. Done here: 

https://github.com/the-events-calendar/tribe-common/pull/2960/commits/0b03b38b7b0e4027653b9273f4850bb632949456

src/Common/Integrations/Harbor/PUE.php
pramodjodhani: https://github.com/the-events-calendar/tribe-common/pull/2960/commits/0b03b38b7b0e4027653b9273f4850bb632949456

src/Tribe/PUE/Checker.php
pramodjodhani: Done. https://github.com/the-events-calendar/tribe-common/pull/2960/changes/e2231e84ef9cbbce0372ffcdaa1e3038f1d9fc19

src/resources/postcss/utilities
pramodjodhani: There is no change from my side here. I have restored the submodule.

src/Common/Libraries/Harbor.php
```diff
@@ -265,6 +275,62 @@ public function is_product_licensed( string $product ): bool {
...
+	 * @return bool
+	 */
+	public function is_license_field_managed_by_harbor( string $tec_product_slug ): bool {
+		if ( ! lw_harbor_has_unified_license_key() ) {
+			return false;
+		}
+
+		return lw_harbor_is_capability_license_active(
+			$this->get_harbor_product_slug( $tec_product_slug )
+		);
```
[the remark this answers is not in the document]
pramodjodhani: @dpanta94 I think this comment makes sense and should be implemented. But I want to confirm this with you, as its possible I might have not full context of why `is_product_licensed` was used in these places.  

`is_product_licensed` internally calls `lw_harbor_is_feature_available` which only checks if the feature is available in the tier, not whether its active or not.

src/Common/Libraries/Harbor.php
```diff
@@ -29,12 +30,21 @@
...
  */
 class Harbor extends Controller_Contract {
+	/**
+	 * Prefix for Liquid Web unified license keys.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const UNIFIED_LICENSE_KEY_PREFIX = 'LWSW-';
```
dpanta94: If this constant is not used externally to the class it should be private to allow us to deprecate it without warning.

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
```diff
@@ -351,4 +351,129 @@ public function auth_url_decorator_provider(): array {
...
+
+	/**
+	 * @test
+	 */
+	public function it_should_skip_remote_validation_for_harbor_managed_product(): void {
+		$this->seed_unified_license_key();
+		$this->seed_harbor_catalog_for_tec( [ 'events-calendar-pro' ] );
+
+		$checker  = new \Tribe__PUE__Checker( 'deprecated', 'events-calendar-pro', [], 'events-calendar-pro/events-calendar-pro.php' );
+		$response = $checker->validate_key( 'any-key-value' );
```
dpanta94, line 487: How do you actually validate that remote validation is not performed ? I would expect a spy on the function `wp_remote_request` asserting that it wasn;t called.

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
```diff
@@ -351,4 +351,129 @@ public function auth_url_decorator_provider(): array {
...
+
+	/**
+	 * @test
+	 */
+	public function it_should_pass_through_pre_validate_for_normal_key_on_non_managed_product(): void {
+		$this->seed_unified_license_key();
+		$this->seed_harbor_catalog_for_tec( [ 'events-calendar-pro' ] );
+
+		$checker = new \Tribe__PUE__Checker( 'deprecated', 'tribe-filterbar', [], 'the-events-calendar-filterbar/the-events-calendar-filterbar.php' );
+		$result  = apply_filters( 'tec_common_pue_pre_validate_key', null, 'legacy-product-key', $checker );
```
dpanta94: Show the other path as well. Here null remained null. When shouldn't that happen ?

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
```diff
@@ -351,4 +351,129 @@ public function auth_url_decorator_provider(): array {
...
+		$this->seed_unified_license_key();
+		$this->seed_harbor_catalog_for_tec( [ 'events-calendar-pro' ] );
+
+		$saved = apply_filters(
+			'tribe_settings_save_field_value',
+			'LWSW-SHOULD-NOT-BE-STORED',
+			'pue_install_key_events_calendar_pro'
+		);
+
+		$this->assertSame( 'legacy-ecp-key', $saved );
```
dpanta94: So the other path as well. When should it be stored ?

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
```diff
@@ -351,4 +351,129 @@ public function auth_url_decorator_provider(): array {
...
+					'type'       => 'license_key',
+					'attributes' => [],
+				],
+			]
+		);
+
+		$this->assertSame( 'disabled', $fields['pue_install_key_events_calendar_pro']['attributes']['disabled'] );
+		$this->assertSame( 'readonly', $fields['pue_install_key_events_calendar_pro']['attributes']['readonly'] );
+
+		$this->assertArrayNotHasKey( 'disabled', $fields['pue_install_key_tribe_filterbar']['attributes'] );
```
dpanta94, line 639: This is a complete assertion showing both paths! Nice!

dpanta94 (changes requested):

composer.json
pramodjodhani: This has been reverted back to `^1.4` since we dont need any new functions, we are now utilising the existing functions `lw_harbor_is_feature_enabled` 👍

src/Common/Libraries/Harbor.php
pramodjodhani: This is fixed.

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
pramodjodhani: This test is improved here: https://github.com/the-events-calendar/tribe-common/pull/2960/commits/90489d59415cbd78142c79b06e414c441061f977

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
pramodjodhani: Utilised With_Uopz::set_fn_return here to check if remote request is not made :https://github.com/the-events-calendar/tribe-common/pull/2960/commits/90489d59415cbd78142c79b06e414c441061f977

tests/integration/Tribe/Common/Integrations/Harbor/PUE_Test.php
pramodjodhani: We already had another test for the other path in `it_should_allow_normal_key_on_save_for_non_managed_license_field`. Though I have improved it further with dataProvider approach. 

https://github.com/the-events-calendar/tribe-common/pull/2960/commits/52d1051c17fddf1ece6e4a6c7c17db45ef717a99

pramodjodhani:
@dpanta94 This is ready for review again. 

There is a test failing, I checked it and found that the test script is crashing because the control goes to the exit statement here: https://github.com/the-events-calendar/the-events-calendar/blob/31784e78fbb240feca0cbfb506e7426de9c594b9/src/Events/Admin/Onboarding/Controller.php#L237

I confirmed this by commenting out the tribe_exit() function and then the test script was working fine except some failures unrelated to this PR (It_should_return_the_current_theme_object, It_should_correctly_identify_an_active_theme)

How do you think we should approach this solution? Maybe we should stub `tribe_exit` so it doesn't crash the whole script.

src/Common/Libraries/Harbor.php
```diff
@@ -119,6 +132,47 @@ public function decorate_uplinks_auth_url(): void {
...
+	 * @param \Tribe__PUE__Checker|null $checker  The PUE checker instance.
+	 *
+	 * @return array|null
+	 */
+	public function filter_tec_common_pue_pre_validate_key( ?array $response, string $key, $checker ): ?array {
+		if ( null !== $response ) {
+			return $response;
+		}
+
+		if ( did_action( 'lw_harbor/loaded' ) ) {
```
dpanta94, line 185: Shouldn't that be a negation?

pramodjodhani: No this is actually correct, this function's responsibility is to guide user when they enter a `LWSW-` key and we do **not** have Harbor loaded. 

We have a [different path](https://github.com/the-events-calendar/tribe-common/pull/2960/changes#diff-628293aa473247953c7f7fbd89f916b25fff0d7f4102531b8949f13a5c175d3bR139) which works when Harbor is loaded

pramodjodhani:
@dpanta94 because `tribe_exit` function silently exits the script, its hard to debug when the exit happens. I think it makes sense to disable it throughout the test suite. But let me know if you think otherwise.

dpanta94:
@pramodjodhani yes i do think otherwise. In tests environment we need to have control, mocking fully a function may mean that in the future we lose something we should catch.

Do you want to take a look [here](https://github.com/the-events-calendar/tribe-common/pull/2966/changes/3fca29f438be5dd52494625fb4162e4b9152232f) ? Im suspecting cherry picking that commit would fix the suite.

pramodjodhani:
@dpanta94 Fair enough. I have added that commit and removed my code from bootstrap.php. 

Everything seems to be in good shape now. Can you please confirm?

dpanta94 (approved):

PR #2966, opened 2026-08-04
SMTNC-1933: Remove the Stellar Sale conditional content

dave-green-uk:
### 🎫 Ticket

[SMTNC-1933](https://linear.app/nexcess/issue/SMTNC-1933/remove-stellar-sale-2026-from-tec-and-et-common)

### 🗒️ Description

Retires the Stellar Sale conditional content in Common.

**Why it resurfaced.** `Stellar_Sale` was date-bound to `July 29th` → `August 5th` with no year pinned, so the window re-opened every year. `get_slug()` resolves to `stellar-sale-<current year>`, but the creatives only ever shipped under `stellar-sale-2025/`. From 2026 onward the banner still passed its date check and rendered — with image URLs pointing at a directory that doesn't exist. That's what showed up while building the maintenance release.

**What changes:**

| | |
|---|---|
| `src/Common/Admin/Conditional_Content/Stellar_Sale.php` | **deprecated, not deleted** — see below |
| `src/Common/Admin/Conditional_Content/Controller.php` | singleton registration, `plugins_loaded()` init, header notice, promotional class list, help hub sidebar |
| `src/resources/images/conditional-content/stellar-sale-2025/` | 13 creatives |
| `src/resources/postcss/.../_main.pcss` | `.tribe-notice-stellar-sale` and `.tribe-notice-stellar-sale-2024` |
| tests / docs | the unused `$stellar_sale` fixture, a filter cleanup, and doc references |

**The class is deprecated, not removed** (review feedback — small chance a third party still resolves it). Every method emits `_deprecated_function()` and returns an inert value:

| method | returns |
|---|---|
| `should_display()` | `false` |
| `hook()` | no-op — registers nothing |
| `get_suite_creative_map()` | `[]` |
| `get_sale_name()` / `get_link_url()` / `get_creative_alt_text()` | `''` |
| `get_start_time()` / `get_end_time()` | `null` |
| `check_seating_license()` | `false` |

`should_display()` returning `false` is what stops every render path in `Promotional_Content_Abstract` — header notice, both banners, both sidebars and the tickets settings section all early-return. The traits stay in place so the inherited public API (`is_dismissible()`, `handle_dismiss()`, `check_capability()`, …) survives for anyone calling it. File-level `_deprecated_file()` follows the existing pattern in `Datetime_Conditional_Abstract.php`.

**What deliberately stays:**

- **`Has_Targeted_Creative_Upsell`** — still used by the deprecated class, and it's the documented counterpart to `Has_Generic_Upsell_Opportunity` for the next targeted campaign.
- **`Tribe\Admin\Notice\Marketing\Stellar_Sale`** — a separate, older `_deprecated_file()` stub from 6.3.0, unrelated to this one.
- **`Black_Friday`** — untouched and still rendering through all the same entry points.

**CSS note — every removal is scoped.** Diffing resolved selector chains before/after: **767 → 730, 37 removed, 0 of them outside a `.tribe-notice-stellar-sale*` ancestor.** All 31 removed `.tribe-marketing-notice*` rules sit under `&.tribe-notice-stellar-sale` or `&.tribe-notice-stellar-sale-2024`, so nothing can regress unless it renders Stellar Sale markup. The seven generic `.tribe-marketing-notice*` blocks (L1401–L1596) are untouched.

The markup that used these classes — `src/admin-views/notices/tribe-stellar-sale.php` — was deleted in `2955ac045` (2024-10-10); its classes map 1:1 to the removed block. Compiled CSS lives in the gitignored `build/`, so there's no built asset in this diff.

### 🔍 Consumer impact

Checked The Events Calendar, Event Tickets and Events Calendar Pro. **No companion PRs needed** — none of them reference `Stellar_Sale`, the `stellar-sale` slug, or its hooks anywhere outside their bundled `common/` checkout. Every `Stellar` match in those repos is the Strauss-prefixed `TEC\Common\StellarWP\…` vendor namespace. The change reaches them through the usual submodule bump.

They do use the Conditional Content system, but only via paths this change leaves intact:

- `tribe( Black_Friday::class )->render_narrow_banner_html()` in both welcome messages
- `do_action( 'tec_conditional_content_header_notice' )` — hook and contract unchanged, still answered by the Controller
- TEC's `Events_Pro_Upsell` (`Is_Dismissible`) and ET's `Inline_Upsell`

**One thing worth a reviewer's eye.** `Stellar_Sale::hook()` is now a no-op, so nothing in Common registers `wp_ajax_tec_conditional_content_dismiss` any more. That is not a regression:

- Nothing left in Common renders a dismiss button — `Black_Friday` isn't dismissible, `Inline_Upsell` is explicitly not dismissible, and `Notifications` uses its own `wp_ajax_ian_dismiss` endpoint and markup. The endpoint and the markup that used it disappear together.
- TEC's `Events_Pro_Upsell` registers its **own** handler and is self-contained.
- `Is_Dismissible::handle_dismiss()` early-returns on `$this->get_slug() !== $slug`, so registrants never handled each other's dismissals in the first place.

Also drops a small layering wart: `check_seating_license()` used to reach from Common into ET's `TEC\Tickets\Seating\Service\Service`; it now returns `false` without touching ET at all.

**Versioning:** `patch`. Deprecating rather than removing the class means this is no longer an API break, so it can ride in the maintenance release as intended.

### 🎥 Artifacts

n/a — this removes UI rather than adding it.

### ✔️ Checklist
- [ ] Code is covered by **NEW** `wpunit` or `integration` tests. — n/a; the deprecated class has no behaviour left to cover
- [x] Code is covered by **EXISTING** `wpunit` or `integration` tests. — `Promotional_Content_Test` and the trait tests still cover `Black_Friday` and the traits. Not run locally; leaving the suite to CI.
- [x] Check the base branch for your PR. — branched from `main`

### 📋 Follow-ups (not in this PR)

Flagging rather than fixing, to keep the diff surgical:

1. **The remaining marketing-notice CSS looks dead too.** `.notice-tribe-banner` and the seven generic `.tribe-marketing-notice*` rules have no markup generating them in this repo, TEC, ET or ECP. Deliberately left alone: unlike the removals above they aren't scoped to anything Stellar-Sale-specific, so removing them needs a check against the premium plugins (ETP, Filter Bar, Community, Promoter…) that I couldn't grep here. Worth its own ticket.
2. **`src/views/v2/components/icons/stellar-icon.php`** is now unreferenced. It's a theme-overridable template, so removing it is a separate call.


changelog/smtnc-1933-remove-stellar-sale-class.yaml
```diff
@@ -1,4 +1,4 @@
-significance: patch
+significance: minor
```
dave-green-uk: Claude suggested minor is more appropriate here given that we're removing a class.

changelog/smtnc-1933-remove-stellar-sale-class.yaml
```diff
@@ -0,0 +1,4 @@
...
+type: deprecated
+entry: 'Removed class: `TEC\Common\Admin\Conditional_Content\Stellar_Sale`'
```
dpanta94: I think this changelog is too technical. One is enough IMO

changelog/smtnc-1933-remove-stellar-sale.yaml
```diff
@@ -0,0 +1,4 @@
+significance: minor
+type: tweak
+entry: Removed the Stellar Sale promotional banner from the admin.
```
dpanta94:
```suggestion
entry: Removed the Stellar Sale promotional banner.
```

src/Common/Admin/Conditional_Content/Stellar_Sale.php
```diff
@@ -1,299 +0,0 @@
...
-/**
- * Set up for Stellar Sale promo.
- *
- * @since 6.8.2
- * @since 6.9.8 Modified to use the Has_Datetime_Conditions trait instead of extending the Datetime_Conditional_Abstract class.
- * @since 6.9.8 Modified to use the Requires_Capability trait.
- * @since 6.9.8 Modified to use the Has_Targeted_Creative_Upsell trait.
- * @since 6.9.8 Modified to use the Is_Dismissible trait.
- */
-class Stellar_Sale extends Promotional_Content_Abstract {
```
dpanta94, line 29: Let's deprecate and have the methods return static defaults

dpanta94 (changes requested):

changelog/smtnc-1933-remove-stellar-sale-class.yaml
dave-green-uk: Rolled this back

dpanta94 (approved):

PR #2979, opened 2026-08-18
[SMTNC-2031] Tribe Common (TEC/ET)

coutinhocouto:
### 🎫 Ticket

[SMTNC-2031](https://linear.app/nexcess/issue/SMTNC-2031/tribe-common-tecet)

### 🎥 Artifacts

image

### 🗒️ Description

Adds a dismissible WordPress admin notice prompting site administrators to update WordPress while a core update is available, via the shared `stellarwp/core-update-notice` library so the dismissal carries across every StellarWP plugin that displays it.

### ⚠️ Blockers

None.

### DIFF
--- a/changelog/smtnc-2031-tribe-common-tecet.yaml
+++ b/changelog/smtnc-2031-tribe-common-tecet.yaml
@@ -1,0 +1,5 @@
+significance: minor
+type: feature
+entry: Added an admin notice prompting administrators to update WordPress when a
+  core update is available.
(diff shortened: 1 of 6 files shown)


redscar (changes requested):
The code looks good to me. However, your tests are failing. Can you confirm that it wasn't introduced in this PR and if it was take care of them?

dpanta94:
Not being shipped

PR #2986, opened 2026-08-31
[SMTNC-1857] FSE Single Event template saves create single-event-N duplicates and revert the active template to default

coutinhocouto:
### 🎫 Ticket

[SMTNC-1857](https://linear.app/nexcess/issue/SMTNC-1857/fse-single-event-template-saves-create-single-event-n-duplicates-and)

### 🗒️ Description

Resolves the Full Site Editing template lookup to a single canonical row when several claim the same slug, and attaches the `wp_theme` term after insert so repeat resolutions stop creating new template rows.

### ⚠️ Blockers
None 

### ✔️ Checklist

### DIFF
--- a/src/Common/Editor/Full_Site/Template_Utils.php
+++ b/src/Common/Editor/Full_Site/Template_Utils.php
@@ -106,8 +106,8 @@ class Template_Utils {
 		$wp_query_args  = [
 			'post_name__in'  => [ $post_name ],
 			'post_type'      => 'wp_template',
-			'post_status'    => [ 'auto-draft', 'draft', 'publish', 'trash' ],
-			'posts_per_page' => 1,
+			'post_status'    => [ 'auto-draft', 'draft', 'publish' ],
+			'posts_per_page' => -1,
 			'no_found_rows'  => true,
 			'tax_query'      => [
 				[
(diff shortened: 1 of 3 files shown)


larodiel (approved):

src/Common/Editor/Full_Site/Template_Utils.php
```diff
@@ -213,4 +235,65 @@ public static function hydrate_block_template_by_post( WP_Post $post ): ?WP_Bloc
 
 		return $template;
 	}
+
+	/**
+	 * Orders the posts claiming a template slug so the canonical one comes first.
+	 *
+	 * Published rows outrank the rest, then the oldest row wins: when this bug has been reseeding
+	 * templates the newest row is the plugin's default markup and the oldest one still holds the
+	 * customization.
```
sdokus:
```suggestion
	 * Published rows outrank the rest, then the oldest row wins.
```
We shouldn't mention bugs in the code/docblocks - I think you should either just remove this part or rephrase to be focused on the 'why' of the current code, not so much focused on the history of how you got there... does that make sense?

src/Common/Editor/Full_Site/Template_Utils.php
```diff
@@ -213,4 +235,65 @@ public static function hydrate_block_template_by_post( WP_Post $post ): ?WP_Bloc
...
+	 *
+	 * @return void
+	 */
+	private static function rename_duplicate_block_template( WP_Post $post, string $post_name ): void {
+		wp_update_post(
+			[
+				'ID'        => $post->ID,
+				'post_name' => $post_name . '-duplicate-' . $post->ID,
+			]
+		);
```
sdokus:
```suggestion
		DB::update(
			DB::prefix( 'posts' ),
			[ 'post_name' => $post_name . '-duplicate-' . $post->ID ],
			[ 'ID' => $post->ID ]
		);
	clean_post_cache( $post->ID );
```

`wp_update_post()` re-saves the whole row, not just the slug, so we should do a direct column update instead of the whole pipeline. Re-saving the whole row can lead to unintentionally stripping the `post_content` out of core/html blocks.

sdokus: If you did go this route, you would just have to import `use TEC\Common\StellarWP\DB\DB;` at the top of the file. 

Also the docblock's second paragraph about `wp_unique_post_slug()` returning early for drafts can then just say the `ID` keeps each renamed row distinct

sdokus (approved):

PR #3026, opened 2026-09-19
[SMTNC-2786] Add capability guard to scale-back tests on tec-common.

larodiel:
[skip-changelog]

### 🎫 Ticket

[SMTNC-2786](https://linear.app/nexcess/issue/SMTNC-2786/add-capability-guard-to-scale-back-tests-on-tec-common)

### 📋 Plan

smtnc-2786

### 🗒️ Description

Guards the shared REST v1 status scale-back tests with the endpoint's post type capabilities, so roles that cannot create or edit the entity expect a permission error instead of a scaled-back status.

### 🎥 Artifacts 

**Before the fix**

Screenshot 2026-09-19 at 00 33 11

** After Fix **

Screenshot 2026-09-19 at 00 32 58

### ✔️ Checklist
- [x] There is an OpenSpec plan for this work in `tec-plans`, and it still describes what this PR actually does.
- [ ] Changelog file(s) added.
- [x] Code is covered by tests.
- [ ] All the **required** tests are passing.
- [x] Artifacts added.
- [ ] The linked ticket includes all the information for a QA engineer to verify the work.
- [ ] Pull Request is ready to be reviewed by a Human.

### 🤖 AI Disclosure

Was AI used for this PR [**Y**/n]: Y

Which model(s): Fable

In what way was AI used: Create a plan, dig into the code to check why the tests were failing, to write the test.

- [x] I own the work being done in this PR and I have personally reviewed it before asking another human to.


tests/_support/Testcases/REST/TEC/V1/Post_Entity_REST_Test_Case.php
```diff
@@ -407,7 +407,18 @@ public function test_create_scales_back_status_to_user_capabilities( string $rol
...
+		$response = $this->assert_endpoint(
+			$this->endpoint->get_base_path(),
+			'POST',
+			$user_can_create ? 201 : ( is_user_logged_in() ? 403 : 401 ),
+			$example
+		);
+
+		if ( ! $user_can_create ) {
+			return;
+		}
```
sdokus, line 421: Will this early bail cause the rest of the assertions in the test to never be reached? Should we skip instead of just returning? Something like: 
```
if ( ! $user_can_create ) {
	$this->markTestSkipped( 'Role cannot create this post type; nothing to scale back.' );
}
```

larodiel: I thought on it, but at the same time I was wondering if it's right, I'll implement that 😊

sdokus (approved):

tests/_support/Testcases/REST/TEC/V1/REST_Test_Case.php
```diff
@@ -115,33 +115,31 @@ public function different_user_roles_provider(): Generator {
...
-			},
-		];
-
-		yield [
-			'editor' => function (): void {
-				$user = $this->factory()->user->create( [ 'role' => 'editor' ] );
-				wp_set_current_user( $user );
-			},
-		];
+		foreach ( $this->get_roles_to_test() as $role ) {
```
dpanta94, line 118: The changes here are good!

tests/_support/Testcases/REST/TEC/V1/Post_Entity_REST_Test_Case.php
```diff
@@ -407,7 +407,18 @@ public function test_create_scales_back_status_to_user_capabilities( string $rol
 
 		wp_set_current_user( $this->factory()->user->create( [ 'role' => $role ] ) );
 
-		$response = $this->assert_endpoint( $this->endpoint->get_base_path(), 'POST', 201, $example );
+		$user_can_create = current_user_can( get_post_type_object( $this->endpoint->get_post_type() )->cap->create_posts );
```
dpanta94, line 410: Since we are now allowing for the provider to be overwritten, do we need the changes in this file ?

i feel not, since we want to to continue testing that the status scales back and be aware if we are testing with a role thats failing so we can update our code.

dpanta94 (changes requested):

tests/_support/Testcases/REST/TEC/V1/Post_Entity_REST_Test_Case.php
larodiel: Hey @dpanta94 it's still being neeed.
> For an endpoint whose post type restricts create_posts/edit_post to different roles (e.g. WooCommerce products, where contributor/author don't hold edit_products), contributor/author won't be able to create/update at all, so without the guard the test would assert against a 403 response instead of a scaled-back status — hence the markTestSkipped() when the role can't act.

Does it makes sense to you?

dpanta94: What im suggesting is for us to prefer the failure over the skip. So that we can amend it when and if it fails again.

dpanta94 (approved):

